PR #7, opened 2018-10-25
TiKV Rust Client RFC

sunxiaoguang:
Introduces the API design for a Rust client for TiKV.


text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,190 @@
...
+
+## Installation.
+
+To utilize the client programmatically, users will be able to add the `tikv-client` crate to their `Cargo.toml`'s dependencies. Then they must use the crate with `use tikv_client;`. Unfortunately due to Rust’s naming conventions this inconsistency is in place.
+
+To utilize the command line client, users will be able to install the binary via `cargo install tikv-client`. They will then be able to access the client through the binary `tikv-client`. If they wish for a different name they can alias it in their shell.
+
+## Usage
+
+## Key structs and traits
```
siddontang: I think we don't need to list the interfaces one by one because there are too many and we may still change some. 

Here I suggest just mention Raw mode, Transaction Mode, the associated client, and a simple example of how to read + write a key through Raw or Txn Client.

siddontang:
PTAL @tikv/pingcap-i18n

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,190 @@
+# Summary
+
+Introduce a full featured, official TiKV (and PD) Rust client. It will be intended to be used as a reference implementation, or to provide C-compatible binding for future clients.
```
lilin90:
```suggestion
Introduce a full featured, official TiKV (and PD) Rust client. It is intended to be used as a reference implementation, or to provide C-compatible binding for future clients.
```

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,190 @@
...
+# Detailed design
+## Supported Targets
+
+We will target the `stable` channel of Rust starting in the Rust 2018 edition. We choose to begin with Rust 2018 so we do not need to concern ourselves with an upgrade path.
+
+We will also support the most recent `nightly` version of Rust, but users should not feel the need to reach for stable unless they are already using it.
+
+## Naming
+
+While the [Rust API Guidelines](https://rust-lang-nursery.github.io/api-guidelines/naming.html) do not perscribe any particular crate name convention.
```
lilin90:
```suggestion
The [Rust API Guidelines](https://rust-lang-nursery.github.io/api-guidelines/naming.html) do not prescribe any particular crate name convention.
```

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,190 @@
...
+
+We choose to name the crate `tikv_client` to conform to the constraints presented to us by the Rust compiler. Cargo permits `tikv-client` and `tikv_client`, but `rustc` does not permit `tikv-client` so we choose to use `tikv_client` to reduce mental overhead.
+
+Choosing to seperate `tikv` and `client` helps potentially unfamiliar users to immediately understand the intent of the package. `tikvclient`, while understandable, is not immediately parsable by a human.
+
+All structures and functions will otherwise follow the [Rust API Guidelines](https://rust-lang-nursery.github.io/api-guidelines/), some of which will be enforced by `clippy`.
+
+## Installation.
```
lilin90:
```suggestion
## Installation
```

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,190 @@
...
+Introduce a full featured, official TiKV (and PD) Rust client. It will be intended to be used as a reference implementation, or to provide C-compatible binding for future clients.
+
+# Motivation
+
+Currently, users of TiKV must use [TiDB's Go Client](https://github.com/pingcap/tidb/blob/master/store/tikv/client.go), which is not well packaged or documented. We would like to ensure that users can easily use TiKV and PD without needing to use TiDB.
+
+We think this would help encourage community participation in the TiKV project and associated libraries. During talks with several potential corporate users we discovered that there was an interest in using TiKV to resolve concerns such as caching and raw key-value stores.
+
+# Detailed design
+## Supported Targets
```
lilin90: -> Supported **t**argets?

sunxiaoguang: Fixed

text/2018-10-25-tikv-client-rust.md
sunxiaoguang: Fixed

text/2018-10-25-tikv-client-rust.md
sunxiaoguang: Fixed

text/2018-10-25-tikv-client-rust.md
sunxiaoguang: Fixed. Please take another look of the newly added API usage section, thanks.

text/2018-10-25-tikv-client-rust.md
sunxiaoguang: I copied some document from go API as template and modified it a little bit here. Please take another look. 
BTW: After I came up with examples. I have a feeling that the API might need some more changes, especially the transactional API. Will create another commit and discuss it [here](https://github.com/tikv/client-rust/pull/3)

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,386 @@
...
+Currently, users of TiKV must use [TiDB's Go Client](https://github.com/pingcap/tidb/blob/master/store/tikv/client.go), which is not well packaged or documented. We would like to ensure that users can easily use TiKV and PD without needing to use TiDB.
+
+We think this would help encourage community participation in the TiKV project and associated libraries. During talks with several potential corporate users we discovered that there was an interest in using TiKV to resolve concerns such as caching and raw key-value stores.
+
+# Detailed design
+## Supported targets
+
+We will target the `stable` channel of Rust starting in the Rust 2018 edition. We choose to begin with Rust 2018 so we do not need to concern ourselves with an upgrade path.
+
+We will also support the most recent `nightly` version of Rust, but users should not feel the need to reach for stable unless they are already using it.
```
Hoverbear: "Read for nightly" :grin:

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,386 @@
...
+
+## Naming
+
+The [Rust API Guidelines](https://rust-lang-nursery.github.io/api-guidelines/naming.html) do not perscribe any particular crate name convention.
+
+We choose to name the crate `tikv_client` to conform to the constraints presented to us by the Rust compiler. Cargo permits `tikv-client` and `tikv_client`, but `rustc` does not permit `tikv-client` so we choose to use `tikv_client` to reduce mental overhead.
```
Hoverbear: We chose `tikv-client`.

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,386 @@
...
+
+Choosing to seperate `tikv` and `client` helps potentially unfamiliar users to immediately understand the intent of the package. `tikvclient`, while understandable, is not immediately parsable by a human.
+
+All structures and functions will otherwise follow the [Rust API Guidelines](https://rust-lang-nursery.github.io/api-guidelines/), some of which will be enforced by `clippy`.
+
+## Installation
+
+To utilize the client programmatically, users will be able to add the `tikv-client` crate to their `Cargo.toml`'s dependencies. Then they must use the crate with `use tikv_client;`. Unfortunately due to Rust’s naming conventions this inconsistency is in place.
+
+To utilize the command line client, users will be able to install the binary via `cargo install tikv-client`. They will then be able to access the client through the binary `tikv-client`. If they wish for a different name they can alias it in their shell.
```
Hoverbear: Maybe we should leave this for a future RFC to prevent this one from becoming huge. Besides we cannot effectively start on this until after the library is done.

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,386 @@
...
+
+### The common data types
+
+- Key: raw binary data
+- Value: raw binary data
+- KvPair: Key-value pair type
+- KeyRange: Half-open interval of keys
+- Config: Configuration for client
+
+### Try the Raw Key-Value API
```
Hoverbear:
```suggestion
### Raw Key-Value API Usage Examples
```

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,386 @@
...
+
+To use the Raw Key-Value API, take the following steps:
+
+1. Create an instance of Config to specify endpoints of PD (Placement Driver) and optional security config
+
+```rust
+    let config = Config::new(vec!["127.0.0.1:2379"]);
```
Hoverbear: A security config is mentioned but the example does not include one

text/2018-10-25-tikv-client-rust.md
Hoverbear: Since this is an RFC we are more focused on "What using this looks and feels like" over teaching users how to use it. We must think beyond the simple cases. :)

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,386 @@
...
+        K: Into<Key>,
+        C: Into<Option<String>>;
+
+    fn batch_delete<I, K, C>(&self, keys: I, cf: C) -> KvFuture<()>
+    where
+        I: IntoIterator<Item = K>,
+        K: Into<Key>,
+        C: Into<Option<String>>;
+
+    fn scan<R, C>(&self, range: R, limit: u32, key_only: bool, cf: C) -> KvFuture<Vec<KvPair>>
```
Hoverbear: I wonder if it is easier for users if we stick to key/range being the first param and CF always being the second. :thinking:

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,386 @@
...
+fn main() {
+    let config = Config::new(vec!["127.0.0.1:3379"]);
+    let raw = raw::RawClient::new(&config)
+        .wait()
+        .expect("Could not connect to tikv");
+
+    let key: Key = b"Company".to_vec().into();
+    let value: Value = b"PingCAP".to_vec().into();
+
+    raw.put((Clone::clone(&key), Clone::clone(&value)), None)
```
Hoverbear: It would be very nice to avoid the clones here.

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,386 @@
...
+
+The Transactional Key-Value API is more complicated than the Raw Key-Value API. Some transaction related concepts are listed as follows.
+
+- Client
+
+    Like the Raw Key-Value client, a Client is client to a TiKV cluster.
+
+- Snapshot
+
+    A Snapshot is the state of a Client at a particular point of time, which provides some readonly methods. The multiple times read from a same Snapshot is guaranteed consistent.
```
Hoverbear: "multiple reads of the same snapshot..."

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,386 @@
...
+
+- Transaction
+
+    Like the Transaction in SQL, a Transaction symbolizes a series of read and write operations performed within the Client. Internally, a Transaction consists of a Snapshot for reads, and a buffer for all writes. The default isolation level of a Transaction is Snapshot Isolation.
```
Hoverbear: Since we're talking about a default we should detail how to configure it

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,386 @@
...
+where
+    P: IntoIterator<Item = I>,
+    I: Into<KvPair>,
+{
+    let mut txn = client.begin().wait().expect("Could not begin transaction");
+    let _: Vec<()> = pairs
+        .into_iter()
+        .map(Into::into)
+        .map(|p| {
+            txn.set(p).wait().expect("Could not set key value pair");
```
Hoverbear: Maybe another an example that uses something like futures::join_all? I think many users won't want to call .wait() here.

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,386 @@
...
+
+```bash
+(Key([107, 101, 121, 49]), Value([118, 97, 108, 117, 101, 49]))
+(Key([107, 101, 121, 49]), Value([118, 97, 108, 117, 101, 49]))
+(Key([107, 101, 121, 49]), Value([118, 97, 108, 117, 101, 49]))
+```
+
+## Programming model
+
+The client instance is thread safe and all the interfaces return futures so users can use the client in either asynchronous or synchronous way. A dedicated event loop thread is created at per client instance basis to drive the reactor to make progress.
```
Hoverbear: created at a`
```suggestion
The client instance is thread safe and all the interfaces return futures so users can use the client in either asynchronous or synchronous way. A dedicated event loop thread is created at a per client instance basis to drive the reactor to make progress.
```

Hoverbear: Is it possible to attach to another event loop?

Hoverbear:
Great job @sunxiaoguang, thanks for tackling this huge task! My hero. :joy:

text/2018-10-25-tikv-client-rust.md
Hoverbear: Hm, you're right. I think in the case of put it's okay... I note that `HashMap::get` takes a borrow though: https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.get

I think in this case mutations, eg `put`, which actually go and send the value (and thus almost certainly need to do memory work) should take. But maybe we can make the searching/get based functions accept just a borrow? Similar to `HashMap` or `HashSet`?

sunxiaoguang: The gRPC part is going to move whatever parameters passed in anyway. In addition the nature of asynchronous programming model requires moving and sending parameters to event loop thread, so it doesn't matter if it's mutation or not the arguments are going to be moved anyway. However I do feel this particular API guideline applies to regular libraries better then the scenario we are facing. So breaking this rule probably is a better choice as it makes our users' life easier.

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,386 @@
...
+
+## Tooling
+
+The `tikv_client` crate will be tested with Travis CI using Rust's standard testing framework. We will also include benchmarking with criterion in the future. For public functions which process user input, we will seek to use fuzz testing such as `quickcheck` to find subtle bugs.
```
rleungx:
```suggestion
The `tikv_client` crate will be tested with Travis CI using Rust's standard testing framework. We will also include benchmark with criterion in the future. For public functions which process user input, we will seek to use fuzz testing such as `quickcheck` to find subtle bugs.
```

sunxiaoguang: Fixed

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,386 @@
...
+
+To utilize the command line client, users will be able to install the binary via `cargo install tikv-client`. They will then be able to access the client through the binary `tikv-client`. If they wish for a different name they can alias it in their shell.
+
+## Usage
+
+## Two types of APIs
+TiKV provides two types of APIs for developers:
+- The Raw Key-Value API
+
+    If your application scenario does not need distributed transactions or MVCC (Multi-Version Concurrency Control) and only need to guarantee the atomicity towards one key, you can use the Raw Key-Value API.
```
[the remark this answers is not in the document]
sunxiaoguang: Fixed

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,386 @@
...
+TiKV provides two types of APIs for developers:
+- The Raw Key-Value API
+
+    If your application scenario does not need distributed transactions or MVCC (Multi-Version Concurrency Control) and only needs to guarantee the atomicity towards one key, you can use the Raw Key-Value API.
+
+- The Transactional Key-Value API
+
+    If your application scenario requires distributed ACID transactions and the atomicity of multiple keys within a transaction, you can use the Transactional Key-Value API.
+
+Generally the Raw Key-Value API has higher throughput and lower latency compare to the Transactional Key-Value API. If distributed ACID transactions is not required, Raw Key-Value API is preferred over Transactional Key-Value API for better performance and ease of use.
```
lilin90:
```suggestion
Generally, the Raw Key-Value API has higher throughput and lower latency compared to the Transactional Key-Value API. If distributed ACID transactions are not required, Raw Key-Value API is preferred over Transactional Key-Value API for better performance and ease of use.
```

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,386 @@
...
+
+```rust
+    let config = Config::new(vec!["127.0.0.1:2379"]);
+```
+
+2. Create a Transactional Key-Value client.
+
+```rust
+    let client = TxnClient::new(&config);
+```
```
lilin90: Please also indent the code blocks in step 1&2 like that in step 5.

text/2018-10-25-tikv-client-rust.md
sunxiaoguang: Fixed

text/2018-10-25-tikv-client-rust.md
sunxiaoguang: Fixed, but I'm not sure if I'm doing it right, please double check it.

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,386 @@
...
+    where
+        P: Into<KvPair>;
+
+    fn delete<K>(&mut self, key: K) -> KvFuture<()>
+    where
+        K: Into<Key>;
+
+    fn seek<K>(&self, key: K) -> KvFuture<Scanner>
+    where
+        K: Into<Key>;    Begin() -> Txn
```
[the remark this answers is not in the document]
sunxiaoguang: Hm, not sure if I misunderstood anything here. I compiled the sample code with nightly toolchain.

text/2018-10-25-tikv-client-rust.md
sunxiaoguang: Not right now. Maybe it is intimidating for users to know eventloop?

text/2018-10-25-tikv-client-rust.md
sunxiaoguang: Fixed

text/2018-10-25-tikv-client-rust.md
sunxiaoguang: Compare to other parameters, the cf parameter is optional and is it a good idea to have an optional parameter in the middle? Although it is an inappropriate example, but C++ default arguments have to come after normal arguments.

BTW: I have another comment on this signature. Having a KvPair seems to make sense because it might be returned from get. But KeyRange is always input arguments, we can easily break it into two arguments at cost of one more argument. Maybe it's better to just remove this type?

Hoverbear: It might make sense. :thinking:

text/2018-10-25-tikv-client-rust.md
sunxiaoguang: I changed it to AsRef<Key> and it does look better. Please take a look at the new commit. @Hoverbear

Hoverbear: @siddontang What do you think about this?

text/2018-10-25-tikv-client-rust.md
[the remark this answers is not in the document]
sunxiaoguang: Oh, I see what you mean now. The Begin() -> Txn was a typo, I probably copied it from golang document and forgot to delete it. It's removed now.

text/2018-10-25-tikv-client-rust.md
Hoverbear: I think it is, let's leave this for the future. :)

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,395 @@
...
+2. Create a Raw Key-Value client.
+
+    ```rust
+        let client = RawClient::new(&config);
+    ```
+
+3. Call the Raw Key-Value client methods to access the data on TiKV. The Raw Key-Value API contains following methods
+
+    ```rust
+        fn get<K, C>(&self, key: K, cf: C) -> KvFuture<Value>
```
gregwebs: what is the `cf` parameter? As someone new to TiKV I don't know what this abbreviation means.

sunxiaoguang: It means column family. Maybe adding document is better than column_family?

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,395 @@
...
+            .map(Into::into)
+            .map(|p| {
+                txn.set(p).wait().expect("Could not set key value pair");
+            }).collect();
+        txn.commit().wait().expect("Could not commit transaction");
+    }
+
+    fn get(client: &TxnClient, key: &Key) -> Value {
+        let txn = client.begin().wait().expect("Could not begin transaction");
+        txn.get(key).wait().expect("Could not get value")
```
gregwebs: What happens to the transaction at this point?

text/2018-10-25-tikv-client-rust.md
gregwebs: A newtype could be helpful here so that the type is `ColumnFamily` rather than `String`.

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,395 @@
...
+    fn main() {
+        let config = Config::new(vec!["127.0.0.1:3379"]);
+        let raw = raw::RawClient::new(&config)
+            .wait()
+            .expect("Could not connect to tikv");
+
+        let key: Key = b"Company".to_vec().into();
+        let value: Value = b"PingCAP".to_vec().into();
+
+        raw.put((Clone::clone(&key), Clone::clone(&value)), None)
```
gregwebs: Adding `None` to all the API calls is unsatisfying. Does it make sense to put the column family in the `Config`? Alternatively, if we have such a small core API surface we could expose two functions with the different signatures.

sunxiaoguang: Hm, maybe adding another method sounds better? User may use multiple column family for different data, so it makes better sense to let user choose which cf to use at per call basis.

sunxiaoguang: Like put_with_cf or put_with_column_family

text/2018-10-25-tikv-client-rust.md
sunxiaoguang: I assume that transaction will be dropped before returning to caller. So basically it is trying to read something in a transaction.

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,395 @@
...
+        where
+            K: AsRef<Key>;
+
+        fn seek<K>(&self, key: K) -> KvFuture<Scanner>
+        where
+            K: AsRef<Key>;
+
+        fn commit(self) -> KvFuture<()>;
+
+        fn rollback(self) -> KvFuture<()>;
```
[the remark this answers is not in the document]
sunxiaoguang: Hm, there should be. Let me add it.

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,395 @@
...
+        where
+            K: AsRef<[Key]>,
+            C: Into<Option<String>>;
+
+        fn put<P, C>(&self, pair: P, cf: C) -> KvFuture<()>
+        where
+            P: Into<KvPair>,
+            C: Into<Option<String>>;
+
+        fn batch_put<I, P, C>(&self, pairs: I, cf: C) -> KvFuture<()>
```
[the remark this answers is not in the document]
sunxiaoguang: Good idea. It definitely looks better than many single letter type variables.

text/2018-10-25-tikv-client-rust.md
gregwebs: I would probably favor `put_cf` so the method name is not really long. Or if we should encourage use of column families, make that the shorter api and have `put_def` for using the default column family.

I would be interested in seeing the builder style. I am actually wondering if it makes sense to try to build it into the Key type since it partitions keys.

text/2018-10-25-tikv-client-rust.md
sunxiaoguang: Good point

sunxiaoguang: That sounds great.

text/2018-10-25-tikv-client-rust.md
sunxiaoguang: CF is only available in Raw KV interface. So we can move Key/Value/KvPair definition to raw and transactional module so they can have different definition.

text/2018-10-25-tikv-client-rust.md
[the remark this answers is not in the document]
sunxiaoguang: BTW: The inlining style you mentioned? Did you mean this?
```rust
    fn batch_put<
        IntoIter: IntoIterator<Item = Pair>,
        Pair: Into<KvPair>,
        ColumnFamily: Into<Option<String>>,
    >(  
        &self,
        _pairs: Pair,
        _cf: ColumnFamily,
    ) -> KvFuture<()>;
```
Using trait bound directly as argument type is prohibited.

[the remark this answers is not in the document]
sunxiaoguang: Oh, I see what you mean now. It would be something like this
```rust
    fn batch_put(                                                                                                                                              
        &self,                                                                                                                                                 
        pairs: impl IntoIterator<Item = impl Into<KvPair>>,                                                                                                    
        cf: impl Into<Option<String>>,                                                                                                                         
    ) -> KvFuture<()>;
```
It's the first time I realize the reason why impl trait exists for arguments. It's definitely easier to read than where syntax. Thanks.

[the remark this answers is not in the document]
sunxiaoguang: Sadly, this particular case isn't supported.
```rust
    fn batch_scan(
        &self,
        ranges: impl AsRef<[impl RangeBounds<Key>]>,
        each_limit: u32,
        key_only: bool,
        cf: impl Into<Option<String>>,
    ) -> KvFuture<Vec<KvPair>>;
```
I'm going to leave this one with the old generic syntax and change all other methods to impl trait syntax.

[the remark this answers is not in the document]
sunxiaoguang: And this way
```rust
    fn batch_scan<Ranges: AsRef<[Bounds]>, Bounds: RangeBounds<Key>, CF: Into<Option<String>>>(                                                                
        &self,
        ranges: Ranges,
        each_limit: u32,
        key_only: bool,
        cf: CF,
    ) -> KvFuture<Vec<KvPair>>;  
```
looks less than where style for me
```rust
    fn batch_scan<Ranges, Bounds, CF>(
        &self,
        ranges: Ranges,
        each_limit: u32,
        key_only: bool,
        cf: CF,
    ) -> KvFuture<Vec<KvPair>>
    where
        Ranges: AsRef<[Bounds]>,
        Bounds: RangeBounds<Key>,
        CF: Into<Option<String>>;
```
What do you think? @gregwebs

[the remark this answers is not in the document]
gregwebs: Great! I agree on `batch_scan`. I am not against `where` at all, just against the common practice of using one letter variable names, particularly when they are not 100% consistent in their meaning.

text/2018-10-25-tikv-client-rust.md
sunxiaoguang: I'm planning to change it to request builder style interface. Haven't done it yet, will try to fix couple of issues we discussed all together.

text/2018-10-25-tikv-client-rust.md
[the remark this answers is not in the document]
sunxiaoguang: It's somewhat like this now
```rust
pub struct BatchScan<'a, AClient: Client>;

impl<'a, AClient: Client> BatchScan<'a, AClient> {
    fn new(client: &'a AClient, ranges: Vec<(Key, Key)>, each_limit: u32) -> Self;
    pub fn key_only(mut self) -> Self; 
    pub fn cf(mut self, cf: impl Into<ColumnFamily>) -> Self;
}

impl<'a, AClient: Client> Future for BatchScan<'a, AClient> {
    type Item = Vec<KvPair>;
    type Error = (); 
    fn poll(&mut self) -> Poll<Self::Item, Self::Error>;
}

pub trait Client {
    type AClient: Client;

    fn batch_scan<Ranges, Bounds>(
        &self,
        ranges: Ranges,
        each_limit: u32,
    ) -> BatchScan<Self::AClient>
    where
        Ranges: AsRef<[Bounds]>,
        Bounds: RangeBounds<Key>;
}
``` 
You can find the complete interface at [here](https://github.com/tikv/client-rust/blob/public_api_rfc/src/raw.rs)

PLTA @gregwebs

text/2018-10-25-tikv-client-rust.md
sunxiaoguang: None is eliminated by builder api

gregwebs:
I really like the API now :)
With data types being built up for operations there is a future possibility of supporting push-down computation from a DSL. A next version though :)

text/2018-10-25-tikv-client-rust.md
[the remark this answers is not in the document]
Hoverbear: Hmmm, I don't think the `AClient: Client` reads very well. In this case since it's pretty clear we expect a `Client` (there aren't multiple bounds, and the trait name is very clear) so I think just `C: Client` is sufficient. This is fairly idiomatic for Rust code.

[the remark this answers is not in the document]
Hoverbear: Notably this looks strange as the associated type in `trait Client { type ... }`

Hoverbear:
@gregwebs Yup, let's keep this RFC minimal and mergable. :)

Hoverbear:
I think we can move into the final comment period. If you have concerns please start a discussion here about it before Friday. We'll then let all existing discussions close before merging.

Thanks for your participation in this process. :)

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,389 @@
+# Summary
+
+Introduce a full featured, official TiKV (and PD) Rust client. It is intended to be used as a reference implementation, or to provide C-compatible binding for future clients.
+
+# Motivation
+
+Currently, users of TiKV must use [TiDB's Go Client](https://github.com/pingcap/tidb/blob/master/store/tikv/client.go), which is not well packaged or documented. We would like to ensure that users can easily use TiKV and PD without needing to use TiDB.
```
crazycoder1988:
```suggestion
Currently, TiKV users must use [TiKV's Go Client provided by TiDB](https://github.com/pingcap/tidb/blob/master/store/tikv/client.go), which is not well packaged or documented. We would like to ensure that users can easily use TiKV and PD without needing to use TiDB.
```

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,389 @@
...
+
+Choosing to seperate `tikv` and `client` helps potentially unfamiliar users to immediately understand the intent of the package. `tikvclient`, while understandable, is not immediately parsable by a human.
+
+All structures and functions will otherwise follow the [Rust API Guidelines](https://rust-lang-nursery.github.io/api-guidelines/), some of which will be enforced by `clippy`.
+
+## Installation
+
+To utilize the client programmatically, users will be able to add the `tikv-client` crate to their `Cargo.toml`'s dependencies. Then they must use the crate with `use tikv_client;`. Unfortunately due to Rust’s naming conventions this inconsistency is in place.
+
+To utilize the command line client, users will be able to install the binary via `cargo install tikv-client`. They will then be able to access the client through the binary `tikv-client`. If they wish for a different name they can alias it in their shell.
```
crazycoder1988: What's the scope of a command line client? Is it like tikv-ctl or something else?

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,389 @@
...
+
+## Usage
+
+## Two types of APIs
+TiKV provides two types of APIs for developers:
```
crazycoder1988: New line

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,389 @@
...
+Introduce a full featured, official TiKV (and PD) Rust client. It is intended to be used as a reference implementation, or to provide C-compatible binding for future clients.
+
+# Motivation
+
+Currently, users of TiKV must use [TiDB's Go Client](https://github.com/pingcap/tidb/blob/master/store/tikv/client.go), which is not well packaged or documented. We would like to ensure that users can easily use TiKV and PD without needing to use TiDB.
+
+We think this would help encourage community participation in the TiKV project and associated libraries. During talks with several potential corporate users we discovered that there was an interest in using TiKV to resolve concerns such as caching and raw key-value stores.
+
+# Detailed design
+## Supported targets
```
crazycoder1988: New line

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,389 @@
...
+
+    If your application scenario requires distributed ACID transactions and the atomicity of multiple keys within a transaction, you can use the Transactional Key-Value API.
+
+Generally, the Raw Key-Value API has higher throughput and lower latency compared to the Transactional Key-Value API. If distributed ACID transactions are not required, Raw Key-Value API is preferred over Transactional Key-Value API for better performance and ease of use.
+
+The client provides two types of APIs in two separate modules for developers to choose from.
+
+### The common data types
+
+- Key: raw binary data
```
crazycoder1988: Better to leave the actual Rust type as well. If it is already a type name, please surround it with ``.

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,389 @@
...
+- Value: raw binary data
+- KvPair: Key-value pair type
+- KeyRange: Half-open interval of keys
```
crazycoder1988: Please define "Half-open"

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,389 @@
...
+- Config: Configuration for client
+
+### Raw Key-Value API Basic Usage
+
+To use the Raw Key-Value API, take the following steps:
+
+1. Create an instance of Config to specify endpoints of PD (Placement Driver) and optional security config
+
+    ```rust
+    let config = Config::new(vec!["127.0.0.1:2379"]);
```
crazycoder1988: I would prefer something like:

```rust
let config = Config::default()
    .pd_address(&[...])
    .tls(...);
```

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,389 @@
...
+    ```
+
+2. Create a Raw Key-Value client.
+
+    ```rust
+    let client = RawClient::new(&config);
```
crazycoder1988: Maybe `RawKvClient` is better because it is actually not a raw client, but a raw (KV) client.

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,389 @@
...
+
+    ```bash
+    Successfully put Key([67, 111, 109, 112, 97, 110, 121]):Value([80, 105, 110, 103, 67, 65, 80]) to tikv
+    Found val: Value([80, 105, 110, 103, 67, 65, 80]) for key: Key([67, 111, 109, 112, 97, 110, 121])
+    Key: Key([67, 111, 109, 112, 97, 110, 121]) deleted
+    ```
+
+Raw Key-Value client is a client of the TiKV server and only supports the GET/BATCH_GET/PUT/BATCH_PUT/DELETE/BATCH_DELETE/SCAN/BATCH_SCAN/DELETE_RANGE commands. The Raw Key-Value client can be safely and concurrently accessed by multiple threads. Therefore, for one process, one client is enough generally.
+
+### Try the Transactional Key-Value API
```
crazycoder1988: The previous section is called "Usage ... " and this section might follow the same style as well

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,389 @@
...
+
+The Transactional Key-Value API is more complicated than the Raw Key-Value API. Some transaction related concepts are listed as follows.
+
+- Client
+
+    Like the Raw Key-Value client, a Client is client to a TiKV cluster.
+
+- Snapshot
+
+    A Snapshot is the state of a Client at a particular point of time, which provides some readonly methods. The multiple reads of the same Snapshot is guaranteed consistent.
```
crazycoder1988: Is it really "the state of a Client at a particular point of time" instead of "the state of the TiKV data ..."?

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,389 @@
...
+    ```
+
+3. (Optional) Modify data using a Transaction.
+
+    The lifecycle of a Transaction is: _begin → {get, set, delete, scan} → {commit, rollback}_.
+
+4. Call the Transactional Key-Value API's methods to access the data on TiKV. The Transactional Key-Value API contains the following methods:
+
+    ```rust
+    fn begin(&self) -> KvFuture<Transaction>;
```
crazycoder1988: What is KvFuture?

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,389 @@
...
+    let client = RawClient::new(&config);
+    ```
+
+3. Call the Raw Key-Value client methods to access the data on TiKV. The Raw Key-Value API contains following methods
+
+    ```rust
+    pub trait Client {
+        type AClient: Client;
+
+        fn get(&self, key: impl AsRef<Key>) -> Get<Self::AClient>;
```
crazycoder1988: What is `Get<T>`?

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,389 @@
...
+
+    fn set<P>(&mut self, pair: P) -> KvFuture<()>
+    where
+        P: Into<KvPair>;
+
+    fn delete<K>(&mut self, key: K) -> KvFuture<()>
+    where
+        K: AsRef<Key>;
+
+    fn seek<K>(&self, key: K) -> KvFuture<Scanner>
```
crazycoder1988: how to do reverse seek and specify upper bound & lower bound?

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,389 @@
...
+
+    fn scan(client: &TxnClient, start: &Key, limit: usize) {
+        let txn = client.begin().wait().expect("Could not begin transaction");
+        let mut scanner = txn.seek(start).wait().expect("Could not seek to start key");
+        let mut limit = limit;
+        loop {
+            if limit == 0 {
+                break;
+            }
+            match scanner.poll() {
```
crazycoder1988: As a usage example, prefer to avoid calling the low level `poll()` API.

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,389 @@
...
+                Ok(Async::Ready(Some(pair))) => {
+                    limit -= 1;
+                    println!("{:?}", pair);
+                }
+                _ => break,
+            }
+        }
+    }
+
+    fn dels<P>(client: &TxnClient, pairs: P)
```
crazycoder1988: It is not pairs, only keys?

text/2018-10-25-tikv-client-rust.md
Hoverbear: @sunxiaoguang I think we talked about leaving the CLI client to another RFC, do you agree? If so lets remove this.

text/2018-10-25-tikv-client-rust.md
Hoverbear: `pd_address` is a required field at this time, TiKV won't work without it, so it's better to have it as a required argument.

text/2018-10-25-tikv-client-rust.md
Hoverbear: I think `tikv_client::RawClient` already implies the `Kv`. Users can choose to rename it when they import it if they are dealing with multiple Client types. You may notice for example Hyper doesn't have `HyperClient` or `HttpClient`, it just has [`Client`](https://hyper.rs/hyper/0.8.0/hyper/client/index.html)

Hoverbear:
@crazycoder1988 Hi! Thanks for your input/review! :)

text/2018-10-25-tikv-client-rust.md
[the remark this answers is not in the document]
sunxiaoguang: Makes sense, let's change it first and see how it looks.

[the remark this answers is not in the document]
sunxiaoguang: Oh, just want to double check if you meant only the AClient: Client inside Client trait. 
There are quite some AClient: Client out there.

text/2018-10-25-tikv-client-rust.md
sunxiaoguang: Sure. Let me remove the tikv-ctl part.

text/2018-10-25-tikv-client-rust.md
sunxiaoguang: Sorry that I forgot to remove KeyRange here. It's been removed and changed to standard RangeBounds trait so it can represent both half-open and closed intervals after edition.

text/2018-10-25-tikv-client-rust.md
sunxiaoguang: @Hoverbear Is it too verbose to include everything here? I mean the Definition of Get/BatchGet/Put/BatchPut/Transaction/Scanner etc.

text/2018-10-25-tikv-client-rust.md
sunxiaoguang: There is a reverse_seek method that resembles the Go Client.

sunxiaoguang: BTW: Just updated it to use RangeBounds as well. Since ScanRequest supports has end_keys now. Using RangeBounds makes better sense.

text/2018-10-25-tikv-client-rust.md
sunxiaoguang: Oops. That was a copy paste mistake. Fixed.

text/2018-10-25-tikv-client-rust.md
sunxiaoguang: It's added to common data types section now. Thanks

text/2018-10-25-tikv-client-rust.md
sunxiaoguang: @Hoverbear PTAL about this line. I'm not sure if I made it clear.

text/2018-10-25-tikv-client-rust.md
sunxiaoguang: This CLI section was removed.

text/2018-10-25-tikv-client-rust.md
sunxiaoguang: Fixed

text/2018-10-25-tikv-client-rust.md
sunxiaoguang: Sure, let me add it.

text/2018-10-25-tikv-client-rust.md
sunxiaoguang: Fixed

text/2018-10-25-tikv-client-rust.md
sunxiaoguang: Added an optional step in example.

text/2018-10-25-tikv-client-rust.md
sunxiaoguang: @lilin90 Any comments?

Hoverbear: Both are OK. But I think we can add the Java client now: https://github.com/tikv/client-java

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,398 @@
...
+
+All structures and functions will otherwise follow the [Rust API Guidelines](https://rust-lang-nursery.github.io/api-guidelines/), some of which will be enforced by `clippy`.
+
+## Installation
+
+To utilize the client programmatically, users will be able to add the `tikv-client` crate to their `Cargo.toml`'s dependencies. Then they must use the crate with `use tikv_client;`. Unfortunately due to Rust’s naming conventions this inconsistency is in place.
+
+## Usage
+
+## Two types of APIs
```
Hoverbear: If I recall correctly we need to advise users that they are mutually exclusive, and they must choose one or the other. I think that is worth mentioning in this section.

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,398 @@
...
+
+The result is like:
+
+    ```bash
+    Successfully put Key([67, 111, 109, 112, 97, 110, 121]):Value([80, 105, 110, 103, 67, 65, 80]) to tikv
+    Found val: Value([80, 105, 110, 103, 67, 65, 80]) for key: Key([67, 111, 109, 112, 97, 110, 121])
+    Key: Key([67, 111, 109, 112, 97, 110, 121]) deleted
+    ```
+
+Raw Key-Value client is a client of the TiKV server and only supports the GET/BATCH_GET/PUT/BATCH_PUT/DELETE/BATCH_DELETE/SCAN/BATCH_SCAN/DELETE_RANGE commands. The Raw Key-Value client can be safely and concurrently accessed by multiple threads. Therefore, for one process, one client is enough generally.
```
Hoverbear: "...a client of the TiKV server" I feel this is obvious. Maybe it can be removed.

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,398 @@
...
+
+    Like the Raw Key-Value client, a Client is client to a TiKV cluster.
+
+- Snapshot
+
+    A Snapshot is the state of the TiKV data at a particular point of time, which provides some readonly methods. The multiple reads of the same Snapshot is guaranteed consistent.
+
+- Transaction
+
+    Like the Transaction in SQL, a Transaction symbolizes a series of read and write operations performed within the Client. Internally, a Transaction consists of a Snapshot for reads, and a buffer for all writes. The default isolation level of a Transaction is Snapshot Isolation.
```
Hoverbear:
```suggestion
    Like Transactions in SQL, a Transaction in TiKV symbolizes a series of read and write operations performed within the service. Internally, a Transaction consists of a Snapshot for reads, and a buffer for all writes. The default isolation level of a Transaction is Snapshot Isolation.
```

text/2018-10-25-tikv-client-rust.md
Hoverbear: I don't have a strong opinion either way.

text/2018-10-25-tikv-client-rust.md
Hoverbear: I was wondering this as well. Maybe @gregwebs has some feedback?

text/2018-10-25-tikv-client-rust.md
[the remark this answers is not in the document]
Hoverbear: I think the `AFoo: Foo` semantic in general is not something I've seen much in Rust, and it reads quite strange.

text/2018-10-25-tikv-client-rust.md
sunxiaoguang: Since they are builder of real requests which tends to be verbose and less interesting. I would prefer to keep them in real API document.

text/2018-10-25-tikv-client-rust.md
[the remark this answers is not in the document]
sunxiaoguang: I changed them to Impl: Client. Maybe that looks better and Impl is widely used to represent an implementation of something which could be appropriate for this scenario.

text/2018-10-25-tikv-client-rust.md
sunxiaoguang: Fixed

text/2018-10-25-tikv-client-rust.md
sunxiaoguang: Added, PTAL @Hoverbear

text/2018-10-25-tikv-client-rust.md
sunxiaoguang: Removed

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,398 @@
...
+    let client = RawClient::new(&config);
+    ```
+
+3. Call the Raw Key-Value client methods to access the data on TiKV. The Raw Key-Value API contains following methods
+
+    ```rust
+    pub trait Client {
+        type Impl: Client;
+
+        fn get(&self, key: impl AsRef<Key>) -> Get<Self::Impl>;
```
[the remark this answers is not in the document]
Hoverbear: It's a good point.

You can see:

```rust
trait Client {
    type Impl: Client;
    
    fn get(&self, key: impl AsRef<str>) -> Get<Self::Impl>;
}

struct SomeClient {}
struct Get<C: Client> {
    client: C,
}

impl Client for SomeClient {
    type Impl = Self; // ???
    fn get(&self, key: impl AsRef<str>) -> Get<Self::Impl> {
        // Here we must return an instance of `Self`, but there is no bound saying we can clone this.
        // Even if there was, is it a good idea to clone the entire client?
        Get { client: *self }
    }
}
```

This is part of the reason when I made my DigitalOcean client's builders actually take the `Client` on the [final call to run the request](https://github.com/Hoverbear/digitalocean/blob/846d59ded3f2133f7d00f280d6429af4bb08f6a3/src/request.rs#L111-L166) (Or let them pass directly to [the client via ` client.*()` functions](https://github.com/Hoverbear/digitalocean/blob/846d59ded3f2133f7d00f280d6429af4bb08f6a3/src/client/reqwest.rs))

Hoverbear:
Since we have some final feedback which may impact the API I'm putting merging this on hold until it is addressed. :)

sunxiaoguang:
> Keyspace would make more sense than TiKV cluster. That is, if your program owns a key prefix I assume it is fine to use the Raw API with that key prefix. That reminds me that it is very useful for the API to have a concept of a prefix. But I think we can add this in a future revision and probably in a non-breaking way.

Right, it should be possible if people use it carefully. 

Talking about prefix, I like this idea. Just like chroot concept in many systems. How about we add it now? It could be a very simple change to Config. What do you think @Hoverbear @siddontang @overvenus ?

gregwebs:
This is an important concept, but I would still recommend that key prefixes be tackled in a next iteration because it requires some study and though.

FoundationDB is where I have seen this idea in use the most and we would have to verify whether those concepts map well to TiKV. Notably they recommend usage of a directory abstraction that has logical hierarchies that is represented differently. I think we would want to spend some time studying this and make a separate RFC for adding prefix related abstractions.

FoundationDB links
* [Namespace](https://apple.github.io/foundationdb/developer-guide.html#developer-guide-namespace-management)
* [Subspace](https://apple.github.io/foundationdb/developer-guide.html#developer-guide-subspaces)
* [Directory](https://apple.github.io/foundationdb/developer-guide.html#developer-guide-directories)

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,391 @@
...
+
+```bash
+(Key([107, 101, 121, 49]), Value([118, 97, 108, 117, 101, 49]))
+(Key([107, 101, 121, 49]), Value([118, 97, 108, 117, 101, 49]))
+(Key([107, 101, 121, 49]), Value([118, 97, 108, 117, 101, 49]))
+```
+
+## Programming model
+
+The client instance is thread safe and all the interfaces return futures so users can use the client in either asynchronous or synchronous way. A dedicated event loop thread is created at a per client instance basis to drive the reactor to make progress.
```
overvenus: > The client instance is thread safe ...

Being thread safe is kind of expensive. If the client has some mutable state we may end up with a `Mutex` inside and every APIs require the `Mutex`. I suggest the client should impls `Send + Clone`(another concurrency schema but more flexible). It allows private mutable state, eg, we change client config on the fly.

>  dedicated event loop thread is created at a per client instance basis to drive the reactor to make progress.

I suggest we should not specify thread model, because it is implementation detail. We should leave some space to make sure we dont back ourselves into a corner.

------------------

Based on the model above, here is the code I (as a user) can imagine:

```rust
let client = Arc::new(TikvClient::new());
for _ in 0..4 { // My app has 4 workers.
  let cli = client.clone();
  thread::spawn(move || {
    loop {
      let resp = cli.request().wait();
      handle(resp);
    }
  }).unwrap();
}
```

Users may always wrap the client in an `Arc` to avoid redundant event loop threads.

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,391 @@
...
+
+    ```rust
+    let client = raw::Client::new(&config);
+    ```
+
+3. Call the Raw Key-Value client methods to access the data on TiKV. The Raw Key-Value API contains following methods
+
+    ```rust
+    impl Client {
+        pub fn new(_config: &Config) -> impl Future<Item = Self, Error = Error> + Send;
```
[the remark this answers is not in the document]
sunxiaoguang: But we need to use nightly toolchain to use it at this time, am I right?. However it's indeed good so we don't need to change API when future 1.0 is released.

text/2018-10-25-tikv-client-rust.md
sunxiaoguang: The reason that there is one dedicated thread is because the tso interface of PD is a bidirectional stream. Using a dedicated thread is a way I figured out to keep the order of responses the same as requests yet can still do batching. Not sure if there are other ways to implement that.

Agree with Send + Clone part. It actually has methods that takes `&mut self` so I think it is not correct to say client is thread safe. Let me fix it.

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,391 @@
...
+
+## Tooling
+
+The `tikv_client` crate will be tested with Travis CI using Rust's standard testing framework. We will also include benchmark with criterion in the future. For public functions which process user input, we will seek to use fuzz testing such as `quickcheck` to find subtle bugs.
+
+The CI will validate all code is warning free, passes `rustfmt`, and passes a `clippy` check without lint warnings.
+
+All code that reaches the `master` branch should not output errors when the following commands are run:
+
+    ```shell
```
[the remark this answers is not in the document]
sunxiaoguang: Fixed

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,391 @@
...
+    ```
+
+4. (Optional) Modify data using a Transaction.
+
+    The lifecycle of a Transaction is: _begin → {get, set, delete, scan} → {commit, rollback}_.
+
+5. Call the Transactional Key-Value API's methods to access the data on TiKV. The Transactional Key-Value API contains the following most commonly used methods:
+
+    ```rust
+    pub fn begin(&self) -> Transaction;
```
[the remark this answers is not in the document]
sunxiaoguang: The transactional api document is confusing. I should put everything here instead of just methods from different types. Let me fix it.

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,391 @@
...
+
+    pub fn get(&self, key: impl AsRef<Key>) -> impl Future<Item = Value, Error = Error> + Send;
```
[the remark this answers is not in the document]
sunxiaoguang: Make sense. Will add a concrete type.

text/2018-10-25-tikv-client-rust.md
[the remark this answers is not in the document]
gregwebs: I am wondering how useful it is here to have a Future (you can't do much until you have connected) or if we will end up with a convenience method that blocks.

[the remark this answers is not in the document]
Hoverbear: We can support this later if it becomes a need. :)

text/2018-10-25-tikv-client-rust.md
sunxiaoguang: How about we remove this section for now and let it to be determined? We might figure out a better way later on.

overvenus: Yes, I agree.

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,404 @@
...
+
+## Supported targets
+
+We will target the `stable` channel of Rust starting in the Rust 2018 edition. We choose to begin with Rust 2018 so we do not need to concern ourselves with an upgrade path.
+
+We will also support the most recent `nightly` version of Rust, but users should not feel the need to reach for nightly unless they are already using it.
+
+## Naming
+
+The [Rust API Guidelines](https://rust-lang-nursery.github.io/api-guidelines/naming.html) do not perscribe any particular crate name convention.
```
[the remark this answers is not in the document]
sunxiaoguang: Fixed, Thank you very much.

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,404 @@
...
+
+To use the Raw Key-Value API, take the following steps:
+
+1. Create an instance of Config to specify endpoints of PD (Placement Driver) and optional security config
+
+    ```rust
+    let config = Config::new(vec!["127.0.0.1:3379"]).with_security(
+        PathBuf::from("/path/to/ca.pem"),
+        PathBuf::from("/path/to/client.pem"),
+        PathBuf::from("/path/to/client-key.pem"),
```
sticnarf: Must we use `PathBuf`s in the security config? Although the config is initialized only once in almost all cases, users might not want to care about the type conversion. What's worse is that this introduces needless copies and allocations when we only have a reference to the path (like the example here).

However, I cannot come up with an alternative elegant enough. The function can receive `AsRef<Path>`s, but we should allow that the types of the parameters are different. Thus, we need three type parameters. It might make the signature of the function ugly. Otherwise, we can ask users to pass `&Path`s in, which looks like the usage her, but require the users to do the conversion themselves.

sunxiaoguang: The type of arguments is `impl Into<PathBuf>`, so you can actually write it as `with_security("/path/to/ca.pem", "/path/to/client.pem", "/path/to/client-key.pem")`

sticnarf: Ah , I forgot the impl Trait syntax. But why `Into<PathBuf>` instead of `AsRef<Path>`?

sunxiaoguang: Config might be cloned and copied to underlying code. Therefore it owns data and PathBuf is what we want.

sunxiaoguang: BTW. In case you didn't know. The latest discussion about public api is at [here](https://github.com/tikv/client-rust/pull/10)

sticnarf: Really sorry for missing that PR.

Having looked at that PR, I still doubt about the necessity of cloning the content of the config. However, it is reasonable that `Config` might be cloned.

So, I would recommend to read the content of the files on creating `Config` and have `Config` own `Arc`s of the contents. And, will it become clearer to have a separate struct to save the credentials? So it will look like:
```rust
struct Credentials {
    ca: Vec<u8>,
    cert: Vec<u8>,
    key: Vec<u8>,
}

struct Config {
    pd_endpoints: Arc<Vec<String>>,
    credentials: Option<Arc<Credentials>>,
}
```

What do you think?

sunxiaoguang: It's probably not a good idea to store key directly in memory.
Talking about Credentials struct, if users don't need to know it why bother adding it?

sticnarf: Agree with you. It increases the risk of exposing keys through memory vulnerabilities (like Heartbleed or even Spectre and Meltdown) indeed(, although possibility is decreased by using Rust). 

The `Credentials` struct is not meant to be public to users. I just consider it an tiny improvement on the implementation. With this, we need only one pattern matching for the `Option` without extra `unwrap`s. (So perhaps the discussion about this should be inside of `client-rust` repo instead of this?)

sunxiaoguang: Oh, I see what you mean. You can always do this.
```rust
    if let (Some(ca), Some(cert), Some(key)) = (c.ca, c.cert, c.key) {
        println!("sec");
    } else {
        println!("nope");
    }
```

sunxiaoguang: I guess implementation detail discussion can go to client-rust repository.

sticnarf: Neat solution. No doubts about your design now. Thanks!

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,480 @@
+# Summary
+
+Introduce a full featured, official TiKV (and PD) Rust client. It is intended
+to be used as a reference implementation, or to provide C-compatible binding
```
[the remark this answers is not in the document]
sunxiaoguang: Fixed

text/2018-10-25-tikv-client-rust.md
```diff
@@ -0,0 +1,480 @@
...
+    ```rust
+    let client = raw::Client::new(&config);
+    ```
+
+3. Call the Raw Key-Value client methods to access the data on TiKV. The Raw
+Key-Value API contains following methods
+
+    ```rust
+    impl Client {
+        pub fn new(_config: &Config) -> Connect;
```
sticnarf: One more question about `Config`: Is it certern that `Connect` takes the ownership of a `Config`? According to [the API Guidelines](https://rust-lang-nursery.github.io/api-guidelines/flexibility.html#caller-decides-where-to-copy-and-place-data-c-caller-control), if we use the config as owned here, a `Config` should be taken instead of a reference.

sunxiaoguang: It's not decided yet. We'll see

Hoverbear:
Hi friends! Are there any remaining points anyone would like to see addressed in this PR? We will likely amend this RFC in the future to reflect the final function signatures etc in the implemented code. That will be a different PR.

Hoverbear:
Now that we've seen https://github.com/tikv/client-rust/pull/14 merged we can revisit & update this document.

text/2018-10-25-tikv-client-rust.md
Hoverbear: See https://github.com/tikv/client-rust/pull/33

Hoverbear:
Trying to resolve the DCO problem on this PR. :)

zhangjinpeng1987:
@sunxiaoguang Seems this RFC is out-of-date with existing tikv/rust-client, would you mind to update it or just clost it?

PR #19, opened 2018-12-19
Introduce markdownlint

breeswish:
We need AST based lints.

### DIFF
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -1,0 +1,7 @@
+{
+  "default": true,
+  "MD013": {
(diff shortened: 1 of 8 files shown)


breeswish:
@siddontang @overvenus PTAL, thanks!

.markdownlint.json
```diff
@@ -0,0 +1,7 @@
+{
```
overvenus: Can it check columns wide?

breeswish: Yes! Its default lint rule is that max line width = 80.

PR #21, opened 2019-02-26
TiKV Command Line RFC

Hoverbear:
Based on #7 we discussed having a command line client.

This RFC proposes one! How exciting.


Hoverbear:
@sunxiaoguang can you take a look since it's based off your RFC? :)

sunxiaoguang:
> @sunxiaoguang can you take a look since it's based off your RFC? :)

Sure. I will have a look in a moment.

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,208 @@
...
+    { key: "cncf", value: "linux foundation" },
+    { key: "tikv", value: "rust" },
+]
+Finished in 0.003s. (txn ts: 1551216224838)
+> commit
+Finished in 0.001s. (txn ts: 1551216224838)
+```
+
+One-off transactional mode is a bit different, as we need some way to maintain
+timestamp across commands. The user may choose how to do this, the reccomended
```
sunxiaoguang: recommended

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,208 @@
...
+    { key: "tikv", value: "rust" },
+]
+Finished in 0.003s. (txn ts: 1551216224838)
+$ tikv-cli txn ${TS} commit
+Finished in 0.004s. (txn ts: 1551216224838)
+```
+
+## Drawbacks
+
+We may not wish to need to maintain this tool, or we may wish for the community
```
sunxiaoguang: 'wish to' or 'need to' ?

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,208 @@
...
+
+## Alternatives
+
+* *(Installation/Repo)* We could use the `tikv-client` repository and provide
+  either an opt-out or an opt-in binary from `src/bin/tikv-cli.rs`. If we chose
+  an opt-in functionality here users mgiht end up with an awkard installation
+  call.
+* (*Use `tikv-cli`*), we could use `tikv-cli` for this, but it creates as fuzzy
+  distinction between management tools and clients.
+* (*No REPL, or only REPL*). We may chose to not have a REPL mode, or we may chose
```
sunxiaoguang: I rarely use one off mode of mysql client. But I think running commands from file is very useful.

Hoverbear: Is there an advantage to just piping to the REPL?

sunxiaoguang: Hm, it actually doesn't make much difference. It's just something that I would expect to have as it is in mysql and hive client.

text/2019-02-26-tikv-client-command-line.md
Hoverbear: Oh, damn. Thanks.

text/2019-02-26-tikv-client-command-line.md
Hoverbear: I think I can reword this, but I meant to say "It increases maintenance burden"...

text/2019-02-26-tikv-client-command-line.md
Hoverbear: What do you think @siddontang ?

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,208 @@
...
+cert_path = "./rsa.cert"
+# The path of the key file for TLS.
+key_path = "./rsa.key"
+
+[output]
+# If the ouput JSON should be minified.
+minify = false
+# Level of logging to use. Available: error. warn, info, debug, trace
+logging = "info"
+```
```
nrc: What are the values here? I expected them to be defaults, but that doesn't match the text. (I would also expect the default log level to be `error`, or possibly `warning`.

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,208 @@
...
+By default the tool will output in prettyprinted JSON. For single entry results,
+it will return a string (including `"`), for key-value pairs it will use a JSON
+object `{ key: "", value: "" }`, for multiple results it will return a JSON
+array. Errors will use an `{ error: "" }` object. Opting out of prettyprint will
+be via the `--minify` flag.
+
+JSON was chosen because it is an efficient, pervasive, human-readable, modern
+standard that has many tools (eg `jq`) to use for post processing if needed.
+
+All standard (non-logging) output will be on `stdout`.
```
nrc: To clarify, "standard (non-logging) output" means the json stuff above? It would be good to mention that the total output on stdout will be valid JSON (will this be the case for stderr too?).

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,208 @@
...
+$ tikv-cli raw get "cncf"
+"linux foundation"
+Finished in 0.002s.
+$ tikv-cli raw scan "b"..
+[
+    { key: "cncf", value: "linux foundation" },
+    { key: "tikv", value: "rust" },
+]
+Finished in 0.003s.
+```
```
nrc: These "Finished in 0.003s.", will they be on stderr or stdout? Are they configurable?

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,208 @@
...
+
+The tool will exit zero on successful requests, and non-zero on unsuccessful
+requests. In the case of REPL mode, it shall always exit zero.
+
+### Interacting with it
+
+Users will interact with the tool through the command line. It will support both
+a one-off mode or a REPL (Read-eval-print-loop) style mode (See alternatives).
+
+Example of the one-off raw mode:
```
nrc: Could you explain what "raw" means before using it here.

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,208 @@
...
+[
+    { key: "cncf", value: "linux foundation" },
+    { key: "tikv", value: "rust" },
+]
+Finished in 0.003s.
+```
+
+REPL transactional mode, notice how the `>>` communicates that the user is in a
+transaction. Also notice how the transaction timestamp is outputted in the
+request complete messages.
```
nrc: Why do we need a transactional one-off mode? Could we take multiple commands as input and thus avoid making a distinction between raw and txn modes?

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,208 @@
...
+]
+Finished in 0.003s. (txn ts: 1551216224838)
+$ tikv-cli txn ${TS} commit
+Finished in 0.004s. (txn ts: 1551216224838)
+```
+
+## Drawbacks
+
+We may not wish to need to maintain this tool, or we may wish for the community
+to maintain it.
```
nrc: I think the drawback here is the maintenance burden. We'll need to find some way to handle that. I don't think we need to specify how we handle that here, just be aware that there is this cost.

text/2019-02-26-tikv-client-command-line.md
nrc: I think it is good to have both, and I don't think it adds too much complexity (as long as we can remove the need for raw and txn modes as described above).

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,208 @@
...
+  either an opt-out or an opt-in binary from `src/bin/tikv-cli.rs`. If we chose
+  an opt-in functionality here users mgiht end up with an awkard installation
+  call.
+* (*Use `tikv-cli`*), we could use `tikv-cli` for this, but it creates as fuzzy
+  distinction between management tools and clients.
+* (*No REPL, or only REPL*). We may chose to not have a REPL mode, or we may chose
+  to *only* have a REPL mode.
+* (*A different timestamp handling for one-off mode*). We may find the way we
+  handle timestamps in transactional one-off mode to be unwieldly. Perhaps there
+  is a better way.
```
nrc: Is this standard in other clients? (MySQL in particular). If not, I would think some way of handling multiple commands at once would work. I guess the use case we want to handle is a script which adds commands to a transaction based on the results of earlier commands in the transaction. Unless there is some clear need for that, then I'd imagine that use case is better supported using an API and we should explicitly not support it.

text/2019-02-26-tikv-client-command-line.md
Hoverbear: Should be all of these: https://docs.rs/slog/2.4.1/slog/enum.Level.html, good catch.

text/2019-02-26-tikv-client-command-line.md
Hoverbear: I think these should be stderr.

text/2019-02-26-tikv-client-command-line.md
Hoverbear: In theory transactional one-off won't be used much.

text/2019-02-26-tikv-client-command-line.md
Hoverbear: Part of the reason it's defined this way is that we don't have a text representation (Eg `BEGIN SELECT * FROM foo; ELECT * FROM bar; COMMIT`)

The only other idea I had to support this use case is `tikv-client txn set foo bar then get bar` or something like that and use a seperator... But I'm not sure this is nicer.

text/2019-02-26-tikv-client-command-line.md
Hoverbear: Does the sentence present not communicate this correctly? Because that's exactly what I meant to say. :)

text/2019-02-26-tikv-client-command-line.md
Hoverbear: Yes the intention is to make sure the output is fairly consistent on stdout so folks can use it in scripts.

Hoverbear:
After some discussion with @nrc we think that it is appropriate to default to `transaction` in most cases. So perhaps we should default to `transaction` and allow the user to configure it via the config with `--mode`

text/2019-02-26-tikv-client-command-line.md
Hoverbear: Added a switch for these

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,232 @@
+# TiKV Rust Client Command Line Tool
+
+## Summary
+
+Use the TiKV Rust Client to build a simple command line tool for inspection and anaylsis.
```
dcalvin:
```suggestion
This article describes how to the TiKV Rust Client to build a simple command line tool for inspection and analysis.
```

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,232 @@
...
+
+## Summary
+
+Use the TiKV Rust Client to build a simple command line tool for inspection and anaylsis.
+
+This tool will stand as an analogue to `mycli` or `mysql-client` for users.
+
+## Motivation
+
+Currently `tikv-ctl` offers some ways to interact with a TiKV cluster, it
```
dcalvin:
```suggestion
Currently `tikv-ctl` offers some ways to interact with a TiKV cluster, which
```

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,232 @@
...
+focuses mostly on the *management* of a TiKV cluster instead of a *client* for
+the cluster.
+
+With the `tikv-client` Rust crate users will find a way to interact with their TiKV
```
dcalvin:
```suggestion
With the `tikv-client` Rust crate, users will find a way to interact with their TiKV
```

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,232 @@
...
+cluster from their application. Some users will find it desirable to be able to
+inspect their cluster from a command line tool, or otherwise outside of their
+application.
+
+Building a simple command line tool to execute commands such as `get`, `set`,
+`delete`, and `scan` would mean users do not need to worry about creating such
```
dcalvin:
```suggestion
`delete`, or `scan` means users do not need to worry about creating such
```

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,232 @@
...
+tools for themselves.
+
+It also provides us with a useful tool during the development cycle, as it will
+be very easy to issue one-off commands to TiKV.
+
+Further, we mentioned in the Rust Client RFC
```
dcalvin:
```suggestion
Plus, this tool is one of the goals we mentioned in the Rust Client RFC
```

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,232 @@
...
+(https://github.com/tikv/rfcs/pull/7) we mentioned we wanted to accomplish this.
```
dcalvin:
```suggestion
(https://github.com/tikv/rfcs/pull/7).
```

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,232 @@
...
+
+## Detailed design
+
+The client will be created using the `tikv-client` crate and its already existing
```
dcalvin:
```suggestion
The client will be created using the `tikv-client` crate and existing
```

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,232 @@
...
+functionality. It will utilize `clap` to provide a simple wrapper around the
```
dcalvin:
```suggestion
functionality of the crate.  `clap` will be used to provide a simple wrapper around the
```

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,232 @@
...
+crate API.
+
+### Install
+
+We will build this tool in the `tikv-cli` repository. It will be published.
```
dcalvin:
```suggestion
This tool will be built in the `tikv-cli` repository and published.
```

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,232 @@
...
+Users will be able to install it with `cargo install tikv-cli`. (See
+*Alternatives* for another option.)
+
+### Configuration
+
+The tool will retrieve its configuration from one of the following places, in
+order of precedence (greatest to least):
```
dcalvin:
```suggestion
an order of precedence (greatest to least):
```

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,232 @@
...
+
+* The command line options, such as `--pd 127.0.0.1:2379`.
```
dcalvin:
```suggestion
* Command line options, such as `--pd 127.0.0.1:2379`.
```

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,232 @@
...
+* A local configuration file `./.tikv-cli.toml`
+* The global configuration file `${XDG_CONFIG_HOME}/tikv-cli.toml`.
```
dcalvin:
```suggestion
* A global configuration file `${XDG_CONFIG_HOME}/tikv-cli.toml`.
```

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,232 @@
...
+* The default options (connect to localhost, no TLS).
```
dcalvin:
```suggestion
* Default options (connecting to localhost, no TLS).
```

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,232 @@
...
+# Default: true
+output-durations = true
+```
+
+On the command line, these options have the same name, but with `_` replaced
+with `-` for idiomatic reasons.
+
+### Input/Output
+
+By default input is by positional arguments in `bash`/`sh` string format.
```
dcalvin: by positional arguments？

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,232 @@
...
+
+By default the tool will output in prettyprinted JSON. For single entry results,
+it will return a string (including `"`), for key-value pairs it will use a JSON
+object `{ key: "", value: "" }`, for multiple results it will return a JSON
+array. Errors will use an `{ error: "" }` object. Opting out of prettyprint will
+be via the `--minify` flag.
+
+JSON was chosen because it is an efficient, pervasive, human-readable, modern
```
dcalvin:
```suggestion
JSON is chosen because it is an efficient, pervasive, human-readable, modern
```

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,232 @@
...
+
+It will feature configurable logging levels with `slog` which will output to `stderr`.
+
+The tool will exit zero on successful requests, and non-zero on unsuccessful
+requests. In the case of REPL mode, it shall exit zero unless the connection is
+found to be lost.
+
+### Interacting with it
+
+Users will interact with the tool through the command line. It will support both
```
dcalvin:
```suggestion
Users will interact with the tool through command line. It will support both
```

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,232 @@
...
+a one-off mode or a REPL (Read-eval-print-loop) style mode (See alternatives).
```
dcalvin: Since it's the mode name, "a" is not needed here.
```suggestion
one-off mode and REPL (Read-eval-print-loop) style mode (See [Alternatives](#Alternatives)).
```

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,232 @@
...
+
+TiKV has two APIs, *raw* and *transactional*, both are supported. Learn about the
```
dcalvin: "both are supported" where? In this tool?  Seems missing something.

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,232 @@
...
+be via the `--minify` flag.
+
+JSON was chosen because it is an efficient, pervasive, human-readable, modern
+standard that has many tools (eg `jq`) to use for post processing if needed.
+
+All standard (non-logging) output will be on `stdout`.
+
+It will feature configurable logging levels with `slog` which will output to `stderr`.
+
+The tool will exit zero on successful requests, and non-zero on unsuccessful
```
dcalvin: Can you explain exit zero to me?

text/2019-02-26-tikv-client-command-line.md
Hoverbear: This is an RFC, so this is a proposed action, not an instruction guide for someone. :) Thanks for the analysis correction!

text/2019-02-26-tikv-client-command-line.md
Hoverbear: "in an order of precedence" sounds like there are several orders instead of just one.

text/2019-02-26-tikv-client-command-line.md
Hoverbear: Yes, eg

```bash
tikv-cli get foo
         ^--^--- Positional Arguments
```

text/2019-02-26-tikv-client-command-line.md
Hoverbear: When a program finishes it returns an exit code, if it is zero it is successful, if it is not zero it is generally not successful. https://en.wikipedia.org/wiki/Exit_status

I clarified it.

Hoverbear:
@sunxiaoguang Do you think you could lend your eyes again? It should be ready.

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,232 @@
...
+```
+
+On the command line, these options have the same name, but with `_` replaced
+with `-` for idiomatic reasons.
+
+### Input/Output
+
+By default input is by positional arguments in `bash`/`sh` string format.
+
+By default the tool will output in prettyprinted JSON. For single entry results,
```
AndreMouche:
```suggestion
By default, the tool will output in pretty printed JSON. For single entry results,
```

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,232 @@
...
+"linux foundation"
+Finished in 0.001s. (txn ts: 1551216224838)
+>> scan "b"..
+[
+    { key: "cncf", value: "linux foundation" },
+    { key: "tikv", value: "rust" },
+]
+Finished in 0.003s. (txn ts: 1551216224838)
+>> commit
+Finished in 0.001s. (txn ts: 1551216224838)
```
AndreMouche: Could we print the `commit_ts` too?

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,232 @@
...
+```
+
+One-off transactional mode is a bit different, a series of commands can be
```
AndreMouche: I think `Auto-commit mode` will be more understandable for me ^^

Hoverbear:
@AndreMouche We have that at https://github.com/tikv/client-rust already. :)

text/2019-02-26-tikv-client-command-line.md
Hoverbear: Always or configurable?

brson:
Sweet idea, and fun.

A repl requires building a language and interpreter and there will probably be ongoing feature requests to expand its capabilities. It might be easier to create python bindings and reuse python's repl. Or maybe some other scripting language that's easy on the eyes.

Hoverbear:
😵 Then it's annoying to install and build. It's much easier to distribute a single musl binary or allow a simple `cargo install` instead of having to mess around with system packages and pip.

I suspect it won't be terrifically hard to create a little repl with clap's `arg_matches` and a loop. Might try to whip up a demo...

text/2019-02-26-tikv-client-command-line.md
AndreMouche: both are ok for me ^_^

text/2019-02-26-tikv-client-command-line.md
Hoverbear: I clarified the description and added that term :)

Hoverbear:
@brson Here's a simple REPL example for interactive mode. It'll parse first token (blah, "blah blah", 123) as the KEY, optionallly the second as the VALUE using the normal clap parser and a simple lexer and readline interface:

```rust
use clap::{App, Arg};
use std::env::args_os;
use linefeed::{Interface, ReadResult};
use scanlex::{Scanner, Token};

fn main() {
    let repl_interface = Interface::new("tikv").unwrap();
    repl_interface.set_prompt("tikv > ").unwrap();

    let mut app = App::new("get")
        .arg(Arg::with_name("KEY")
            .required(true))
        .arg(Arg::with_name("VALUE")
            .required(false));

    let matches = app.get_matches_from_safe_borrow(args_os());
    println!("{:?}", matches);

    loop {
        if let Some(res) = repl_interface.read_line_step(None).unwrap() {
            match res {
                ReadResult::Input(input) => {
                    let scanner = Scanner::new(&input);

                    let tokens = scanner.map(|v| match v {
                        Token::Iden(x) | Token::Str(x) => x,
                        Token::Int(y) => format!("{}", y),
                        _ => panic!("No arg"),
                    });

                    let preface = vec![String::from("tikv")];
                    let iter = preface.into_iter().chain(tokens.into_iter());

                    let matches = app.get_matches_from_safe_borrow(iter);
                    println!("{:?}", matches);
                }
                ReadResult::Eof => {
                    break;
                },
                _ => {}
            }
        }
    }
    println!("{:?}", matches);
}
```

You can have a try and see that it feels fine to use and doesn't require a lot of work.

brson:
> Then it's annoying to install and build. It's much easier to distribute a single musl binary or allow a simple `cargo install` instead of having to mess around with system packages and pip. Besides! Python bindings are a whole other project.

That's a problem with Python specifically, yes. There may be other suitable languages.

ice1000:
For REPL I'll recommend rustyline which can do completion.

Hoverbear:
Hi @sciamp-dev, thanks for volunteering to get involved with this project! We're really excited to learn with you and I'm looking forward to working with you directly.

I know you've already had a look at this RFC, but could you please verify it and give your feedback/approval?

@zhangjinpeng1987 @brson do you think you could take a look at this project and give feedback so our friend can get started? @siddontang do you think you could help me open a `client-cli` repo for this and give me maintainer status? I'll work with @sciamp-dev to create an initial implementation in the coming weeks.

sciamp-dev:
Hi everyone, I'm exited to work with you, too! And thanks @Hoverbear :)

I agree with the use of RustyLine for REPL mode, reading the docs it looks more complete and it has more functionalities then other crates like liner or linefeed.
As for the idea of making Python bindings, if there's a solution in Rust I think it's better and more user-friendly.

ice1000:
I'd say making use of some existing Rust-implemented scripting languages and add binding for them or create our own PL are both good ideas.

brson:
I think somebody should go ahead and get started &mdash; I'd rather have code written than not. Whatever that tool looks like will inevitably evolve and undoubtedly end up different. I would not be surprised if the first iteration ends up scrapped, redesigned and rewritten with lessons learned.

So that's what I think should happen, but below are some things to consider.

DSLs have a mixed track record. I'm not though going to seriously advocate binding a scripting language for the DSL. The examples in this RFC are very simple, and if that ends up being the extent of the tool's functionality then a DSL is fine. A sprawling DSL is a poor programming language.

The duality between the one-off mode and the repl probably imposes yet-unknown limits on the grammar.
The one-off mode accepts the tokens of its grammar as individual command line arguments, which has limitations compared to the arbitrary string parsing possible in the repl. Other scripting languages tend to have a flag that just accepts a single string containing script text. With the simple grammar implied here, just statements, keywords, and strings, no variables, no expressions, the duality between cli arguments and tokens looks neat. But that design might not hold up.

A limited language for the one-off mode can be augmented with other tools like `jq`, but that isn't true for the repl. The repl described is not powerful.

Hoverbear:
🎉 @sciamp-dev please review and give feedback/approve this PR, I've reached out to the benevolent rulers of the TiKV org to create the repo for us. :)

sciamp-dev:
Sure @Hoverbear ! I approve this PR, we can go ahead and start.
I agree with what @brson said. I'm doing some quick tests with RustyLine, PyO3(/cpython) and wasm-bindgen for binding with Python and JS. But I guess we'll choose/adapt considering whatever needs we'll have.

Hoverbear:
We (@siddontang , @sciamp-dev ) discussed this on our community chat and decided to start with this being part of the `client-rust` repo and making it an opt-in binary until the project fully matures. This will help it act as a nice test bed for the Rust client, and give it some extra maintenance eyes.

I also corrected some mistakes around `tikv-cli` vs `tikv-client`.

I also added a note about output formats as we do have some commonly used formats and we need to consider our non-UTF-8 users. Please, take a look! :)

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,256 @@
...
+* A global configuration file `${XDG_CONFIG_HOME}/tikv-client.toml` (or
+  comparable cross platform options)
+* Default options (connect to localhost, no TLS).
+
+The options are:
+
+```toml
+[connection]
+# A list of `pd` endpoints.
+# Default: None
```
sciamp-dev: As for line 51 of this RFC, default should be localhost without TLS

```suggestion
# Default: 127.0.0.1:2379 (no TLS)
```

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,256 @@
...
+# Default: transaction
+mode = "transaction"
+# If the tool should output on stderr the duration of commands.
+# Default: true
+output_durations = true
+# Control the encoding.
+# TiKV has a Unified Key Format which is used in cases where a more human readable format is
+# required for non-UTF-8 data.
+# Avaiable: utf-8, ukf, profobuffer, hex
+key-encoding = "utf-8"
```
sciamp-dev:
```suggestion
key_encoding = "utf-8"
```

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,256 @@
...
+value-encoding = "utf-8"
```
sciamp-dev:
```suggestion
value_encoding = "utf-8"
```

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,256 @@
...
+
+It will feature configurable logging levels with `slog` which will output to `stderr`.
+
+The tool will exit with exit code zero on successful requests, and non-zero on unsuccessful
+requests. In the case of REPL mode, it shall exit zero unless the connection is
+found to be lost.
+
+TiKV is often used in cases where a key or value is not a UTF-8 compatible
+string, or may print undesirable control characters to the console. We provide the
+`key-encoding` and `value-encoding` configuration options to control this.
```
sciamp-dev:
```suggestion
`key_encoding` and `value_encoding` configuration options to control this.
```

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,256 @@
...
+
+For most users, their natural first attempt with the client will be to use UTF-8
+strings for both keys and values. In order to prevent confusion, we default to
+UTF-8 for both and make the options for `*-encoding` discoverable.
+
+The `key-encoding` and `value-encoding` options support `utf-8`, `ukf` (TiKV's
```
sciamp-dev:
```suggestion
The `key_encoding` and `value_encoding` options support `utf-8`, `ukf` (TiKV's
```

text/2019-02-26-tikv-client-command-line.md
```diff
@@ -0,0 +1,256 @@
...
+# Available: raw, transaction
+# Default: transaction
+mode = "transaction"
+# If the tool should output on stderr the duration of commands.
+# Default: true
+output_durations = true
+# Control the encoding.
+# TiKV has a Unified Key Format which is used in cases where a more human readable format is
+# required for non-UTF-8 data.
+# Avaiable: utf-8, ukf, profobuffer, hex
```
sciamp-dev:
```suggestion
# Avaiable: utf-8, ukf, protobuffer, hex
```

Hoverbear:
Closing. I am not longer involved with the project and have no stake in this merging. Feel free to create a new PR.

PR #24, opened 2019-03-06
Use pull request ID to number RFCs

### LINKED ISSUE #22: Should we give each RFC an numeric identifier?
I see in other projects' RFCs, they usually mark RFCs with a simple number. Then when they references to a certain RFC, they simply says "RFC #1234". And the number usually is the pull request ID that proposes that RFC. Should we also do like this?

MyonKeminta:
Signed-off-by: MyonKeminta <MyonKeminta@users.noreply.github.com>

Try to number RFCs with their pull request ID, so we can use a simple numeric ID to identify RFCs.

Discussion is needed for this, I think.

Fixes #22


Hoverbear:
I suggest we number them by the order accepted. This way we have a strictly monotonically increasing number.

MyonKeminta:
@Hoverbear It seems that we need to number it after merging the PR? Is that ok?

nrc:
FWIW, Rust moved from a simple count to using the PR number. We found it much more convenient to use the PR number since "RFC 42" is always "RFC 42" before and after it is merged, plus there is less scope for mistakes and a simpler procedure for accepting an RFC.

Hoverbear:
@nrc Ok, let's copy Rust then. Keep things consistent.

README.md
```diff
@@ -12,9 +12,9 @@ stakeholders can be confident about the direction the project is evolving in.
 
 ## How to submit an RFC
 
-1. Copy `template.md` into `text/YYYY-MM-DD-my-feature.md`.
+1. Copy `template.md` into `text/0000-my-feature.md`.
 2. Write the document and fill in the blanks.
-3. Submit a pull request.
+3. Submit a pull request. Then you can change the `0000` in your filename to your pull request ID in another commit.
```
brson: I think this can be a little clearer. In particular the word "can" is too weak. This step is also two steps in one. I was also unsure about the term "pull request __ID__", and from glancing through the GitHub API docs I think the canonical term is "pull request __number__", which I also think is clearer than "ID".

This is also the only indication in the document that RFC numbers are identical to PR ids, but it does not say so explicitly - it says that filenames contain PR Ids. (This point is nitpicky, but in an ideal document there would be a clear statement like "TiKV RFCs are identified by number, and that number is the same as the pull request number for the submitted RFC." That's a bigger change than I would ask for on this PR though.

So I would suggest changing this to instead say more like

`
4. In a new commit, change the `0000` in your filename to be the same as your pull request number, with leading zeros to 4 digits. This is your RFC number.
5. Push the new commit to the same branch to update the pull request with the correct RFC number.
`

tisonkun:
I took over this task at #66 , please take a look :-)

PR #39, opened 2020-01-31
create an RFC for key spaces

gregwebs:
I am still working on this to add some more detail, but hopefully the big picture should be clear.


text/2020-01-30-key-spaces.md
```diff
@@ -0,0 +1,181 @@
...
+TiKV is a monolithic KV store, so each application must choose a key prefix.
+Key Spaces formalizes this process by registering a key space for an application in TiKV itself.
+
+When combined with authentication (not part of this proposal), this can secure an application's key space.
+
+Additionally, key spaces make it possible for TiKV to avoid conflicts during a data restore by changing the key prefix assigned to a key space.
+
+## Motivation
+
+The use case for this is running multiple applications against TiKV (not just a single TiDB).
```
zhangjinpeng1987: I think this is not the recommended way

text/2020-01-30-key-spaces.md
```diff
@@ -0,0 +1,181 @@
...
+Note that we may be motivated to run multiple instances of a single application such as TiDB.
+
+* Avoiding even accidental conflicting use of key space in TiKV is currently ad-hoc and unreliable
+* Securing data in TiKV is difficult. A necessary first step is clearly defining the boundaries of data.
```
zhangjinpeng1987: I think we can define the boudaries now. What is the defficulty?

text/2020-01-30-key-spaces.md
```diff
@@ -0,0 +1,181 @@
...
+I think this is a good idea, but I also think it can simply be proposed independently, although it may be better to discuss after the key spaces RFC is accepted so that it can leverage some of the key space implementation.
+
+This RFC will not discuss authentication.
+Authentication is necessary to provide any meaningful security for key spaces, but I can start a separate RFC for this.
+
+#### Client view
+
+When a client first connects to the TiKV system, it will register itself to get a key space.
+This happens in the metadata component (PD).
+All requests to the storage component (TiKV) will include the key space.
```
zhangjinpeng1987: We know which TiKV the request should be sent to depends on the whole key, the way you proposed will a big challenge for the current routing way.

text/2020-01-30-key-spaces.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Summary
+
+When multiple applications store data in TiKV, we need a clear pattern for avoiding conflicting use of the key space in TiKV.
+TiKV is a monolithic KV store, so each application must choose a key prefix.
+Key Spaces formalizes this process by registering a key space for an application in TiKV itself.
+
+When combined with authentication (not part of this proposal), this can secure an application's key space.
+
+Additionally, key spaces make it possible for TiKV to avoid conflicts during a data restore by changing the key prefix assigned to a key space.
```
zhangjinpeng1987: Can you give a detailed example for restore?

text/2020-01-30-key-spaces.md
```diff
@@ -0,0 +1,181 @@
...
+
+Key Spaces
+
+## Summary
+
+When multiple applications store data in TiKV, we need a clear pattern for avoiding conflicting use of the key space in TiKV.
+TiKV is a monolithic KV store, so each application must choose a key prefix.
+Key Spaces formalizes this process by registering a key space for an application in TiKV itself.
+
+When combined with authentication (not part of this proposal), this can secure an application's key space.
```
zhangjinpeng1987: I' like to know more details about authentication, can you provide an example here?

text/2020-01-30-key-spaces.md
gregwebs: Why?

text/2020-01-30-key-spaces.md
gregwebs: Yes you can define boundaries by writing them down on a piece of paper way but how do you keep it up to date when new applications are created and prevent someone from accessing the key space that belongs to your application in an automated way?

text/2020-01-30-key-spaces.md
gregwebs: But a key space still resolves to a key range. So it should still be equivalent to sending an entire key prefix. What is the challenge?

text/2020-01-30-key-spaces.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Motivation
+
+The use case for this is running multiple applications against TiKV (not just a single TiDB).
+Note that we may be motivated to run multiple instances of a single application such as TiDB.
+
+* Avoiding even accidental conflicting use of key space in TiKV is currently ad-hoc and unreliable
+* Securing data in TiKV is difficult. A necessary first step is clearly defining the boundaries of data.
+
+Once key spaces are in place, all applications and TiKV itself should understand what data belongs to whom and how to avoid key space conflict.
```
disksing: I'd like to propose a client-only alternative. Maybe we can only introduce the concept of keyspace in client side and client takes care of key prefix stuff. I could be much simpler as tikv or pd does not need any change.

gregwebs: We effectively do have client-side key spaces right now. But I guess some clients believe they own the entirety of the key space and this API change would encourage clients to not do that?
But this won't meet my requirements of having an automated way of ensuring there is no key conflict.

From the point of view of an abstract client-side driver these would both look the same where the user configures a namespace. And I would like the client-side to implement a sub spaces feature (spaces with a keyspace) on their own, just as is done in FoundationDB. So what you are suggesting would also be like adding only sub spaces: every client can write to the whole key space but first creates a sub space that it hopes won't conflict.

text/2020-01-30-key-spaces.md
gregwebs: Did you see the section "Backup and Restore" ?

text/2020-01-30-key-spaces.md
zhangjinpeng1987: The request must know the whole key before sending to TiKV

text/2020-01-30-key-spaces.md
zhangjinpeng1987: IMO client-side key space is more flexible, like Redis on TiKV, you can define your own key spaces.

text/2020-01-30-key-spaces.md
zhangjinpeng1987: Can you explain why need to change the prefix?

text/2020-01-30-key-spaces.md
zhangjinpeng1987: Is runing multiple applicactions against TiKV a strong demand?

text/2020-01-30-key-spaces.md
```diff
@@ -0,0 +1,181 @@
...
+This allows TiKV to change the prefix when doing a restore and also allows it to perform key prefix compression.
+Note that some storage engines may already be able to perform prefix compression.
+For backwards compatibility purposes, a client is also allowed to be registered with an explicit key prefix.
+
+#### The default key space
+
+A TiKV client may ignore key spaces. In a fresh TiKV deployment with this feature in place, that client will be using the default key space.
+A configuration option `disable-default-key-space` when set to `true` forces clients to specify a key space rather than using the default key space. 
+
+#### Backwards compatibility
```
zhangjinpeng1987: Is it impact TiDB?

text/2020-01-30-key-spaces.md
gregwebs: For example if someone wanted to combine 2 separate TiDB clusters into one TiKV cluster. They could have a collision which can be ensuring that when restored they use different key prefix

text/2020-01-30-key-spaces.md
gregwebs: Its a chicken and egg problem. This is rarely done now because we lack basic features for it such as the one in the RFC. But users do ask about it.
For TiKV to be usable as a general purpose KV store these kinds of feature must be present. If it were usable as a general purpose KV store, users would start running KV workloads against it.

text/2020-01-30-key-spaces.md
gregwebs: It can since the key space resolves to a prefix. But I assumed PD would tell the client which TiKV to send its requests to. Do we have some documentation on this aspect?

text/2020-01-30-key-spaces.md
gregwebs: Yes

text/2020-01-30-key-spaces.md
gregwebs: How would it be more flexible? Ease of implementation is different than flexibility. An opposite of flexible can be brittle, and it is certainly more brittle because there is no de-coupling between logical and physical key space to prevent collisions: that must be ensured by ad-hoc mechanisms.

I do think we should encourage clients to utilize sub spaces as needed independently of this proposal.

text/2020-01-30-key-spaces.md
zhangjinpeng1987: I hope https://github.com/tikv/client-go/blob/master/locate/region_cache.go can show you some details

text/2020-01-30-key-spaces.md
zhangjinpeng1987: > combine 2 separate TiDB clusters into one TiKV cluster

This sounds interesting, I don't know if this demand exists, @shenli what's your opinion for it.

text/2020-01-30-key-spaces.md
shafreeck: Etcd has the ACL and supports to grant permission for a role on a key prefix. It is an RBAC implementation(IMO). https://github.com/etcd-io/etcd/blob/master/Documentation/op-guide/authentication.md#working-with-roles

shafreeck: Besides, [distributedio/titan](https://github.com/distributedio/titan) uses namespace as the key prefix which combines with the authentication feature: User uses a token which is signed by a key with a namespace to connect to titan.

text/2020-01-30-key-spaces.md
shafreeck: I think there are two points to be discussed here.
* Why the feature is necessary?
People prefer simpler, newer tech-stack and fewer costs to achieve all goals, like using one database for all products. The HA and scalability make TiKV a good solution for some startups, however, they may do not have large amount of data, running multiple instances of TiDB on one TiKV cluster is a natural choice for them.
However, things are always hard, when multiple products rely on one TiKV cluster, data isolation is necessary for data security and resource usage. `Key Spaces` makes the data invisible to each other but it can not isolate the resource like CPU or memory usage. The latter may be solved by virtualization-tech.

* Why the feature should be put on TiKV?
As describes above. It is hard to isolate resource usage on TiKV and maybe the better solution is virtualization(VMs, Containers...). Another choice is to put this feature above TiKV which keeps TiKV simple and clean.

BusyJay:
Key spaces sound like a solution to the problem tikv/tikv#3922. And it may also be useful if we want to implement multi-tenancy in cloud.

text/2020-01-30-key-spaces.md
gregwebs: Resource isolation is already an issue for TiDB users. TiDB itself supports multiple applications. And just storing data in 2 different tables can be a form of multi-tenant data with different workload requirements that impact each other. Even for a single table there can be transnational workloads and analytical workloads at the same time. So resource isolation already needs to be addressed and it is already up to users to determine how to handle this. After this RFC is implemented, it will still be up to users to determine how to achieve the workload isolation they need and balance that with cost and operational complexity. But it will be easier for them to experiment with different topologies.

Putting this feature client-side makes for less work for TiKV developers but nobody has explained what the advantage is otherwise. A client-side implementation will then fail to meet the requirements in the RFC.

gregwebs:
> Key spaces sound like a solution to the problem [tikv/tikv#3922](https://github.com/tikv/tikv/issues/3922).

A key space would be designated as transactional or non-transactional when it is created?

text/2020-01-30-key-spaces.md
gregwebs: Thanks for the etcd link. @shafreeck I would like to find out more about the Titan auth.

text/2020-01-30-key-spaces.md
gregwebs: I still don't understand the issue here since a key space is equivalent to a key prefix. Perhaps we need to develop this proposal to a greater level of detail.

zhangjinpeng1987:
> Key spaces sound like a solution to the problem [tikv/tikv#3922](https://github.com/tikv/tikv/issues/3922). And it may also be useful if we want to implement multi-tenancy in the cloud.

I don't think it is a good solution to issue [tikv/tikv#3922](https://github.com/tikv/tikv/issues/3922). TiKV supports multi-tenancy in the cloud is a big topic, there are many things need to consider, but this RFC is a good beginning.

text/2020-01-30-key-spaces.md
zhangjinpeng1987: In short, client can't request TiKV `Put(keyspace, key, value)` directly, because you don't know which TiKV you should request, instead, you should handle the keyspace logic(maybe need to request PD for some meta-information, this can be handled in tikv-client) at first, and then you know which TiKV you should send the `Put(encoded key, value)` request to.

text/2020-01-30-key-spaces.md
zhangjinpeng1987: For the resource isolation in TiDB, 4.0 PD has a feature named placement rules, it supports balancing 2 different tables to different machines or different zone, this depends on the rule you specified.

text/2020-01-30-key-spaces.md
```diff
@@ -0,0 +1,181 @@
...
+
+For an existing deployment we will need an upgrade process.
+We can have a configuration flag "allow-unmanaged-key-space".
+For a fresh deployment this will be `false`, but if there is an existing deployment it should be `true`.
+When true, clients that do not specify a key space will continue using the key space the same way they do now.
+When false, clients that do not specify a key space will use the default key space.
```
zhangjinpeng1987: What is the prefix key for default key space?

text/2020-01-30-key-spaces.md
```diff
@@ -0,0 +1,181 @@
...
+#### The default key space
+
+A TiKV client may ignore key spaces. In a fresh TiKV deployment with this feature in place, that client will be using the default key space.
+A configuration option `disable-default-key-space` when set to `true` forces clients to specify a key space rather than using the default key space. 
+
+#### Backwards compatibility
+
+For an existing deployment we will need an upgrade process.
+We can have a configuration flag "allow-unmanaged-key-space".
+For a fresh deployment this will be `false`, but if there is an existing deployment it should be `true`.
```
zhangjinpeng1987: For the existing deployment, does this configuration should keep true alway? IMO it should.

text/2020-01-30-key-spaces.md
gregwebs: How about "default"?

zhangjinpeng1987:
> > Key spaces sound like a solution to the problem [tikv/tikv#3922](https://github.com/tikv/tikv/issues/3922). And it may also be useful if we want to implement multi-tenancy in the cloud.
> 
> I don't think it is a good solution to issue [tikv/tikv#3922](https://github.com/tikv/tikv/issues/3922). TiKV supports multi-tenancy in the cloud is a big topic, there are many things need to consider, but this RFC is a good beginning.

I'm very glad to discuss key space under the assumption that we(TiKV/TiDB) are going to support multi-tenancy in cloud. I know we can use key space to do many many other things, but multi-tenancy in cloud is the most attractive one for me(maybe wrong).
If this assumption is true, the most important thing I think is the resource isolation, especially for those users that don't have too much data(but their QPS is not neglectable), we should limit these users' CPU/IO/network usage individually, and the memory cache also has a quote.

gregwebs:
> the most important thing I think is the resource isolation

Yes. Key spaces registered for applications provide a crucial piece of meta information that can be used to help with resource isolation. Identifying the application is required for this. We could solve it in other ways, but here again I think Key Spaces provide the best possible building block. But I think this topic requires an entirely separate RFC, I don't think we will be able to productively discuss it here.

zhangjinpeng1987:
> > the most important thing I think is the resource isolation
> 
> Yes. Key spaces registered for applications provide a crucial piece of meta information that can be used to help with resource isolation. Identifying the application is required for this. We could solve it in other ways, but here again I think Key Spaces provide the best possible building block. But I think this topic requires an entirely separate RFC, I don't think we will be able to productively discuss it here.

Got it.
@sunxiaoguang If TiKV/PD supports key space, I would say it almost supports the table concept in TiKV. I think this is helpful for your TableStore base on TiKV, would you mind to take a look at this RFC? Thank you for advance.

text/2020-01-30-key-spaces.md
gregwebs: Currently the client knows where to send the key because PD tells it? It seems that the client just needs to ask PD what TiKV a (keyspace, key) should be sent to instead?

It would be nice for the client to not have to know the actual prefix. But I think it may also possible for the client to fully resolve the key space to a prefix. When adding on a feature such as auth, TiKV would need to recognize the prefix as belonging to a key space.

text/2020-01-30-key-spaces.md
gregwebs: It must remain true until all applications can be migrated to use key spaces.

text/2020-01-30-key-spaces.md
shafreeck: https://medium.com/@shafreeck/titan-a-distributed-redis-protocol-compatible-nosql-database-4379a1307661

The `Multi-tenancy support` section gives some details.

gregwebs: Thanks, that seems similar to what I came up with for my [auth proposal](https://github.com/tikv/rfcs/pull/40)

gregwebs:
I added an [auth proposal](https://github.com/tikv/rfcs/pull/40). This builds on top of key spaces and hopefully helps to motivate them a bit more, particularly in demonstrating why they can't be implemented client-side.

@zhangjinpeng1987 and others asked about this.

huachaohuang:
Just come across and find out that someone finally proposes keyspaces. Please map each table / index of TiDB into a different keyspace, and allow different keyspaces to use different settings (like the number of replicas, storage options, etc), it will make a lot of things easier.

gregwebs:
@huachaohuang we already have some ability to have different settings for different tables. So I am not sure if keyspaces are needed for that.
Currently I am suggesting that TiDB still manage its table spaces in a single key space (I would call these client-side namespaces) and that key spaces should define the boundaries between applications. But maybe if namespaces were additionally registered in TiKV (still as a separate concept from key spaces) it could help formalize the table concept in TiKV.

huachaohuang:
@gregwebs Mapping a table into a keyspace has some benefits. First of all, we can remove the `t{table_id}_` prefix for record keys and `t{table_id}_i{index_id}_` prefix for index keys. Once we remove these prefixes from keys, we can move tables between clusters. For example, you can merge two small clusters into one without worrying about table id conflicts, because you can simply allocate another keyspace. Otherwise, you need to rewrite all the data just because the table id is changed.

gregwebs:
I see, you want prefix abstraction. The keyspace abstraction in this proposal guarantees no conflicts when importing data by altering the prefix during import, but cannot guarantee no conflicts when live merging. With key spaces in this proposal there is still a concrete prefix that it is chosen at insertion time and used in TiKV storage which is still a single KV monolith. This prefix is still stored in RocksDB.

What you are looking for seems to be a break-up of the TiKV monolith. I can imagine this: data regions have an identity in addition to a prefix and the identity is mapped to a key space and can can be altered. However, this is a fundamental change to how TiKV operates that is beyond the scope of this proposal.

To achieve live-merging with this proposal the best approach is probably to more explicitly manage a global key space. An organization could maintain a global key space registry such that different clusters will have key spaces with non-conflicting prefixes (seed each cluster with a different prefix).

coocood:
@gregwebs 

I have another keyspace design for multi-tenant at https://github.com/ngaut/unistore/pull/393

The key difference is the cluster id is used as tenant ID, and the prefix for tenant id is fixed.

gregwebs:
I think the cluster id could potentially be useful to associate with the default namespace. However, if we attempt to use the cluster id for this we will change the meaning of the cluster id and greatly complicate it so we are better off not attempting to re-use the concept here.

But overall the proposals seem to be the same.

text/2020-01-30-key-spaces.md
gregwebs: I updated the text here to better explain the need for key spaces. Please add your comments back if they still need to be addressed.

text/2020-01-30-key-spaces.md
gregwebs: I explained this a little better in the RFC now.

text/2020-01-30-key-spaces.md
gregwebs: The proposal has been updated so that TiKV does not have knowledge of key spaces, but they are still registered in PD.

text/2020-01-30-key-spaces.md
gregwebs: This section is no longer present. Essentially it is moved to the [key prefix RFC](https://github.com/tikv/rfcs/pull/56).

gregwebs:
This RFC is changed now so that TiKV has no knowledge of key spaces. I created a new RFC for TiKV to assist with sending prefixes that can be used for key spaces: https://github.com/tikv/rfcs/pull/56

text/2020-01-30-key-spaces.md
gregwebs: This RFC is changed now so that TiKV has no knowledge of key spaces. I created a new RFC for TiKV to assist with sending prefixes that can be used for key spaces: #56

text/2020-01-30-key-spaces.md
gregwebs: I created a new RFC for TiKV to assist with sending prefixes that can be used for key spaces: #56. This prefix capability can be used to create client-only key spaces.
This RFC is changed now so that TiKV has no knowledge of key spaces, but they are still registered in PD so that all clients can be aware of other registered key prefixes.

text/2020-01-30-key-spaces.md
```diff
@@ -8,119 +8,115 @@ When multiple applications store data in TiKV, we need a clear pattern for avoid
...
-I suggest that we can add an identity id as an alternative field to an application name. Similarly, the output key space object can include the identity id.
+* prefix
 
-Addtional GET APIs will also return keyspace objects
+This API will return the key space object. The key space name is the id of the keyspace and cannot be changed.
+The usage of an application name as an identity may end up being a de-normalization of data in a future iteration when true identity with authentication is added.
+The prefix that is generated is a binary sequence. This will start with a single byte sequence: this optimizes for the fact that few TiKV installations will have more than 255 applications.
+The sequence starts with 0x01 and increment by 1. The 0x00 prefix will remain reserved.
+A byte sequence containing 0xFF for all bytes is not a valid prefix because this is the end of the range. When a sequence contains just 0xFF, a byte of 0x00 is appended and the byte sequence is now one byte longer. This pattern allows us to start with just a 1 byte sequence but allows expansion to any length.
```
coocood: This is basically a varint encoding, it doesn' solve the problem that a key with shorter prefix may conflict with a key with longer prefix.

coocood: Why not just use fixed 4 bytes, the simplicity benefit overweight the storage cost.

gregwebs: yes this is similar to varint encoding. The difference is that varint encoding reserves an entire bit to indicate more length. This reserves only a single byte value (0xFF) to indicate more length.

Either technique does in fact prevent conflicts. I guess I need to more explicitly state that every 2 byte sequence generated by this pattern will start with 0xFF.

gregwebs: I switched to the varint encoding

gregwebs:
The RFC is updated for changes to the key prefixes specification and clarity. Clients can now specify the prefix they want to register: this helps with backwards compatibility.

text/2020-01-30-key-spaces.md
```diff
@@ -0,0 +1,197 @@
...
+DELETE /keyspaces/{id}
+
+requires id. Deletion is background task. This DELETE API sets the time of the API call in a field `deleted_at`.
+When deletion is completed, it will set a field `delete_completed_at` and the prefix may be re-used in a new key space.
+
+In a token-based auth system (proposed in another RFC) deletion will need to wait until client tokens expire before it can begin deletion.
+
+
+
+#### Key Prefix specification
```
BusyJay, line 102: I suggest to use id instead of bytes prefix. TiKV doesn't have to store the KV with given prefix, it can totally store them in separate CFs. Not using byte prefix gives more flexibility for implementations and also easy to get backward compatible. Client can use the default key space when non is specified.

gregwebs: Column families would lead to increased overhead when a key space is very small? But this is more of an implementation detail that the specification should try to accomodate.

gregwebs: For efficiency we may want tiny key spaces to be stored together but to split larger key spaces into column families or RocksDB instances. Currently we don't store each region in its own column family.

TiKV has no knowledge of key spaces in this proposal. I think that we can extend the PD placement rules (or create something similar) to be able to specify for different key ranges storage properties such as whether to use column families or a separate RocksDB instance. These storage properties would need to be sent to TiKV. TiKV would need to route keys to the correct storage based on the key prefix.

I think prefix matching is equivalent to using an id for a prefix. If key space ids are used and stored in the same column family then they must be translated into a prefix.

gregwebs: Any id would also be a valid binary prefix. The difference between the two is really whether there is a separate field for the prefix/id or whether it is just prepended. I tried to have a separate field for the prefix (sending via a header) in https://github.com/tikv/rfcs/pull/56, but I then noted:

> There are a lot of places where prefixes could now be used. But there are also already a lot of places that assume a full key (with the 'z' first byte prefix) is already being given.

I am also worried about forgetting to send the prefix.

gregwebs: My preference is to use a binary prefix so that we don't require anything new of TiKV and that clients also are not forced to do anything differently. Once auth is enforced clients need to authenticate but otherwise don't have to do anything differently. And any programs that need full key space access (PD, full backup) don't have to do anything differently. We can still take the approach of #56 to allow for sending prefixes separately and variants of that are possible where we use a separate GRPC API field for a key space prefix.

BusyJay (commented):
Rest LGTM. /cc @nolouch PTAL

text/2020-01-30-key-spaces.md
```diff
@@ -0,0 +1,197 @@
...
+Update a key space
+
+requires id and description. Only the description field can be updated.
+
+Addtional GET APIs will also return keyspace objects
+
+* Lookup by ID, key space name
+* List all key spaces
+
+DELETE /keyspaces/{id}
```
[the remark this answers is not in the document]
gregwebs: I am happy with either approach. Probably we should just try to stay consistent with the existing PD APIs.

gregwebs:
I added more backwards compatibility details.

subu-cliqz:
@gregwebs Any idea on how this could work with merge operators / column family ?
https://docs.rs/rocksdb/0.15.0/rocksdb/merge_operator/index.html

gregwebs:
> @gregwebs Any idea on how this could work with merge operators / column family ?
> https://docs.rs/rocksdb/0.15.0/rocksdb/merge_operator/index.html

Do you have any specific concerns?

subu-cliqz:
> > @gregwebs Any idea on how this could work with merge operators / column family ?
> > https://docs.rs/rocksdb/0.15.0/rocksdb/merge_operator/index.html
> 
> Do you have any specific concerns?

Yes. From a brief reading of the RFC, it looks like the idea is to keep all of the keyspace maintenance logic in the PD.

My concern is that the merge operator would have to defined for each column family. At first glance, this looks hard to do without making the application / column family relationship more explicit in TiKV ?

I realize that the `merge` functionality is not supported in TiKV as of now, but I plan on working on that.

Thanks for answering my questions :)

gregwebs:
@subu-cliqz I am glad you are helping TiKV. This proposal does not add new column families so I don't think it would change your work. One piece of feedback was to equate a column family to a keyspace but I do not want to have that in the RFC.

text/2020-01-30-key-spaces.md
jainlavina: This discussion is interesting because we have been looking into TiKV and TiDB and while TiDB has a good set of features, TiKV is really not usable as a standalone K,V store due to lack of this feature. It would be nice to add "tables" or "keyspaces" into TiKV and have TiDB leverage that as well. Those would be similar to how most other K,V stores provide ability to "create table" and then add a key, value to a table without having to deal with the overhead of sql parsing and query planning.

PR #40, opened 2020-02-06
TiKV auth

gregwebs:


text/2020-02-04-auth.md
```diff
@@ -0,0 +1,199 @@
...
+
+
+## Motivation
+
+TiKV has no authentication.
+TiDB supports authentication through TiDB.
+This has been done in the TiDB layer probably in partly because it is a large effort to achieve MySQL compatible auth mechanisms.
+However, TiKV exposes its data over a network connection, whereas MySQL does not.
+
+From a security perspective, this means that any attacker that can achieve network access to a TiKV node can exfiltrate all its data.
```
BusyJay: We supports SSL connections, which can limit access to authorized peers.

gregwebs: Thanks, I updated the RFC to point that out.

text/2020-02-04-auth.md
```diff
@@ -0,0 +1,228 @@
...
+And all these client machines with the ability to access TiKV must be trusted as root users.
+
+
+## Detailed design
+
+#### Key Spaces and user spaces
+
+I first want to establish the concept of a “key space”.
+TiDB is one potential user of TiKV.
+One also might want to run 2 entirely separate TiDB clusters against the same TiKV cluster or run other applications against TiKV.
```
lysu: can we avoid 1 TiDB use 2 "key space"?

gregwebs: I am not sure what this question is asking, perhaps you can add more detail? I would expect 1 TiDB to use 1 key space. This auth proposal would ensure it accesses only its key space.

lysu: Yes, I mean it will be more complex if we have 1 TiDB use multiple TiKV key space(TiDB maybe need choose different user to access different key ranges).

if we expect 1 TiDB to use 1 key space that will be ok :D

gregwebs: If TiDB had multiple key spaces it would use a different key space for each MySQL schema/database or perhaps for every table. But I don't think there is actually any benefit to this approach if the delegation with restrictions that is proposed here is implemented.

I changed the proposal now to only briefly mention the key space proposal.

gregwebs:
I added #46 for implementation details. This PR will remain to provide a motivation and comparison with other possible implementations: I think that information is distracting when one would like to focus on understanding the implementation.

text/2020-02-04-auth.md
```diff
@@ -0,0 +1,206 @@
...
+
+
+#### Shared secret and bootstraping
+
+In the above scenario, TiKV must have a mechanism to be able to decrypt the token and ensure it was not tampered with.
+One way of doing this is to encrypt it with a secret that only TiKV knows.
+At startup, TiKV can get the secret from PD.
+However, we must be able to prevent impersonation of a TiKV node, and we can do this with our existing auth mechanism
+
+TiKV itself must also authenticate.
```
iosmanthus, line 113: This could be done by the `put_store` operation from TiKV to PD. The token could be refreshed by the following heartbeat.

PR #41, opened 2020-02-26
RFC: Key Visualizer: The visualized hotspot profiler

andylokandy:
## Summary

Add an enabled-by-default `KeyVisualizer` component to `PD`, providing a visual webpage to show key heatmap of the `TiKV` cluster, which is useful for troubleshooting and reasoning inefficient application usage patterns. `KeyVisualizer` is accessible in-browser via the URL from `PD`.

## Motivation

At present, if someone wants to troubleshoot a performance issue in `TiKV`, he will usually make use of some existing diagnostic tools, such as `pd-ctl`, `Prometheus` and `Grafana`, which are hard to learn, intuitive, not efficient, unable to recognize the pattern of application and hard to make correct judgment basing on them.

Google `BigTable` has similar problems. So they provided a heatmap profiler called [`BigTable KeyVisualizer`](https://cloud.google.com/bigtable/docs/keyvis-overview) ([Video](https://youtu.be/aKJlghIygQw)), which is recommended that the user first use to troubleshoot performance problems. After investigation, we realized that it has
good features that also suits `TiKV`, thus we added some of those features to our design.

[Rendered](https://github.com/andylokandy/rfcs/blob/master/text/2019-11-21-key-visualizer.md) 

[Rendered (zh-CN)](https://gist.github.com/andylokandy/f4740b28192591b63501297d154717dd)


HunDunDM:
Resubmit in #42

PR #43, opened 2020-03-08
RFC: Using chunk format in coprocessor framework

TennyZhuang:
Signed-off-by: TennyZhuang <zty0826@gmail.com>


text/2020-03-07-copr-chunk.md
```diff
@@ -0,0 +1,150 @@
...
+The [chunk format](https://github.com/pingcap/tidb/blob/master/util/chunk/chunk.go) is a column based memory format and has much more compact memory layout than `VectorValue` with nullable values.
+
+## Motivation
+
+There are multiple benifits to use chunk format during operator evaluation:
+
+1. For many SQL builtin functions, they have the same null handling pattern. If some of input args are NULL, then the result is also NULL. We can call `merge_nulls` with input bitmap to output column before call the actual implementation. Then the actual function can iterate the column without null flag to reach better cache locality.
+2. SIMD instructions require continuous data in memory.
+3. Keep the same format with TiDB, so that it's possible to unify the computing framework between TiDB and TiKV.
+
```
breeswish: 4. Using the Chunk format can improve performance for variable sized types like `Bytes`.

text/2020-03-07-copr-chunk.md
```diff
@@ -0,0 +1,150 @@
...
+pub fn md5(arg: &Option<Bytes>) -> Result<Option<Bytes>> {
+    // ...
+}
+```
+
+to
+
+```rust
+#[rpn_fn]
+pub fn md5(arg: Option<&Bytes>) -> Result<Option<Bytes>> {
```
breeswish: `args: &[u8]`

breeswish: This signature might not be very efficient for returning values in Chunk format, since in order to put these return values into a chunk column, it must be copied again.

text/2020-03-07-copr-chunk.md
```diff
@@ -0,0 +1,150 @@
...
+}
+```
+
+This requires modify all rpn_function implementation, maybe we can find a better solution.
+
+Although we can implement all builtin functions based on the new ScalarRef args, it's not efficient enough. We have to read both bitmap and data map to construct an `Option<&T>`. Instead, for most builtin function, if some of the input args is NULL, then the output will be NULL. So that we can add a new attribute in `rpn_fn` macro.
+
+
+```rust
+#[rpn_fn(merge_nulls = [0])]
```
breeswish: we can make this default. 99% functions have the same NULL behaviour.

text/2020-03-07-copr-chunk.md
```diff
@@ -0,0 +1,150 @@
...
+Using `chunk::Column` format in the `coprocessor` framework instead of `VectorValue`.
+
+The [chunk format](https://github.com/pingcap/tidb/blob/master/util/chunk/chunk.go) is a column based memory format and has much more compact memory layout than `VectorValue` with nullable values.
+
+## Motivation
+
+There are multiple benifits to use chunk format during operator evaluation:
+
+1. For many SQL builtin functions, they have the same null handling pattern. If some of input args are NULL, then the result is also NULL. We can call `merge_nulls` with input bitmap to output column before call the actual implementation. Then the actual function can iterate the column without null flag to reach better cache locality.
+2. SIMD instructions require continuous data in memory.
```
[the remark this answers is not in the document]
TennyZhuang: We can do some experiments about it.

text/2020-03-07-copr-chunk.md
```diff
@@ -0,0 +1,248 @@
...
+```
+
+The input has been optimized to `&[u8]`, but the result should keep `Vec<u8>`.
+
+Although we can implement all builtin functions based on the new ScalarRef args, it's not efficient enough. We have to read both bitmap and data map to construct an `Option<&T>`. Instead, for most builtin function, if some of the input args is NULL, then the output will be NULL. We can handle NULL by default.
+
+
+```rust
+#[rpn_fn]
+pub fn md5(arg: Bytes) -> Result<Option<Vec<u8>>>;
```
breeswish: We should call it BytesRef

text/2020-03-07-copr-chunk.md
```diff
@@ -0,0 +1,248 @@
...
+Writer::<Bytes>;
+Writer::<JsonRef>;
+```
+
+Then the signature can be modified to:
+
+```rust
+#[rpn_fn]
+pub fn md5(arg: Bytes, writer: Writer<Bytes>) -> Result<Guard> {
+    writer.write(hex_digest(MessageDigest::md5(), arg));
```
breeswish: note that in this form, it is possible to first write something, then return an error.

text/2020-03-07-copr-chunk.md
```diff
@@ -0,0 +1,248 @@
...
+#[rpn_fn]
+pub fn md5(arg: Bytes) -> Result<Option<Vec<u8>>>;
+```
+
+We should add new attribute `nullable` in `rpn_fn`, so that some special functions can handle NULL manually.
+
+```rust
+#[rpn_fn(nullable = true)]
+pub fn md5(arg: Option<Bytes>) -> Result<Option<Vec<u8>>>;
+```
```
breeswish: I think this part can be reorganized to be more clear. We can present the design first.

text/2020-03-07-copr-chunk.md
```diff
@@ -0,0 +1,248 @@
+# RFC: Using chunk format in coprocessor framework
+
+## Summary
+
+Using [TiDB Chunk format](https://github.com/pingcap/tidb/blob/9627132db25e5cb560ba6ed89c2e39c32edb267a/util/chunk/chunk.go) in the `coprocessor` framework instead of `VectorValue`.
```
breeswish: Links are not unified, and are duplicated.

text/2020-03-07-copr-chunk.md
```diff
@@ -0,0 +1,248 @@
...
+    // TODO
+}
+
+pub trait FixedSizeTypedVector<T>: TypedVector<T> { /* TODO */ }
+
+pub trait VarSizeTypedVector<T>: TypedVector<T> { /* TODO */ }
+```
+
+
+### Rpn Expression
```
breeswish: This section is too long.. Make more sections.

text/2020-03-07-copr-chunk.md
TennyZhuang: If Err is returned, we should call discard in framework

text/2020-03-07-copr-chunk.md
```diff
@@ -0,0 +1,214 @@
...
+// From
+#[rpn_fn]
+pub fn logical_and(lhs: &Option<i64>, rhs: &Option<i64>) -> Result<Option<i64>>;
+
+// To
+#[rpn_fn]
+pub fn logical_and(lhs: &i64, rhs: &i64) -> Result<Option<i64>>;
+
+// Or
+// NULL will be merged to output by default. specify nullable to handle NULL mannully.
```
codeworm96: `manually`

text/2020-03-07-copr-chunk.md
```diff
@@ -0,0 +1,214 @@
...
+pub trait PartialWriter<T>: Writer<T> {
+    begin(self) -> PartialGuard;
+}
+```
+
+The definition of `Writer` and `PartialWriter` ensures some properties:
+
+1. `Guard` is zero-overhead, and the only way to achieve a `Guard` is calling `write` or `finish`.
+2. `write` can be called at most once in one `Writer`.
+3. Only one of `write` and `begin` can be called at most once in one `PartialWriter`.
```
codeworm96: `called ... on`

text/2020-03-07-copr-chunk.md
```diff
@@ -0,0 +1,214 @@
...
+}
+
+pub trait PartialWriter<T>: Writer<T> {
+    begin(self) -> PartialGuard;
+}
+```
+
+The definition of `Writer` and `PartialWriter` ensures some properties:
+
+1. `Guard` is zero-overhead, and the only way to achieve a `Guard` is calling `write` or `finish`.
```
codeworm96: Need to mention that `Guard` ensures `write` or `finish` has been called at least once in the implementation of the coprocessor. I think the name is confusing. `Guard` does not guard any resources. It is proof that `write` or `finish` has been called. Maybe we can call it `evidence`/`witness`?

TennyZhuang: @breeswish PTAL

text/2020-03-07-copr-chunk.md
```diff
@@ -0,0 +1,214 @@
...
+For fixed size type:
+
+```rust
+// From
+#[rpn_fn]
+pub fn logical_and(lhs: &Option<i64>, rhs: &Option<i64>) -> Result<Option<i64>>;
+
+// To
+#[rpn_fn]
+pub fn logical_and(lhs: &i64, rhs: &i64) -> Result<Option<i64>>;
```
TennyZhuang: Here is a question. For fixed sized type, is it better to use value type instead of ref? almost all fixed type <= 8 byte.

text/2020-03-07-copr-chunk.md
TennyZhuang:
```suggestion
2. `write` can be called at most once on one `Writer`.
3. Only one of `write` and `begin` can be called at most once on one `PartialWriter`.
```

skyzh:
PTAL, thanks! @TennyZhuang @breeswish 
Before merging, @TennyZhuang may need to squash and sign DCO for all commits.
Also, GitHub's markdown doesn't support multi-line table. Markdown lint warns too long lines on table cells, which can't be fixed.

skyzh:
This RFC has been already implemented. cc @andylokandy could you please give an approval and get this merged?

andylokandy (approved):

PR #44, opened 2020-03-11
rfc: versioned kv

fredchenbj:
Signed-off-by: fredchenbj <cfworking@163.com>

### DIFF
--- a/text/2020-03-11-versioned-kv.md
+++ b/text/2020-03-11-versioned-kv.md
@@ -2,105 +2,196 @@
 
 ## Summary
 
-This proposal presents another KV interface - Versioned KV (VerKV), a multi-versioned KV between RawKV and TxnKV.
+This proposal presents another KV interface - Versioned KV (VerKV), a
+multi-versioned KV between RawKV and TxnKV.
 
 ## Motivation
 
-At present, Yidian Zixun adopts TiKV RawKV for Key-Value storage. Due to performance concerns, TxnKV is not used. However, RawKV only provides simple KV semantics, and does not support multiple versions based on timestamp, which causes some problems: 
-  - When services are migrated from HBase to TiKV, RawKV does not fulfil some major functionality requirements such as data expiration and historical data version;
-  - When developing backup and synchronization functions based on RawKV, it is difficult to solve the problems of online data migration, incremental backup, and data synchronization between two data centers because timestamp is not supported.
+At present, Yidian Zixun adopts TiKV RawKV for Key-Value storage. Due to
+performance concerns, TxnKV is not used. However, RawKV only provides simple KV
+semantics, and does not support multiple versions based on timestamp,
+which causes some problems:
 
-In order to solve the above problems, this proposal presents another kv interface - VerKV, which mainly adds version information such as timestamp to the key. At the same time, this interface will be designed to be approximate to the semantics of HBase, which will facilitate the service migration from HBase to TiKV, thereby integrating TiKV better into the HBase ecosystem.
+- When services are migrated from HBase to TiKV, RawKV does not fulfil some
+  major functionality requirements such as data expiration and historical data
+   version;
+- When developing backup and synchronization functions based on RawKV,
+  it is difficult to solve the problems of online data migration, incremental
+   backup, and data synchronization between two data centers because timestamp
+  is not supported.
+
+In order to solve the above problems, this proposal presents another kv
+interface - VerKV, which mainly addsversion information such as timestamp to
+the key. At the same time, this interface will be designed to be approximate
+to the semantics of HBase, which will facilitate the service migration from
+HBase to TiKV, thereby integrating TiKV better into the HBase ecosystem.
 
 ## Detailed designs
 
 ### Functions and scenarios
 
-Here are the functionality designs and the corresponding application scenarios for VerKV:
-   - Support data migration, incremental backup, CDC (Change Data Capture) and cross-IDC cluster synchronization application scenarios include:
-     - Online data replication between TiKV clusters, which includes a full migrations followed by incremental migrations;
-     - Cluster data backup, which includes a first full backup followed perform an incremental data backups every other time;
-     - CDC service created based on incremental data acquisition; can reliably provide incremental data services;
-     - In cross data center cluster synchronization scenarios, VerKV can not only process  incremental data synchronization, but also handle write conflicts caused by the same key.
+Here are the functionality designs and the corresponding application
+scenarios for VerKV:
 
-  - Support multi-versioned data. The corresponding application scenarios include: 
-    - Data version retrieval. For certain businesses, multiple versions of historical data must be maintained for for backtracking and analysis. Reliance on third-party storage and consumption of computing resources (user dynamic characteristics, User behavior link tracking) needs to be reduced, and the data processing process optimized;
-    - Data version control, for fast data recovery after problematic data writes in businesses.
+- Support data migration, incremental backup, CDC (Change Data Capture)
+  and cross-IDC cluster synchronization application scenarios include:
+  
+  - Online data replication between TiKV clusters, which includes a full
+    migrations followed by incremental migrations;
+  - Cluster data backup, which includes a first full backup followed perform
+    an incremental data backups every other time;
+  - CDC service created based on incremental data acquisition; can reliably
+    provide incremental data services;
+  - In cross data center cluster synchronization scenarios, VerKV can not
+    only process incremental data synchronization, but also handle write
+    conflicts caused by the same key.
 
-  - Support data expiration based on TTL. The corresponding application scenarios include:
-    - Setting expiration time to obtain data of a recent period
-    - Setting expiration time to clear inactive data for a specified amount of time
+- Support multi-versioned data. The corresponding application scenarios
+  include:
+  
+  - Data version retrieval. For certain businesses, multiple versions of
+    historical data must be maintained for for backtracking and analysis.
+    Reliance on third-party storage and consumption of computing resources
+    (user dynamic characteristics, User behavior link tracking) needs to be
+    reduced, and the data processing process optimized;
+  - Data version control, for fast data recovery after problematic data
+    writes in businesses.
 
+- Support data expiration based on TTL. The corresponding application
+  scenarios include:
+  
+  - Setting expiration time to obtain data of a recent period
+  - Setting expiration time to clear inactive data for a specified amount
+    of time
 
 ### Compatibility
 
-The design adds a new Column Family called `VerCF` in RocksDB, which avoids poential compatibility issues with the existing RawKV annd TxnKV data and processing logics, and could be created dynamically when needed. This design allows maximum reuse of the current implementations of TiKV and PD.
+The design adds a new Column Family called `VerCF` in RocksDB, which avoids
+potential compatibility issues with the existing RawKV annd TxnKV data and
+processing logics, and could be created dynamically when needed.
+This design allows maximum reuse of the current implementations of TiKV and PD.
 
 ### Data model and interface
 
-VerKV is a kv-like interface similar to the RawKV interface. It hides version information from upper-layer services. Its data model is multi-versioned KV between RawKV and TxnKV. 
+VerKV is a kv-like interface similar to the RawKV interface. It hides version
+information from upper-layer services. Its data model is multi-versioned KV
+between RawKV and TxnKV.
 
 #### Data model
 
-The data model is like `key:version -> value`. The version uses the current timestamp by default, or can be specified by the user (such as the global `ts` obtained from the PD TSO). And the implementation is subject to further change if we use the user TS feature of RocksDB.
+The data model is like `key:version -> value`. The version uses the current
+timestamp by default, or can be specified by the user (such as the global `ts`
+obtained from the PD TSO). And the implementation is subject to further change
+if we use the user TS feature of RocksDB.
 
-The maximum number of versions in the cluster can be configured via `MaxVerNum`. The default value is 1. Redundant versions are removed during version recycling (compaction or garbage collection).
+The maximum number of versions in the cluster can be configured via `MaxVerNum`.
+The default value is 1. Redundant versions are removed during version
+recycling (compaction or garbage collection).
 
 #### Data storage
 
-`user_key` is encoded according to Memory Comparable Encoding in the TiKV client sdk, and the encoding algorithm is padding in 8-byte units. This operation ensures that ts is appended to the key and the written user_key (key: ts) can keep the same order as the key itself;
+`user_key` is encoded according to Memory Comparable Encoding in the TiKV client
+sdk, and the encoding algorithm is padding in 8-byte units. This operation
+ ensures that ts is appended to the key and the written user_key (key: ts)
+can keep the same order as the key itself.
 
-The timestamp stored in user_key is its inversion result. The purpose is to arrange the newer data (that is, the data with larger timestamp) in front of the older data (that is, data with smaller timestamp). This feature can be utilized in performance optimizing for the data scan process.
+The timestamp stored in user_key is its inversion result. The purpose is
+to arrange the newer data (that is, the data with larger timestamp) in front
+ of the older data (that is, data with smaller timestamp).
+This feature can be utilized in performance optimizing for the data
+scan process.
 
 #### Data distribution
 
-To facilitate the acquisition and efficient management of multiple versions of a key, the split point of the region will not use the timestamp of the key, so that multiple versions of the key can be guaranteed to be in the same region and the same CF. Considering the size of the region, MaxVerNum will be restrained.
+To facilitate the acquisition and efficient management of multiple versions of
+a key, the split point of the region will not use the timestamp of the key
+, so that multiple versions of the key can be guaranteed to be
+in the same region and the same CF. Considering the size of the region,
+MaxVerNum will be restrained.
 
-The process of data balancing, scaling, and migration are the same as that of RawKV. In the process of scanning all versions of a key, if the maximum number of versions is exceeded, the redundant versions will be discarded.
+The process of data balancing, scaling, and migration are the same as that
+of RawKV. In the process of scanning all versions of a key, if the maximum
+ number of versions is exceeded, the redundant versions will be discarded.
 
 #### RPC Interface
 
-With reference to the proto interface, the following interfaces are defined as a new service other than tikv:
-  - Get/BatchGet - Gets the values for latest version of the key by default 
-  - GetNum/BatchGetNum - Get the values for a specified number of versions of a key. `version_num` must be less than or equal to `MaxVerNum`. 
-  - GetNumWithSV/BatchGetNumWithSV - This interface is like `GetNum/BatchGetNum`, but add requirement about the key with versions ranging from the specified start version to the current version.
-  - GetWithTTL/BatchGetWithTTL - Gets the unexpired values within the TTL.
-  - Scan - Scans the values of all keys of the latest version within a range
-  - ScanNum - Scans the number of all keys within a range to get the corresponding values of a specified number of versions
-  - ScanNumWithSV - Scans all the keys within a range and gets the corresponding values ranging from the specified start version to the current version 
-  - Put/BatchPut - Increases the new version value of a key, where version is the current timestamp by default
-  - Delete/BatchDelete - Deletes all versions of a specified key (`key:cur_ts -> delete`)
-  - DeleteRange - Deletes all versions of key in a range using rocksdb DeleteRange
+With reference to the proto interface, the following interfaces are defined as
+a new service other than tikv:
 
-The above fetch operations(get/scan) can use `rocksdb prefix_extractor` to improve the performance of iterator seek. `key: ts-> delete` indicates that all data before the `ts` has been deleted.
+- Get/BatchGet - Gets the values for latest version of the key by default
+- GetNum/BatchGetNum - Get the values for a specified number of versions of a
+  key. `version_num` must be less than or equal to `MaxVerNum`.
+- GetNumWithSV/BatchGetNumWithSV - This interface is like `GetNum/BatchGetNum`,
+  but add requirement about the key with versions ranging from the specified
+  start version to the current version.
+- GetWithTTL/BatchGetWithTTL - Gets the unexpired values within the TTL.
+- Scan - Scans the values of all keys of the latest version within a range
+- ScanNum - Scans the number of all keys within a range to get the
+  corresponding values of a specified number of versions
+- ScanNumWithSV - Scans all the keys within a range and gets the
+  corresponding values ranging from the specified start version to the current
+  version
+- Put/BatchPut - Increases the new version value of a key, where version is
+  the current timestamp by default
+- Delete/BatchDelete - Deletes all versions of a specified key
+  (`key:cur_ts -> delete`)
+- DeleteRange - Deletes all versions of key in a range using rocksdb DeleteRange
+
+The above fetch operations(get/scan) can use `rocksdb prefix_extractor` to
+improve the performance of iterator seek. `key: ts-> delete` indicates that
+ all data before the `ts` has been deleted.
 
 ### Version recycling
 
-The design of the version recycling will affect cluster performance and the functional designs of incremental backup and CDC. There are two major approaches: 
-  - using multi-version Garbage Collection (GC) similar to TxnKV, and periodically starting the gc worker in the background to delete redundant versions of data;
-  - using compaction filter via rocksdb compaction to delete redundant versions. 
+The design of the version recycling will affect cluster performance and the
+functional designs of incremental backup and CDC. There are two major
+approaches:
 
-The GC method consumes CPU or storage resources, but the version recycling is accurate and efficient; The compaction filter method does not take up CPU or storage resources. However, compaction only has a local applicable scope, so multi-version recycling efficiency is low.
+- using multi-version Garbage Collection (GC) similar to TxnKV, and
+  periodically starting the gc worker in the background to delete redundant
+  versions of data;
+- using compaction filter via rocksdb compaction to delete redundant versions.
 
-This proposal adopts a combination of the above two methods, with compaction as the major mechanism and GC as the supplement. The default maximum number of versions is 1, which can default to the compaction filter recycling version, so recycling efficiency is moderate. If the maximum number of versions is more than 1 or more efficient version recycling is required, you can dynamically configure GC to enable it.
+The GC method consumes CPU or storage resources, but the version recycling is
+accurate and efficient; The compaction filter method does not take up CPU or
+storage resources. However, compaction only has a local applicable scope, so
+multi-version recycling efficiency is low.
 
-The compaction filter is customized based on the maximum number of versions, `MaxVerNum`, and delete tombstone based on key instead of key: `ts`. So only one `MaxVerNum` can be set for one `cf`.
+This proposal adopts a combination of the above two methods, with compaction as
+the major mechanism and GC as the supplement. The default maximum number of
+versions is 1, which can default to the compaction filter recycling version
+, so recycling efficiency is moderate. If the maximum number of versions is more
+than 1 or more efficient version recycling is required, you can dynamically
+configure GC to enable it.
+
+The compaction filter is customized based on the maximum number of versions,
+`MaxVerNum`, and delete tombstone based on key instead of key: `ts`. So only
+one `MaxVerNum` can be set for one `cf`.
 
 ### Data increment
 
-The `ScanNumWithSV` rpc interface is used to get all data increments of the entire cluster or within a specified range [start_ts, cur_ts). When this interface is called, the snapshot of the entire cluster or the current cur_ts is obtained before scanning.
+The `ScanNumWithSV` rpc interface is used to get all data increments of the
+entire cluster or within a specified range [start_ts, cur_ts). When this
+interface is called, the snapshot of the entire cluster or the current cur_ts
+is obtained before scanning.
 
-VerKV integrated to tidb/br and tidb/ticdc can better facilitate the implementation of full/incremental backup and CDC services.  
+VerKV integrated to tidb/br and tidb/ticdc can better facilitate the
+implementation of full/incremental backup and CDC services.
 
 ### Data replication
 
-- A -> B - asynchronous one-way replication; Cluster B accesses the CDC service of cluster A to pull or push incremental data of each period.
-- A <-> B -  asynchronous two-way replication, which is equivalent to two one-way replication processes. Write conflicts can be resolved according to ts.
+- A -> B: asynchronous one-way replication; Cluster B accesses the CDC service
+of cluster A to pull or push incremental data of each period.
+- A <-> B: asynchronous two-way replication, which is equivalent to two one-way
+replication processes. Write conflicts can be resolved according to ts.
 
 ### Others
 
-Rocksdb User Timestamp（WIP）can also be considered to improve the engine efficiency. 
+Rocksdb User Timestamp（WIP）can also be considered to improve the engine
+efficiency.
 
 ## Drawbacks
 
-This VerKV adds version information, so it would decrease the performance of data fetch interface. And the data version recycling would also add pressure to  system.
+This VerKV adds version information, so it would decrease the performance
+of data fetch interface. And the data version recycling would also add
+pressure to  system.
--- a/text/2020-03-11-versioned-kv.md
+++ b/text/2020-03-11-versioned-kv.md
@@ -77,12 +77,18 @@ VerKV is a kv-like interface similar to the RawKV interface. It hides version
 information from upper-layer services. Its data model is multi-versioned KV
 between RawKV and TxnKV.
 
-#### Data model
+#### Data model and Consistency
 
 The data model is like `key:version -> value`. The version uses the current
 timestamp by default, or can be specified by the user (such as the global `ts`
 obtained from the PD TSO). And the implementation is subject to further change
 if we use the user TS feature of RocksDB.
+
+For consistency, strong and weak consistency would be support. This is
+closely with the TS obtainment method, i.e. if need strong consistency, users
+should use the global PD TSO; if need weak consistency, users may use
+simple local ts method. Under weak consistency situation, the tools like
+`br` would be flexible to handle data consistency.
 
 The maximum number of versions in the cluster can be configured via `MaxVerNum`.
 The default value is 1. Redundant versions are removed during version
@@ -180,10 +186,10 @@ implementation of full/incremental backup and CDC services.
 
 ### Data replication
 
-- A -> B: asynchronous one-way replication; Cluster B accesses the CDC service
-of cluster A to pull or push incremental data of each period.
-- A <-> B: asynchronous two-way replication, which is equivalent to two one-way
-replication processes. Write conflicts can be resolved according to ts.
+- A -> B: asynchronous one-way replication, Cluster B accesses the CDC
+service of cluster A to pull or push incremental data of each period.
+- A <-> B: asynchronous two-way replication, which is equivalent to two
+one-way replication processes. Write conflicts can be resolved according to ts.
 
 ### Others
 
@@ -194,4 +200,4 @@ efficiency.
 
 This VerKV adds version information, so it would decrease the performance
 of data fetch interface. And the data version recycling would also add
-pressure to  system.
+pressure to system.
--- a/text/2020-03-11-versioned-kv.md
+++ b/text/2020-03-11-versioned-kv.md
@@ -149,6 +149,10 @@ improve the performance of iterator seek. `key: ts-> delete` indicates that
 
 ### Version recycling
 
+The version recycling would recycling the redundant versions more than
+`MaxVerNum`, e.g. one key has five versions [v1 v2 v3 v4 v3], and the
+`MaxVerNum` is set to be 3, then [v1 v2] would be recycled.
+
 The design of the version recycling will affect cluster performance and the
 functional designs of incremental backup and CDC. There are two major
 approaches:
--- a/text/2020-03-11-versioned-kv.md
+++ b/text/2020-03-11-versioned-kv.md
@@ -80,7 +80,7 @@ between RawKV and TxnKV.
 #### Data model and Consistency
 
 The data model is like `key:version -> value`. The version uses the global `ts`
-obtained from the PD TSO by default . Then strong consistency (single key level
+obtained from the PD TSO by default. Then strong consistency (single key level
  linearizability) would be support. And the implementation is subject to
  further change if we use the user TS feature of RocksDB.
 
--- a/text/2020-03-11-versioned-kv.md
+++ b/text/2020-03-11-versioned-kv.md
@@ -1,0 +1,106 @@
+# Versioned KV (VerKV)
+
+## Summary
+
+This proposal presents another KV interface - Versioned KV (VerKV), a multi-versioned KV between RawKV and TxnKV.
+
+## Motivation
+
+At present, Yidian Zixun adopts TiKV RawKV for Key-Value storage. Due to performance concerns, TxnKV is not used. However, RawKV only provides simple KV semantics, and does not support multiple versions based on timestamp, which causes some problems: 
+  - When services are migrated from HBase to TiKV, RawKV does not fulfil some major functionality requirements such as data expiration and historical data version;
+  - When developing backup and synchronization functions based on RawKV, it is difficult to solve the problems of online data migration, incremental backup, and data synchronization between two data centers because timestamp is not supported.
+
+In order to solve the above problems, this proposal presents another kv interface - VerKV, which mainly adds version information such as timestamp to the key. At the same time, this interface will be designed to be approximate to the semantics of HBase, which will facilitate the service migration from HBase to TiKV, thereby integrating TiKV better into the HBase ecosystem.
+
+## Detailed designs
+
+### Functions and scenarios
+
+Here are the functionality designs and the corresponding application scenarios for VerKV:
+   - Support data migration, incremental backup, CDC (Change Data Capture) and cross-IDC cluster synchronization application scenarios include:
+     - Online data replication between TiKV clusters, which includes a full migrations followed by incremental migrations;
+     - Cluster data backup, which includes a first full backup followed perform an incremental data backups every other time;
+     - CDC service created based on incremental data acquisition; can reliably provide incremental data services;
+     - In cross data center cluster synchronization scenarios, VerKV can not only process  incremental data synchronization, but also handle write conflicts caused by the same key.
+
+  - Support multi-versioned data. The corresponding application scenarios include: 
+    - Data version retrieval. For certain businesses, multiple versions of historical data must be maintained for for backtracking and analysis. Reliance on third-party storage and consumption of computing resources (user dynamic characteristics, User behavior link tracking) needs to be reduced, and the data processing process optimized;
+    - Data version control, for fast data recovery after problematic data writes in businesses.
+
+  - Support data expiration based on TTL. The corresponding application scenarios include:
+    - Setting expiration time to obtain data of a recent period
+    - Setting expiration time to clear inactive data for a specified amount of time
+
+
+### Compatibility
+
+The design adds a new Column Family called `VerCF` in RocksDB, which avoids poential compatibility issues with the existing RawKV annd TxnKV data and processing logics, and could be created dynamically when needed. This design allows maximum reuse of the current implementations of TiKV and PD.
+
+### Data model and interface
+
+VerKV is a kv-like interface similar to the RawKV interface. It hides version information from upper-layer services. Its data model is multi-versioned KV between RawKV and TxnKV. 
+
+#### Data model
+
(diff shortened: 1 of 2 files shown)


BusyJay:
@overvenus PTAL

text/2020-03-11-versioned-kv.md
```diff
@@ -0,0 +1,396 @@
...
+
+The design adds a VerKV Column Family in RocksDB, which avoids poential compatibility issues with the existing RawKV annd TxnKV data and processing logics, and could be created dynamically when needed. This design allows maximum reuse of the current implementations of TiKV and PD.
+
+### Data model and interface
+
+VerKV is a kv-like interface similar to the RawKV interface. It hides version information from upper-layer services. Its data model is multi-versioned KV between RawKV and TxnKV. 
+
+#### Data model
+
+The data model is `key:version -> value`. The version uses the current timestamp by default, or can be specified by the user (such as the global `ts` obtained from the PD TSO).
```
BusyJay: The data schema is better leave to undecided. So that we may utilize user TS support from rocksdb directly.

text/2020-03-11-versioned-kv.md
```diff
@@ -0,0 +1,396 @@
...
+This VerKV adds version information, so it would decrease the performance of data fetch interface. And the data version recycling would also add pressure to  system.
+
+## Appendix
+
+### Proto3 definition
+
+```
+// used to transport a value and the version as a result for get methods.
+// here T stands for Timestamp.
+struct TValue {
```
zhangjinpeng1987: I prefer VerValue

text/2020-03-11-versioned-kv.md
```diff
@@ -0,0 +1,396 @@
...
+  - Delete/BatchDelete - Deletes all versions of a specified key (`key:cur_ts -> delete`)
+  - DeleteRange - Deletes all versions of key in a range using rocksdb DeleteRange
+
+The above fetch operations(get/scan) can use `rocksdb prefix_extractor` to improve the performance of iterator seek. `key: ts-> delete` indicates that all data before the `ts` has been deleted.
+
+### Version recycling
+
+The design of the version recycling will affect cluster performance and the functional designs of incremental backup and CDC. There are two major approaches: 
+  - using multi-version Garbage Collection (GC) similar to TxnKV, and periodically starting the gc worker in the background to delete redundant versions of data;
+  - using compaction filter via rocksdb compaction to delete redundant versions.
```
zhangjinpeng1987: I prefer this way.

text/2020-03-11-versioned-kv.md
fredchenbj: OK

text/2020-03-11-versioned-kv.md
fredchenbj: For rocksdb user TS now is not stable, the first edition would not use this feature. The data model would be ok now. Here, I would add some comments about other possibilities.

text/2020-03-11-versioned-kv.md
fredchenbj: If compaction filter would have a good gc effect, it is better way.

text/2020-03-11-versioned-kv.md
```diff
@@ -42,7 +42,7 @@ VerKV is a kv-like interface similar to the RawKV interface. It hides version in
 
 #### Data model
 
-The data model is `key:version -> value`. The version uses the current timestamp by default, or can be specified by the user (such as the global `ts` obtained from the PD TSO).
+The data model is like `key:version -> value`. The version uses the current timestamp by default, or can be specified by the user (such as the global `ts` obtained from the PD TSO). And the implementation would be changed when consider to use rocksdb user TS feature.
```
dcalvin:
```suggestion
The data model is like `key:version -> value`. The version uses the current timestamp by default, or can be specified by the user (such as the global `ts` obtained from the PD TSO). And the implementation is subject to further change if we use the user TS feature of RocksDB.
```

text/2020-03-11-versioned-kv.md
```diff
@@ -0,0 +1,395 @@
...
+
+Rocksdb User Timestamp（WIP）can also be considered to improve the engine efficiency. 
+
+## Drawbacks
+
+This VerKV adds version information, so it would decrease the performance of data fetch interface. And the data version recycling would also add pressure to  system.
+
+## Appendix
+
+### Proto3 definition
```
BusyJay: I suggest to remove the definitions. They are subject to change and becomes misleading/useless in the end.

fredchenbj: done

text/2020-03-11-versioned-kv.md
```diff
@@ -0,0 +1,107 @@
...
+
+The design adds a VerKV Column Family in RocksDB, which avoids poential compatibility issues with the existing RawKV annd TxnKV data and processing logics, and could be created dynamically when needed. This design allows maximum reuse of the current implementations of TiKV and PD.
+
+### Data model and interface
+
+VerKV is a kv-like interface similar to the RawKV interface. It hides version information from upper-layer services. Its data model is multi-versioned KV between RawKV and TxnKV. 
+
+#### Data model
+
+The data model is like `key:version -> value`. The version uses the current timestamp by default, or can be specified by the user (such as the global `ts` obtained from the PD TSO). And the implementation is subject to further change if we use the user TS feature of RocksDB.
```
Connor1996: To support TTL in TinyKV, I think we should determine the format of the timestamp, maybe just use the PD TSO.

text/2020-03-11-versioned-kv.md
```diff
@@ -0,0 +1,107 @@
...
+
+### Data increment
+
+The `ScanNumWithSV` rpc interface is used to get all data increments  of the entire cluster or within a specified range [start_ts, cur_ts). When this interface is called, the snapshot of  the entire cluster or the current cur_ts is obtained before scanning.
+
+VerKV integrated to tidb/br and tidb/ticdc can better facilitate the implementation of full/incremental backup and CDC services.  
+
+### Data replication
+
+- A-> B - asynchronous one-way replication; Cluster B accesses the CDC service of cluster A to pull or push incremental data of each period.
```
Connor1996:
```suggestion
- A -> B - asynchronous one-way replication; Cluster B accesses the CDC service of cluster A to pull or push incremental data of each period.
```

text/2020-03-11-versioned-kv.md
```diff
@@ -0,0 +1,107 @@
...
+   - Support data migration, incremental backup, CDC (Change Data Capture) and cross-IDC cluster synchronization application scenarios include:
+     - Online data replication between TiKV clusters, which include a full migrations followed by incremental migrations;
+     - Cluster data backup, which includes a first full backup followed  perform an incremental data backups every other time;
+     - CDC service created based on incremental data acquisition; can reliably provide incremental data services;
+     - In cross data center cluster synchronization scenarios, VerKV can not only process  incremental data synchronization, but also handle write conflicts caused by the same key.
+
+  - Support multi-versioned data. The corresponding application scenarios include: 
+    - Data version retrieval. For certain businesses, multiple versions of historical data must be maintained for for backtracking and analysis. Reliance on third-party storage and consumption of computing resources (user dynamic characteristics , User behavior link tracking) needs to be reduced, and the data processing process optimized;
```
Connor1996:
```suggestion
    - Data version retrieval. For certain businesses, multiple versions of historical data must be maintained for for backtracking and analysis. Reliance on third-party storage and consumption of computing resources (user dynamic characteristics, User behavior link tracking) needs to be reduced, and the data processing process optimized;
```

text/2020-03-11-versioned-kv.md
```diff
@@ -0,0 +1,107 @@
...
+
+The GC method consumes CPU or storage resources, but the version recycling is accurate and efficient; The compaction filter method does not take up CPU or storage resources. However, compaction only has a local applicable scope, so multi-version recycling efficiency is low.
+
+This proposal adopts a combination of the above two methods, with compaction as the major mechanism and GC as the supplement. The default maximum number of versions is 1, which can default to the compaction filter recycling version, so recycling efficiency is moderate. If the maximum number of versions is more than 1 or more efficient version recycling is required, you can dynamically configure GC to enable it.
+
+The compaction filter is customized based on the maximum number of versions, `MaxVerNum`, and delete tombstone based on key instead of key: `ts`. So only one `MaxVerNum` can be set for one `cf`.
+
+### Data increment
+
+The `ScanNumWithSV` rpc interface is used to get all data increments  of the entire cluster or within a specified range [start_ts, cur_ts). When this interface is called, the snapshot of  the entire cluster or the current cur_ts is obtained before scanning.
```
Connor1996:
```suggestion
The `ScanNumWithSV` rpc interface is used to get all data increments of the entire cluster or within a specified range [start_ts, cur_ts). When this interface is called, the snapshot of the entire cluster or the current cur_ts is obtained before scanning.
```

text/2020-03-11-versioned-kv.md
```diff
@@ -0,0 +1,107 @@
...
+#### Data distribution
+
+To facilitate the acquisition and efficient management of multiple versions of a key, the split point of the region will not use the timestamp of the key, so that multiple versions of the key can be guaranteed to be in the same region and the same CF. Considering the size of the region, MaxVerNum will be restrained.
+
+The process of data balancing, scaling, and migration are the same as that of RawKV. In the process of scanning all versions of a key, if the maximum number of versions is exceeded, the redundant versions will be discarded.
+
+#### RPC Interface
+
+With reference to the proto interface, the following interfaces are defined as a new service other than tikv:
+  - Get/BatchGet -  Gets the values for latest version of the key by default
```
Connor1996:
```suggestion
  - Get/BatchGet - Gets the values for latest version of the key by default 
```

text/2020-03-11-versioned-kv.md
```diff
@@ -0,0 +1,107 @@
...
+
+#### RPC Interface
+
+With reference to the proto interface, the following interfaces are defined as a new service other than tikv:
+  - Get/BatchGet -  Gets the values for latest version of the key by default 
+  - GetNum/BatchGetNum - Get the values for a specified number of versions of a key. `version_num` must be less than or equal to `MaxVerNum`. 
+  - GetNumWithSV/BatchGetNumWithSV - This interface is like `GetNum/BatchGetNum`, but add requirement about the key with versions ranging from the specified start version to the current version.
+  - GetWithTTL/BatchGetWithTTL - Gets the unexpired values within the TTL.
+  - Scan - Scans the values of all keys of the latest version within a range
+  - ScanNum -  Scans the number of all keys within a range to get the corresponding values of a specified number of versions
```
Connor1996:
```suggestion
  - ScanNum - Scans the number of all keys within a range to get the corresponding values of a specified number of versions
```

text/2020-03-11-versioned-kv.md
```diff
@@ -0,0 +1,107 @@
...
+  - ScanNumWithSV - Scans all the keys within a range and gets the corresponding values ranging from the specified start version to the current version 
+  - Put/BatchPut -  Increases the new version value of a key, where version is the current timestamp by default
```
Connor1996:
```suggestion
  - Put/BatchPut - Increases the new version value of a key, where version is the current timestamp by default
```

text/2020-03-11-versioned-kv.md
```diff
@@ -0,0 +1,107 @@
...
+
+At present, Yidian Zixun adopts TiKV RawKV for Key-Value storage. Due to performance concerns, TxnKV is not used. However, RawKV only provides simple KV semantics, and does not support multiple versions based on timestamp, which causes some problems: 
+  - When services are migrated from HBase to TiKV, RawKV does not fulfil some major functionality requirements such as data expiration and historical data version;
+  - When developing backup and synchronization functions based on RawKV, it is difficult to solve the problems of online data migration, incremental backup, and data synchronization between two data centers because timestamp is not supported.
+
+In order to solve the above problems, this proposal presents another kv interface - VerKV, which mainly adds version information such as timestamp to the key. At the same time, this interface will be designed to be approximate to the semantics of HBase, which will facilitate the service migration from HBase to TiKV, thereby integrating TiKV better into the HBase ecosystem.
+
+## Detailed designs
+
+### Functions and scenarios
```
Connor1996: Do we need to support multiple cfs or just one cf

text/2020-03-11-versioned-kv.md
```diff
@@ -0,0 +1,107 @@
...
+    - Data version control, for fast data recovery after problematic data writes in businesses.
+
+  - Support data expiration based on TTL. The corresponding application scenarios include:
+    - Setting expiration time to obtain data of a recent period
+    - Setting expiration time to clear inactive data for a specified amount of time
+
+
+### Compatibility
+
+The design adds a VerKV Column Family in RocksDB, which avoids poential compatibility issues with the existing RawKV annd TxnKV data and processing logics, and could be created dynamically when needed. This design allows maximum reuse of the current implementations of TiKV and PD.
```
Connor1996:
```suggestion
The design adds a new Column Family called `VerCF` in RocksDB, which avoids poential compatibility issues with the existing RawKV annd TxnKV data and processing logics, and could be created dynamically when needed. This design allows maximum reuse of the current implementations of TiKV and PD.
```

Connor1996: I'm not sure about the name, better to decide it now?

text/2020-03-11-versioned-kv.md
```diff
@@ -0,0 +1,107 @@
...
+     - Online data replication between TiKV clusters, which include a full migrations followed by incremental migrations;
+     - Cluster data backup, which includes a first full backup followed  perform an incremental data backups every other time;
+     - CDC service created based on incremental data acquisition; can reliably provide incremental data services;
+     - In cross data center cluster synchronization scenarios, VerKV can not only process  incremental data synchronization, but also handle write conflicts caused by the same key.
+
+  - Support multi-versioned data. The corresponding application scenarios include: 
+    - Data version retrieval. For certain businesses, multiple versions of historical data must be maintained for for backtracking and analysis. Reliance on third-party storage and consumption of computing resources (user dynamic characteristics , User behavior link tracking) needs to be reduced, and the data processing process optimized;
+    - Data version control, for fast data recovery after problematic data writes in businesses.
+
+  - Support data expiration based on TTL. The corresponding application scenarios include:
```
nrc: Why does this need to be implemented in TiKV rather than by the client?

text/2020-03-11-versioned-kv.md
Connor1996: RawKV supports get/put with cf, but restrict the scope with [defaut, write, lock]. Maybe VerKV should not support multiple cfs, and only operate on its dedicated cf?  /cc @zhangjinpeng1987 @BusyJay

text/2020-03-11-versioned-kv.md
```diff
@@ -0,0 +1,107 @@
...
+    - Setting expiration time to obtain data of a recent period
+    - Setting expiration time to clear inactive data for a specified amount of time
+
+
+### Compatibility
+
+The design adds a VerKV Column Family in RocksDB, which avoids poential compatibility issues with the existing RawKV annd TxnKV data and processing logics, and could be created dynamically when needed. This design allows maximum reuse of the current implementations of TiKV and PD.
+
```
nrc: We already have some problem with mixing raw and transactional APIs (as described in the motivation). It seems to me that there is potential to make this worse by adding a third API. As part of this RFC I would like to see some mechanism for better enforcing separation of API use.

text/2020-03-11-versioned-kv.md
```diff
@@ -0,0 +1,107 @@
...
+### Data model and interface
+
+VerKV is a kv-like interface similar to the RawKV interface. It hides version information from upper-layer services. Its data model is multi-versioned KV between RawKV and TxnKV.
```
nrc: Are there VerKV transactions? (Sorry for a such a beginner question - I'm not familiar with HBase)

text/2020-03-11-versioned-kv.md
```diff
@@ -0,0 +1,107 @@
...
+
+### Data model and interface
+
+VerKV is a kv-like interface similar to the RawKV interface. It hides version information from upper-layer services. Its data model is multi-versioned KV between RawKV and TxnKV. 
+
+#### Data model
+
+The data model is like `key:version -> value`. The version uses the current timestamp by default, or can be specified by the user (such as the global `ts` obtained from the PD TSO). And the implementation is subject to further change if we use the user TS feature of RocksDB.
+
+The maximum number of versions in the cluster can be configured via `MaxVerNum`. The default value is 1. Redundant versions are removed during version recycling (compaction or garbage collection).
```
nrc: Is there such a limit for the transactional API?

text/2020-03-11-versioned-kv.md
```diff
@@ -0,0 +1,107 @@
...
+
+To facilitate the acquisition and efficient management of multiple versions of a key, the split point of the region will not use the timestamp of the key, so that multiple versions of the key can be guaranteed to be in the same region and the same CF. Considering the size of the region, MaxVerNum will be restrained.
+
+The process of data balancing, scaling, and migration are the same as that of RawKV. In the process of scanning all versions of a key, if the maximum number of versions is exceeded, the redundant versions will be discarded.
+
+#### RPC Interface
+
+With reference to the proto interface, the following interfaces are defined as a new service other than tikv:
+  - Get/BatchGet -  Gets the values for latest version of the key by default 
+  - GetNum/BatchGetNum - Get the values for a specified number of versions of a key. `version_num` must be less than or equal to `MaxVerNum`.
```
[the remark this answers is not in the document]
nrc: "must be less than or equal to `MaxVerNum`. " Why? Seems that we don't need this restriction

text/2020-03-11-versioned-kv.md
```diff
@@ -0,0 +1,107 @@
...
+To facilitate the acquisition and efficient management of multiple versions of a key, the split point of the region will not use the timestamp of the key, so that multiple versions of the key can be guaranteed to be in the same region and the same CF. Considering the size of the region, MaxVerNum will be restrained.
+
+The process of data balancing, scaling, and migration are the same as that of RawKV. In the process of scanning all versions of a key, if the maximum number of versions is exceeded, the redundant versions will be discarded.
+
+#### RPC Interface
+
+With reference to the proto interface, the following interfaces are defined as a new service other than tikv:
+  - Get/BatchGet -  Gets the values for latest version of the key by default 
+  - GetNum/BatchGetNum - Get the values for a specified number of versions of a key. `version_num` must be less than or equal to `MaxVerNum`. 
+  - GetNumWithSV/BatchGetNumWithSV - This interface is like `GetNum/BatchGetNum`, but add requirement about the key with versions ranging from the specified start version to the current version.
```
nrc: What is the current version?

text/2020-03-11-versioned-kv.md
```diff
@@ -0,0 +1,107 @@
...
+  - GetWithTTL/BatchGetWithTTL - Gets the unexpired values within the TTL.
```
nrc: This seems unnecessary given `GetNumWithSV/BatchGetNumWithSV `, is there a difference?

text/2020-03-11-versioned-kv.md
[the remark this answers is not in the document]
fredchenbj: yes, it would be option. Here I refer to the api of HBase.

text/2020-03-11-versioned-kv.md
```diff
@@ -0,0 +1,107 @@
...
+  - Scan - Scans the values of all keys of the latest version within a range
+  - ScanNum -  Scans the number of all keys within a range to get the corresponding values of a specified number of versions
+  - ScanNumWithSV - Scans all the keys within a range and gets the corresponding values ranging from the specified start version to the current version 
+  - Put/BatchPut -  Increases the new version value of a key, where version is the current timestamp by default
+  - Delete/BatchDelete - Deletes all versions of a specified key (`key:cur_ts -> delete`)
```
nrc: To clarify, this just adds a tombstone version, it does not remove any data from the DB? So using Get with an earlier timestamp will still return some data?

text/2020-03-11-versioned-kv.md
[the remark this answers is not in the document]
fredchenbj: the time of the request

text/2020-03-11-versioned-kv.md
```diff
@@ -0,0 +1,107 @@
...
+  - DeleteRange - Deletes all versions of key in a range using rocksdb DeleteRange
+
+The above fetch operations(get/scan) can use `rocksdb prefix_extractor` to improve the performance of iterator seek. `key: ts-> delete` indicates that all data before the `ts` has been deleted.
+
+### Version recycling
+
```
nrc: How does TiKV know to run version recycling for a CF? Is it specified in the config?

text/2020-03-11-versioned-kv.md
nrc: Are these proposed API functions or new tools or something else?

text/2020-03-11-versioned-kv.md
```diff
@@ -0,0 +1,107 @@
...
+- A <-> B -  asynchronous two-way replication, which is equivalent to two one-way replication processes. Write conflicts can be resolved according to ts.
+
+### Others
+
+Rocksdb User Timestamp（WIP）can also be considered to improve the engine efficiency. 
+
+## Drawbacks
+
+This VerKV adds version information, so it would decrease the performance of data fetch interface. And the data version recycling would also add pressure to  system.
+
```
nrc: It seems to me the primary drawback is the maintenance burden this is likely to add. It adds a lot of API surface, a new encoding scheme, and a new kind of GC. It seems that the latter two could interact with other new or existing features, so the added complexity will potentially affect any work, not just work on VerKV.

It does sound nice to have better support for HBase though! I would like to see some analysis of the cost/benefit trade-off for this RFC.

text/2020-03-11-versioned-kv.md
Connor1996: I have the same question

text/2020-03-11-versioned-kv.md
nrc: I thought raw kv lets you write to any CF?

text/2020-03-11-versioned-kv.md
nrc: Furthermore it seems to shift some responsibility around timestamps to TiKV, which will also increase complexity.

text/2020-03-11-versioned-kv.md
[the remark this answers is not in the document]
fredchenbj: > "must be less than or equal to `MaxVerNum`. " Why? Seems that we don't need this restriction

General, it is needed to restrict the num of value versions, it would help save space, and be good for performance.

text/2020-03-11-versioned-kv.md
fredchenbj: To be consist with RawKV，here just write one cf "verkv".

text/2020-03-11-versioned-kv.md
fredchenbj: Yes, it could be implement in the client sdk, but implemented in tikv would be better for less data communicate between tikv and client.

text/2020-03-11-versioned-kv.md
fredchenbj: For now,  have not considered transactions, later maybe have it.

text/2020-03-11-versioned-kv.md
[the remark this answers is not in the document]
nrc: But if version_num is > MaxVerNum, then TiKV should still only return MaxVerNum number of entries because there shouldn't be more than that in the DB (if there are, then it seems the client may want to know those extra entries too)

text/2020-03-11-versioned-kv.md
fredchenbj: `GetWithTTL` is similar with `GetNumWithSV`，but `GetWithTTL` is related with ttl, while `GetNumWithSV` is related with start version.

text/2020-03-11-versioned-kv.md
fredchenbj: yes, `Delelte` use rocksdb's delete interface, so it is not remove any data internally. Use Get  would not return those data.

text/2020-03-11-versioned-kv.md
fredchenbj: now, the cf is the fixed one "vercf".

text/2020-03-11-versioned-kv.md
fredchenbj: new function would be integrated with `br` and `cdc`.

text/2020-03-11-versioned-kv.md
fredchenbj: For now, VerKV would be used singly, and not affect the work of rawkv and txnkv. The maintenance burden may be high or low. I would have not better cost/benefit trade-off now. And maybe leave it later when implement it.

text/2020-03-11-versioned-kv.md
```diff
@@ -0,0 +1,106 @@
...
+    - Setting expiration time to clear inactive data for a specified amount of time
+
+
+### Compatibility
+
+The design adds a new Column Family called `VerCF` in RocksDB, which avoids poential compatibility issues with the existing RawKV annd TxnKV data and processing logics, and could be created dynamically when needed. This design allows maximum reuse of the current implementations of TiKV and PD.
+
+### Data model and interface
+
+VerKV is a kv-like interface similar to the RawKV interface. It hides version information from upper-layer services. Its data model is multi-versioned KV between RawKV and TxnKV.
```
overvenus: Could you specify the consistency of verkv? It's very important to backup and CDC service.

text/2020-03-11-versioned-kv.md
```diff
@@ -0,0 +1,106 @@
...
+
+The design adds a new Column Family called `VerCF` in RocksDB, which avoids poential compatibility issues with the existing RawKV annd TxnKV data and processing logics, and could be created dynamically when needed. This design allows maximum reuse of the current implementations of TiKV and PD.
+
+### Data model and interface
+
+VerKV is a kv-like interface similar to the RawKV interface. It hides version information from upper-layer services. Its data model is multi-versioned KV between RawKV and TxnKV. 
+
+#### Data model
+
+The data model is like `key:version -> value`. The version uses the current timestamp by default, or can be specified by the user (such as the global `ts` obtained from the PD TSO). And the implementation is subject to further change if we use the user TS feature of RocksDB.
```
overvenus: I suggest keeping only one time-source, PD. Different time-sources may require different implementation of the CDC, it could be difficult.

text/2020-03-11-versioned-kv.md
overvenus: No, but txn API has a time-based limitation (GC lifetime, default 10min).

text/2020-03-11-versioned-kv.md
```diff
@@ -0,0 +1,106 @@
...
+#### Data distribution
+
+To facilitate the acquisition and efficient management of multiple versions of a key, the split point of the region will not use the timestamp of the key, so that multiple versions of the key can be guaranteed to be in the same region and the same CF. Considering the size of the region, MaxVerNum will be restrained.
+
+The process of data balancing, scaling, and migration are the same as that of RawKV. In the process of scanning all versions of a key, if the maximum number of versions is exceeded, the redundant versions will be discarded.
+
+#### RPC Interface
+
+With reference to the proto interface, the following interfaces are defined as a new service other than tikv:
+  - Get/BatchGet - Gets the values for latest version of the key by default
```
overvenus: Can we just keep Get only, and support batch by default?

text/2020-03-11-versioned-kv.md
```diff
@@ -0,0 +1,155 @@
...
+  - Delete/BatchDelete - Deletes all versions of a specified key (`key:cur_ts -> delete`)
+  - DeleteRange - Deletes all versions of key in a range using rocksdb DeleteRange
+
+The above fetch operations(get/scan) can use `rocksdb prefix_extractor` to improve the performance of iterator seek.
+`key: ts-> delete` indicates that all data before the `ts` has been deleted.
+
+### Version recycling
+
+The design of the version recycling will affect cluster performance and the functional designs of
+incremental backup and CDC. There are two major approaches:
```
overvenus: Agree version recycling affect backup and CDC, could you specify under what condition data need to be recycled?

overvenus:
Left a few questions, the overall LGTM, well done!

text/2020-03-11-versioned-kv.md
fredchenbj: Good point, I have add something about data consistency.

text/2020-03-11-versioned-kv.md
fredchenbj: for strong consistency, it is good to use pd tso. For weak consistency, is it possible for cdc to handle it flexibly?

text/2020-03-11-versioned-kv.md
fredchenbj: add `batchget` is just to be consistent with `rawkv` ~

text/2020-03-11-versioned-kv.md
fredchenbj: I have add some content about version recycling condtion.

text/2020-03-11-versioned-kv.md
```diff
@@ -77,13 +77,19 @@ VerKV is a kv-like interface similar to the RawKV interface. It hides version
...
 
-#### Data model
+#### Data model and Consistency
 
 The data model is like `key:version -> value`. The version uses the current
 timestamp by default, or can be specified by the user (such as the global `ts`
 obtained from the PD TSO). And the implementation is subject to further change
 if we use the user TS feature of RocksDB.
 
+For consistency, strong and weak consistency would be support. This is
```
overvenus: Much better! could you specify what strong consistency you refer to? Single key level linearizability or snapshot isolation?

text/2020-03-11-versioned-kv.md
```diff
@@ -149,6 +149,10 @@ improve the performance of iterator seek. `key: ts-> delete` indicates that
 
 ### Version recycling
 
+The version recycling would recycling the redundant versions more than
+`MaxVerNum`, e.g. one key has five versions [v1 v2 v3 v4 v3], and the
```
overvenus:
```suggestion
`MaxVerNum`, e.g. one key has five versions [v1 v2 v3 v4 v5], and the
```

v5?

text/2020-03-11-versioned-kv.md
```diff
@@ -149,6 +149,10 @@ improve the performance of iterator seek. `key: ts-> delete` indicates that
...
+`MaxVerNum` is set to be 3, then [v1 v2] would be recycled.
```
overvenus: Feels like incremental backup will always lose some data if a key is updated very frequently no matter which consistency we choose. Is it okay for your case?

text/2020-03-11-versioned-kv.md
fredchenbj: single key level linearlizability

text/2020-03-11-versioned-kv.md
fredchenbj: it is okay for my case

text/2020-03-11-versioned-kv.md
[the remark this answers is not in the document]
fredchenbj: Yes, client can only know `MaxVerNum` number of entries at most. In most situations, this semantic is okay.

text/2020-03-11-versioned-kv.md
fredchenbj: the problem with mixing raw and transactional APIs is for those two apis use the same `default` cf. so here verkv api use a separated cf. later, it is possible for rawkv to use other cf, then it is ok to use those three apis simultaneously.

text/2020-03-11-versioned-kv.md
overvenus: > For weak consistency, is it possible for cdc to handle it flexibly?

It depends on whether we accept data loss/duplication on CDC service. With weak time-source, it could be difficult or impossible to prevent data loss/duplication.

The behavior of backup and CDC could be undefined if we allow using different time sources.

text/2020-03-11-versioned-kv.md
overvenus: It feels redundant though. I prefer just keep Get, it makes the RPC interface clearer. What do you think @siddontang @zhangjinpeng1987 @BusyJay ?

text/2020-03-11-versioned-kv.md
overvenus: Thank you for the clarification!

text/2020-03-11-versioned-kv.md
nrc: I mean that given that the client should supply the current time, then a TTL limit can be easily computed from the start time and vice versa. (I'm assuming this is using a supplied TTL, not the TTL of the data, which I would expect to be taken into account in any case)

text/2020-03-11-versioned-kv.md
nrc: aiui, that is not the only problem - using the raw API it is possible to edit any CF, which can mean circumventing or just breaking the transactional API. So it is only ok to use both APIs if you don't access any of the three CFs used for transactions with the raw API. There is also back-compat hazard (e.g., if a user has a CF called VerKV already).

As a mechanism for separating APIs, I think separating CFs is fine, I guess we would need some kind of opt-in for users who deliberately want to access VerKV/default/write/lock CFs

text/2020-03-11-versioned-kv.md
```diff
@@ -0,0 +1,210 @@
...
+### Data model and interface
+
+VerKV is a kv-like interface similar to the RawKV interface. It hides version
+information from upper-layer services. Its data model is multi-versioned KV
+between RawKV and TxnKV.
+
+#### Data model and Consistency
+
+The data model is like `key:version -> value`. The version uses the current
+timestamp by default, or can be specified by the user (such as the global `ts`
```
nrc: In either case, I would prefer the client to supply the timestamp. If TiKV knows the current time, that is adding a chunk of new functionality/complexity. In fact, it would be really nice if timestamps could remain opaque to TiKV as far as possible (which I think means we can't support the TTL expiry functionality in TiKV).

nrc:
What does it mean to accept/merge this RFC? If we are to merge the RFC and implementation, would it be easy to remove the implementation in the future if it turns out to have too high a maintenance burden?

One issue I have with this design is that it is a very specific mechanism and does not scale well. If we want to support N data models and APIs, we have to add O(N) amount of code to TiKV and we introduce O(N^2) possible interactions between interfaces. I would much prefer a design that let us support *any* data model and API, e.g., by adding a new client.

I'm also a bit wary of adding such a big feature without analysing the benefits and costs more closely. Is this likely to bring value to existing users? Bring new users? How many? What potential future feature would this RFC make more difficult to implement? How much code is likely to change when implementing, etc.?

dcalvin:
> What does it mean to accept/merge this RFC? If we are to merge the RFC and implementation, would it be easy to remove the implementation in the future if it turns out to have too high a maintenance burden?
> 
> One issue I have with this design is that it is a very specific mechanism and does not scale well. If we want to support N data models and APIs, we have to add O(N) amount of code to TiKV and we introduce O(N^2) possible interactions between interfaces. I would much prefer a design that let us support _any_ data model and API, e.g., by adding a new client.
> 
> I'm also a bit wary of adding such a big feature without analysing the benefits and costs more closely. Is this likely to bring value to existing users? Bring new users? How many? What potential future feature would this RFC make more difficult to implement? How much code is likely to change when implementing, etc.?

@sunxiaoguang PTAL. Could you explain a little about how VerKV would be a useful  application case for Zhihu?

text/2020-03-11-versioned-kv.md
sunxiaoguang: I would prefer client giving timestamp as well.

sunxiaoguang:
> > What does it mean to accept/merge this RFC? If we are to merge the RFC and implementation, would it be easy to remove the implementation in the future if it turns out to have too high a maintenance burden?
> > One issue I have with this design is that it is a very specific mechanism and does not scale well. If we want to support N data models and APIs, we have to add O(N) amount of code to TiKV and we introduce O(N^2) possible interactions between interfaces. I would much prefer a design that let us support _any_ data model and API, e.g., by adding a new client.
> > I'm also a bit wary of adding such a big feature without analysing the benefits and costs more closely. Is this likely to bring value to existing users? Bring new users? How many? What potential future feature would this RFC make more difficult to implement? How much code is likely to change when implementing, etc.?
> 
> @sunxiaoguang PTAL. Could you explain a little about how VerKV would be a useful application case for Zhihu?

I would say it could be think of as a brand new enhanced version of RawKV that fixes some legacy issues that couldn't be fixed easily. Since the abstraction can fully cover what RawKV does right now, with acceptable extra costs. We can probably  deprecate RawKV in the future and only support Versioned KV instead.

fredchenbj:
In our business situation, the RawKV had met some problems like increment backup, change data capture, ttl etc. Those problems couldn't be solved well without the meaning of timestamp or version. So we would like to use VerKV to replace RawKV in the future.  Now this rfc is not complete on some aspect for my limited capability, so everyone is welcome to give help or advices.

fredchenbj:
PTAL @overvenus

dcalvin:
This RFC has been also a proposal for GSoC, and someone seems interested in this project. https://github.com/tikv/tikv/issues/6508

text/2020-03-11-versioned-kv.md
```diff
@@ -0,0 +1,196 @@
...
+### Data model and interface
+
+VerKV is a kv-like interface similar to the RawKV interface. It hides version
+information from upper-layer services. Its data model is multi-versioned KV
+between RawKV and TxnKV.
+
+#### Data model and Consistency
+
+The data model is like `key:version -> value`. The version uses the global `ts`
+obtained from the PD TSO by default . Then strong consistency (single key level
```
Connor1996:
```suggestion
obtained from the PD TSO by default. Then strong consistency (single key level
```

fredchenbj: have fixed it. @Connor1996

dcalvin:
> > > What does it mean to accept/merge this RFC? If we are to merge the RFC and implementation, would it be easy to remove the implementation in the future if it turns out to have too high a maintenance burden?
> > > One issue I have with this design is that it is a very specific mechanism and does not scale well. If we want to support N data models and APIs, we have to add O(N) amount of code to TiKV and we introduce O(N^2) possible interactions between interfaces. I would much prefer a design that let us support _any_ data model and API, e.g., by adding a new client.
> > > I'm also a bit wary of adding such a big feature without analysing the benefits and costs more closely. Is this likely to bring value to existing users? Bring new users? How many? What potential future feature would this RFC make more difficult to implement? How much code is likely to change when implementing, etc.?
> > 
> > 
> > @sunxiaoguang PTAL. Could you explain a little about how VerKV would be a useful application case for Zhihu?
> 
> I would say it could be thought of as a brand new enhanced version of RawKV that fixes some legacy issues which couldn't be otherwise fixed easily. Since the abstraction can fully cover what RawKV does right now, with acceptable extra costs. We can probably deprecate RawKV in the future and only support Versioned KV instead.

@nrc PTAL again

text/2020-03-11-versioned-kv.md
```diff
@@ -0,0 +1,203 @@
...
+As compared with RawKV, this VerKV adds version information, so it would
+ decrease the performance of data fetch interface and the version recycling
+would add pressure to system.
+
+And for VerKV is going to replace RawKV, it could not be updated directly
+ from RawKV to VerKV, and have to migrated date to a new cluster with VerKV
+using some method.
+
+Meanwhile, add a new API would add maintenance burden during a period of time
+ before deprecating RawKV in the future.
```
dcalvin:
```suggestion
Compared with RawKV, this VerKV adds version information, so it would
 decrease the performance of data fetch interface and the version recycling
would add pressure to the system.

Migration from RawKV to VerKV cannot be one directly.  The data has to be  
migrated to a new cluster with VerKV using some extra methods.

Meanwhile, introducing a new API would add maintenance burden for a period of time
before RawKV is deprecated in the future.
```

text/2020-03-11-versioned-kv.md
```diff
@@ -0,0 +1,196 @@
...
+- Support data expiration based on TTL. The corresponding application
+  scenarios include:
+
+  - Setting expiration time to obtain data of a recent period
+  - Setting expiration time to clear inactive data for a specified amount
+    of time
+
+### Compatibility
+
+The design adds a new Column Family called `VerCF` in RocksDB, which avoids
```
yiwu-arbug: maybe we can support multiple CF, by prefix all such CF with `VerCF_` prefix? Similarly we can prefix RawKV CFs with `RawCF_` prefix, then we can support mixed use of all these incompatible APIs.

text/2020-03-11-versioned-kv.md
```diff
@@ -0,0 +1,196 @@
...
+
+  - Data version retrieval. For certain businesses, multiple versions of
+    historical data must be maintained for for backtracking and analysis.
+    Reliance on third-party storage and consumption of computing resources
+    (user dynamic characteristics, User behavior link tracking) needs to be
+    reduced, and the data processing process optimized;
+  - Data version control, for fast data recovery after problematic data
+    writes in businesses.
+
+- Support data expiration based on TTL. The corresponding application
```
yiwu-arbug: Can you be more specific on how to support TTL?

text/2020-03-11-versioned-kv.md
```diff
@@ -0,0 +1,196 @@
...
+VerKV is a kv-like interface similar to the RawKV interface. It hides version
+information from upper-layer services. Its data model is multi-versioned KV
+between RawKV and TxnKV.
+
+#### Data model and Consistency
+
+The data model is like `key:version -> value`. The version uses the global `ts`
+obtained from the PD TSO by default. Then strong consistency (single key level
+ linearizability) would be support. And the implementation is subject to
+ further change if we use the user TS feature of RocksDB.
```
yiwu-arbug: Don't worry about user TS of rocksdb - the key format is compatible.

text/2020-03-11-versioned-kv.md
```diff
@@ -0,0 +1,196 @@
...
+The design of the version recycling will affect cluster performance and the
+functional designs of incremental backup and CDC. There are two major
+approaches:
+
+- using multi-version Garbage Collection (GC) similar to TxnKV, and
+  periodically starting the gc worker in the background to delete redundant
+  versions of data;
+- using compaction filter via rocksdb compaction to delete redundant versions.
+  The default maximum number of versions is 1, which can default to the
+  compaction filter recycling version , so recycling efficiency is moderate.
```
yiwu-arbug: As we are adding compaction filter GC for TxnKV, these two items boil down to compatibility with TxnKV GC.

text/2020-03-11-versioned-kv.md
fredchenbj: With `getWithTTL` interface, it would only return data which's timestamp is bigger than `timestampNow - TTL`

text/2020-03-11-versioned-kv.md
fredchenbj: multple CF is not supported temporarily, but will reserve a field of cf for implementing multiple CF later.

text/2020-03-11-versioned-kv.md
```diff
@@ -0,0 +1,202 @@
...
+- ScanNum - Scans the number of all keys within a range to get the
+  corresponding values of a specified number of versions
+- ScanNumWithSV - Scans all the keys within a range and gets the
+  corresponding values ranging from the specified start version to the current
+  version
+- Put/BatchPut - Increases the new version value of a key, where version is
+  the current timestamp by default
+- Delete/BatchDelete - Deletes all versions of a specified key
+  (`key:cur_ts -> delete`)
+- DeleteRange - Deletes all versions of key in a range using rocksdb DeleteRange
```
nrc: Could we start with a more minimal API and then add some of these later once we have some experience with the API? One way to do that is to prioritize the commands, and we'd implement the p1 commands, then wait a bit, then the p2 commands, etc.

Some orthogonal questions:
* could the four `get` commands be combined by having a max num, min version, and TTL arguments? Likewise with Scan.
* I still think that the GetWithTTL command is not needed: e.g., if the current ts is 1000 and the client wants GetWithTTL with a TTL of 100, then the client could call GetNumWithSV with ts = 900 (and num is some large number).
* The description of `Put` is confusing to me. Do you mean "adds a new value for a key with a given timestamp"? (As mentioned elsewhere, I don't think TiKV should know the current time).

text/2020-03-11-versioned-kv.md
```diff
@@ -0,0 +1,202 @@
...
+
+The above fetch operations(get/scan) can use `rocksdb prefix_extractor` to
+improve the performance of iterator seek. `key: ts-> delete` indicates that
+ all data before the `ts` has been deleted.
```
nrc: How is `delete` represented in the store?

fredchenbj: the `delete` is represented same with rocksdb's implementation.

text/2020-03-11-versioned-kv.md
fredchenbj: Ok, I would modify about `get` and remove `GetWithTTL`. for `Put`, the timestamp is obtained from the PD TSO when the client sdk is doing the request .

dcalvin:
LGTM

PR #45, opened 2020-03-12
add rfc for Load-based Region Splitting

lidaobing:


text/2020-03-12-load-based-region-splitting.md
```diff
@@ -0,0 +1,80 @@
...
+  prevented from being merged immediately. Therefore, the overall loads of the
+  two regions should be checked before the region merge.
+* To avoid back-and-forth splitting and merging, the merging load threshold
+  should be slightly lower than splitting load threshold (e.g., 20% lower).
+
+### Detailed Design
+
+1. When the read load of a region exceeds the splitting threshold, start
+   counting the traffic of keys. Through probability sampling of requests on
+   this region, 20 keys are randomly selected first.
```
zhangjinpeng1987: `20 keys are randomly selected` i don't know if there is a efficient way to do this.

lidaobing: @zhangjinpeng1987 

if there are N keys accessed in the first second, wen can find the 20 keys with following method, for the most case, it's complexity is O(1), for the worst case, it's O(N)

```python
keys = [...]
result = set([])
for i in range(20):
    idx = random(0, N)
    idx2 = idx
    while True:
        if keys[idx2] not in result:
            result.add(keys[idx2])
            break
        idx2 = (idx2 + 1) % N
        if idx2 == idx:
            return result # not enough keys
return result       
```

if the worst case is not acceptable, we can limit the scan size to 100(or even 1), but  we'll have a larger chance for not having enough keys. And I think less than 20 keys is not a problem.

```python
keys = [...]
result = set([])
for i in range(20):
    idx = random(0, N)
    idx2 = idx
    for j in range(100):
        idx2 = (idx + j) % N
        if keys[idx2] not in result:
            result.add(keys[idx2])
            break
return result       
```

zhangjinpeng1987: We sampled the keys by 4MB distance and store them in the SST's user properties, we can say these keys split the region into many small pieces, I think we can make use of these keys to speed up the selection process.

PR #49, opened 2020-04-23
Add rfc: rawkv-txnkv-mode-switching

innerr:
Signed-off-by: Liu Cong <liucong@pingcap.com>

### DIFF
--- a/text/2020-04-23-rawkv-txnkv-mode-switching.md
+++ b/text/2020-04-23-rawkv-txnkv-mode-switching.md
@@ -2,39 +2,59 @@
 
 ## Summary
 
-This RFC proposes to forbid users using RawKV API on the TiKV cluster by default and provide a new API command to let users enable it.
+This RFC proposes to forbid users using RawKV API on the TiKV cluster by default,
+and provide a new API command to let users enable it.
 
 ## Motivation
 
-Some painful incidents happened before and still happening, because of the mixed-using of RawKV API and TransactionKV API, some key-values were overwritten by RawKV API, which originally written by TransactionKV API, then led to data corruption.
+Some painful incidents happened before and still happening,
+because of the mixed-using of RawKV API and TransactionKV API,
+some key-values were overwritten by RawKV API,
+which originally written by TransactionKV API,
+then led to data corruption.
 
-We should add more warnings about this in our documents, but it's not enough, documents could easily be ignored.
+We should add more warnings about this in our documents, but it's not enough,
+documents could easily be ignored.
 
-Providing a switching API could prevent people from doing this wrong thing without noticing, when users need to step into the danger zone, they need to explicitly open the guarded switch. And they could shut down the switch anytime to gain more safety.
+Providing a switching API could prevent people from misuse without noticing,
+when users need to step into the danger zone,
+they need to explicitly open the guarded switch.
+And they could shut down the switch anytime to gain more safety.
 
 ## Detailed design
 
-This `mode` flag should be persisted, each member of the TiKV cluster should see the same value in the same time, so it should be write/read by RawKV commands. Or we just use the dynamic configuration feature to achieve our goal.
+This `mode` flag should be persisted,
+each member of the TiKV cluster should see the same value in the same time,
+so it should be write/read by RawKV commands.
+Or we just use the dynamic configuration feature to achieve our goal.
 
 `mode` should be an enum:
-- `ModeRawKV`, indicate that the TxnKV commands could not be called. When they are called they will return errors.
-- `ModeTxnKV`, indicate that the RawKV commands could not be called.
-- `ModeTxnKVRawKV`, works as now.
+
+- `ModeRawKV` indicate that the TxnKV commands could not be called.
+  When they are called they will return errors.
+- `ModeTxnKV` indicate that the RawKV commands could not be called.
+- `ModeTxnKVRawKV` works as now.
+
 The default value of `mode` should be `TxnKV`
 
 Provide a config entry `mode` under `[server]` in the config file.
-It's necessary when we trying to rolling-update a cluster which already running mixed API.
+It's necessary when we trying to rolling-update a cluster which already
+running mixed API.
 
 Add a pair of command `Get/SetMode(mode)` to the API.
 
 ## Drawbacks
 
-The VerKV commands should not be affected by the `mode` flag, because they use different keyspace and could be mixed-using with either RawKV or TxnKV commands.
+The VerKV commands should not be affected by the `mode` flag,
+because they use different keyspace and could be mixed-using with
+either RawKV or TxnKV commands.
 So, the semantic meaning of the `mode` flag will be a little confusing.
 
 ## Alternatives
 
-Modified the RawKV implement, let them manipulate keys in different keyspace as VerKV did.
+Modified the RawKV implement, let them manipulate keys in
+different keyspace as VerKV did.
+
 - This will impact data compatibility, especially on cluster upgrading.
 - Use RawKV to manipulate TxnKV data sometimes could be useful, eg, data fixing.
 
--- a/text/2020-04-23-rawkv-txnkv-mode-switching.md
+++ b/text/2020-04-23-rawkv-txnkv-mode-switching.md
@@ -3,52 +3,52 @@
 ## Summary
 
 This RFC proposes to forbid users using RawKV API on the TiKV cluster by default,
-and provide a new API command to let users enable it.
+and provide a way for users to enable it.
 
 ## Motivation
 
 Some painful incidents happened before and still happening,
-because of the mixed-using of RawKV API and TransactionKV API,
-some key-values were overwritten by RawKV API,
-which originally written by TransactionKV API,
-then led to data corruption.
+because of the mixed-using of RawKV interface and TransactionKV interface,
+some key-values were overwritten by RawKV,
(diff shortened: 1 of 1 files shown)


text/2020-04-23-rawkv-txnkv-mode-switching.md
```diff
@@ -0,0 +1,43 @@
...
+## Drawbacks
+
+The VerKV commands should not be affected by the `mode` flag, because they use different keyspace and could be mixed-using with either RawKV or TxnKV commands.
+So, the semantic meaning of the `mode` flag will be a little confusing.
+
+## Alternatives
+
+Modified the RawKV implement, let them manipulate keys in different keyspace as VerKV did.
+- This will impact data compatibility, especially on cluster upgrading.
+- Use RawKV to manipulate TxnKV data sometimes could be useful, eg, data fixing.
```
nrc, line 39: From the VerKV RFC, it sounds like using the transactional API to manipulate metadata and raw kv for data is a use case too

text/2020-04-23-rawkv-txnkv-mode-switching.md
```diff
@@ -0,0 +1,43 @@
...
+`mode` should be an enum:
+- `ModeRawKV`, indicate that the TxnKV commands could not be called. When they are called they will return errors.
+- `ModeTxnKV`, indicate that the RawKV commands could not be called.
+- `ModeTxnKVRawKV`, works as now.
+The default value of `mode` should be `TxnKV`
+
+Provide a config entry `mode` under `[server]` in the config file.
+It's necessary when we trying to rolling-update a cluster which already running mixed API.
+
+Add a pair of command `Get/SetMode(mode)` to the API.
```
nrc, line 28: Is there a use case for dynamically changing the mode?

text/2020-04-23-rawkv-txnkv-mode-switching.md
innerr: I don't have a confirmed case for that base on the information on hand, maybe @zhangjinpeng1987 could confirm this.

IMO, this use case should not be encouraged, the interface of RawKV is clean, data-safe, the TxnKV is also clean, but the mixed-using interface is not, it doesn't provide data-safe semitics. So I hope that users can use RawKV or TxnKV along, or use them at different times in the same cluster in some rare situations, but never mixed-using at the same time.

Keeping the mixed-mode just for: A maintains; B compatible for exists use case if there is.

text/2020-04-23-rawkv-txnkv-mode-switching.md
innerr: I think dynamically changing configs is a "better be" option than static configuration, it avoids cluster rebooting.

text/2020-04-23-rawkv-txnkv-mode-switching.md
```diff
@@ -0,0 +1,63 @@
...
+each member of the TiKV cluster should see the same value in the same time,
+so it should be write/read by RawKV commands.
+Or we just use the dynamic configuration feature to achieve our goal.
+
+`mode` should be an enum:
+
+- `ModeRawKV` indicate that the TxnKV commands could not be called.
+  When they are called they will return errors.
+- `ModeTxnKV` indicate that the RawKV commands could not be called.
+- `ModeTxnKVRawKV` works as now.
```
zhangjinpeng1987, line 36: This will mislead users to mix using rawkv and txnkv, I suggest to remove it.

nrc: Don't we need this for backwards compatabilty?

text/2020-04-23-rawkv-txnkv-mode-switching.md
```diff
@@ -0,0 +1,63 @@
...
+documents could easily be ignored.
+
+Providing a switching API could prevent people from misuse without noticing,
+when users need to step into the danger zone,
+they need to explicitly open the guarded switch.
+And they could shut down the switch anytime to gain more safety.
+
+## Detailed design
+
+This `mode` flag should be persisted,
```
yiwu-arbug, line 26: where does this get persisted? is it a special key in the underlying engine, or in a separate file?

text/2020-04-23-rawkv-txnkv-mode-switching.md
yiwu-arbug: shall we make use of the dynamic config feature instead of adding a new command?

text/2020-04-23-rawkv-txnkv-mode-switching.md
```diff
@@ -0,0 +1,63 @@
...
+running mixed API.
+
+Add a pair of command `Get/SetMode(mode)` to the API.
+
+## Drawbacks
+
+The VerKV commands should not be affected by the `mode` flag,
+because they use different keyspace and could be mixed-using with
+either RawKV or TxnKV commands.
+So, the semantic meaning of the `mode` flag will be a little confusing.
```
yiwu-arbug, line 51: maybe rename the enums to be "RawKVOnly", "RawKVMixed" and "DisallowRawKV"? Also the name `mode` is too general, maybe rename to `rawkv-mode`?

text/2020-04-23-rawkv-txnkv-mode-switching.md
innerr: @yiwu-arbug good idea

text/2020-04-23-rawkv-txnkv-mode-switching.md
innerr: @nrc yes it just means for backward compatibility, if @zhangjinpeng1987 can confirm it's not necessary, then we should remove it

text/2020-04-23-rawkv-txnkv-mode-switching.md
innerr: If we use dynamic config as you mention above, then it will be persisted along with other dynamic config entries (in PD)

text/2020-04-23-rawkv-txnkv-mode-switching.md
innerr: Good idea, if `Mixed` is not needed, then it could be just `enable-rawkv`, and the flag value can be `true|false` then

innerr:
Addressed comments and rephrased some words, PTAL @yiwu-arbug @nrc @zhangjinpeng1987

text/2020-04-23-rawkv-txnkv-mode-switching.md
fredchenbj: for the TxnKV and RawKV simultaneously used in one TiKV cluster situation is like that we have two tables, one uses TxnKV API to store metadata for ACID, other uses RawKV API to store data for performance.

text/2020-04-23-rawkv-txnkv-mode-switching.md
```diff
@@ -0,0 +1,63 @@
...
+
+- Store in a special key in the underlying engine and use a new command
+  to manipulate it
+- Use the dynamic configuration feature to achieve our goal.
+
+Since the dynamic configuration is already supported, it should be good for this.
+
+`kv-mode` should be an enum:
+
+- `ModeTxnKV` indicates that the RawKV commands could not be called.
```
breeswish, line 37: Do we still need to allow RawKV access? I guess the RawKV interface was added for debugging purpose at first..

text/2020-04-23-rawkv-txnkv-mode-switching.md
```diff
@@ -0,0 +1,63 @@
...
+  When they are called they will return errors.
+  This is the default value.
+- `ModeRawKV` indicates that the TxnKV commands could not be called.
```
BusyJay, line 40: So users have to choose either TxnKV or RawKV? They can't be mixed anymore?

yiwu-arbug: @BusyJay AFAIK that's the point of this RFC.

BusyJay:
Why merge? I think there are still questions not answered.

text/2020-04-23-rawkv-txnkv-mode-switching.md
BusyJay: Then how can users who have used the mix mode get around the compatibility problem?

yiwu-arbug: @BusyJay If user aware of the the option and choose to use mixed mode, they will notice the compatibility issue. AFAIK the proposal is not to solve the compatibility issue, but to warn user of the problem.

innerr:
@BusyJay My fault, saw 2 LGTMs but didn't see the questions uncovered, we could still discuss here and I will fix all the issues related.

text/2020-04-23-rawkv-txnkv-mode-switching.md
innerr: @fredchenbj Good question. this PR was merged early by mistake(me), we could still discuss here. I thing Sunxiaoguang also mentioned this. Maybe we still need to provide the `MixedMode` (removed by early discuss)

text/2020-04-23-rawkv-txnkv-mode-switching.md
innerr: @BusyJay Maybe we still need to provide the `MixedMode` (removed by early discuss)

PR #50, opened 2020-04-28
encryption-at-rest RFC

yiwu-arbug:
Signed-off-by: Yi Wu <yiwu@pingcap.com>

### DIFF
--- a/text/2020-04-27-encryption-at-rest.md
+++ b/text/2020-04-27-encryption-at-rest.md
@@ -1,0 +1,195 @@
+# Encryption-At-Rest
+
+An earlier version of this proposal can be find at: https://docs.google.com/document/d/14u0YKHXxjE0a9wAd3tsu0Hqjmknjr0bYU_EMzZSbzkA/edit?usp=sharing
+
+## Summary
+
+This is a design proposal to support encryption-at-rest in TiKV. In this document we will discuss design and implementation details of:
+* How to config TiKV to enable encryption and how TiKV handle encryption key rotation.
+* How TiKV encrypt data files through underlying storage engine (i.e. RocksDB)
+* How TiKV handle temporary files (e.g. snapshot files) when encryption is enabled.
+* How TiKV interact with other TiDB tools when encryption is enabled.
+
+## Motivation
+
+Encryption at rest means that data is encrypted when it is stored. As opposed to encryption in flight (TLS) or encryption in use (rarely used). Different things could be doing encryption at rest (SSD drive, file system, cloud vendor, etc), but if the database does  encryption by itself this helps ensure that attackers must authenticate with the database . For example, even when an attacker gains access to the physical machine, he cannot access data by copying files on disk.
+
+A block cipher is an algorithm, that given a symmetry key, encrypts a fixed-sized data block (plaintext) into a block of same size (ciphertext). AES is a block cipher of fixed 128 bits (16 bytes) block. The key size could be 128, 192, 256 bits, referred to as AES128, AES192, AES256 respectively. 
+
+Encrypting data block by block using the same key is insecure because it leaks out patterns of the data. For example, an attacker would know if two blocks have the same ciphertext, it means they have the same plaintext, even if the key is not compromised. Block cipher mode of operation is a kind of algorithm to encrypt each block of data differently so that different blocks with the same plaintext yield different ciphertext. CTR (also called counter mode) is one such algorithm which is commonly used and suitable to support random reads. Each block is encrypted together with its block index, making sure different blocks are encrypted differently. This Wiki page has a very good explanation of the matter and the CTR algorithm: https://en.wikipedia.org/wiki/Block_cipher_mode_of_operation.
+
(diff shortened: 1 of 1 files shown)


yiwu-arbug:
Original proposal can be find at: https://docs.google.com/document/d/14u0YKHXxjE0a9wAd3tsu0Hqjmknjr0bYU_EMzZSbzkA/edit?usp=sharing

Updates include:
* Reorganize the flow of the proposal, to discuss higher level usage first, then the key management details, then specific details about rocksdb and snapshot and BR.
* Updated content for KMS usage
* Update content for snapshot and BR handling.

text/2020-04-27-encryption-at-rest.md
```diff
@@ -0,0 +1,195 @@
...
+The key manager module stores two additional files:
+
+* encryption key dictionary (key.dict): Maps data key id to data keys, together with metadata. Also stores the data key id used to encrypt the encryption file dictionary. The content is encrypted using the master key.
+* encryption file dictionary (file.dict): Maps data file paths to key id and metadata. This file is stored as plaintext.
+
+Except for the encryption key dictionary, TiKV never persists any keys in plaintext or ciphertext.
+
+Both of the files share the same file format, as depicted as follows:
+
++----------------------------------------------+---------------------+
```
overvenus, line 72:
````suggestion
```
+----------------------------------------------+---------------------+
````

File: text/2020-04-27-encryption-at-rest.md [hunk 2]
```diff
@@ -0,0 +1,195 @@
...
+|                   header                     |                     |
++----------+-----------+-----------+-----------+ serialized content, |
+| version  |   flags   |    crc    | file size | could be encrypted  |
+| (1 byte) | (3 bytes) | (4 bytes) | (8 bytes) |  (variable length)  |
++----------+-----------+-----------+-----------+---------------------+
```
overvenus, line 77:
````suggestion
+----------+-----------+-----------+-----------+---------------------+
```
````

File: text/2020-04-27-encryption-at-rest.md [hunk 3]
```diff
@@ -0,0 +1,195 @@
...
+
+This is a design proposal to support encryption-at-rest in TiKV. In this document we will discuss design and implementation details of:
+* How to config TiKV to enable encryption and how TiKV handle encryption key rotation.
+* How TiKV encrypt data files through underlying storage engine (i.e. RocksDB)
+* How TiKV handle temporary files (e.g. snapshot files) when encryption is enabled.
+* How TiKV interact with other TiDB tools when encryption is enabled.
+
+## Motivation
+
+Encryption at rest means that data is encrypted when it is stored. As opposed to encryption in flight (TLS) or encryption in use (rarely used). Different things could be doing encryption at rest (SSD drive, file system, cloud vendor, etc), but if the database does  encryption by itself this helps ensure that attackers must authenticate with the database . For example, even when an attacker gains access to the physical machine, he cannot access data by copying files on disk.
```
overvenus, line 15: Please wrap this sentence into 80 chars per line.
 ```suggestion
Encryption at rest means that data is encrypted when it is stored. As opposed to encryption in flight (TLS) or encryption in use (rarely used). Different things could be doing encryption at rest (SSD drive, file system, cloud vendor, etc), but if the database does encryption by itself this helps ensure that attackers must authenticate with the database. For example, even when an attacker gains access to the physical machine, he/she cannot access data by copying files on disk.
```

File: text/2020-04-27-encryption-at-rest.md [hunk 4]
```diff
@@ -0,0 +1,195 @@
...
+
+Alternatively passing master key stored in a file is also supported, but it is not recommended to use in production. In that case the plaintext key is used directly to encrypt and decrypt (TiKV) data keys.
+
+In both of the cases, AES256-GCM is used to encrypt the (TiKV) data keys. The main reason to use GCM mode instead of CTR mode is not to verify authenticity of the persisted data keys, but rather provide a way to verify if the right master key is provided to decrypt the key. Without using an authentication encryption method like GCM, we don't have reliable way to tell if user provide a different key than the one use previously to encrypt the data keys. The use of AES256 here is non-configurable, in order to reduce configurable knobs that could confuse user.
```
overvenus, line 57:
```suggestion
In both of the cases, AES256-GCM is used to encrypt the (TiKV) data keys. The main reason to use GCM mode instead of CTR mode is not to verify authenticity of the persisted data keys, but rather provide a way to verify if the right master key is provided to decrypt the key. Without using an authentication encryption method like GCM, we don't have a reliable way to tell if user provides a different key than the one used previously to encrypt the data keys. The use of AES256 here is non-configurable, in order to reduce configurable knobs that could confuse user.
```

JaySon-Huang:
Do we support to enable this feature for a cluster already have data in it? Or users should never change this configuration once they have their clusters deployed?
If users want to enable this feature for their existing data, they have to deploy a new cluster with this feature enabled, then migrate data from their old cluster?

gregwebs:
It can be enabled for an existing cluster: https://docs.pingcap.com/tidb/dev/encryption-at-rest

BusyJay:
Should this be merged?

gregwebs:
> Should this be merged?

That would make sense since it is implemented.

yiwu-arbug:
No bandwidth to get this merged. Closing for now.

BusyJay:
Isn't it implemented already in tikv repository?

gregwebs:
Yes, this is implemented now.

BusyJay:
There should be something wrong with the development cycle that implementation is merged before RFC.

gregwebs:
Most development does not have an RFC. I think that we should celebrate that there is one here and try to bring out more rather than feel bad about something.

PR #51, opened 2020-05-14
compare-and-set

huangnauh:


text/2020-05-14-compare-and-set.md
```diff
@@ -0,0 +1,20 @@
...
+
+It allows to conditionally update a key, the condition being that the key has a certain value(maybe not exist).
+
+## Motivation
+
+It allows to implement read-modify-write scenarios.
+
+## Design
+
+- We can run the Compare-And-Set operation in the Scheduler which use Latch to control write concurrency.
```
BusyJay, line 13: Can you add more details? I don't think people can get how it work by two sentences. 😆 

For example, what does the command look like? How does CAS interact with other commands? What changes need to be made for current implement? What does the whole data flow look like?

huangnauh: Changed in https://github.com/tikv/rfcs/pull/51/commits/984955776458967d45bc3af9acdafbcc62617988. Please have another look.

text/2020-05-14-compare-and-set.md
```diff
@@ -0,0 +1,79 @@
...
+    IngestSST = 8;
+    ReadIndex = 9;
+    CheckAndSet = 10;
+}
+```
+
+### RawKVScheduler
+
+```
+pub struct RawLatches {
```
BusyJay, line 50: Why not reuse existing latches?

text/2020-05-14-compare-and-set.md
```diff
@@ -0,0 +1,79 @@
...
+
+
+### Modify::CheckAndSet
+
+```
+pub enum Modify {
+    Delete(CfName, Key),
+    Put(CfName, Key, Value),
+    DeleteRange(CfName, Key, Key, bool),
+    CheckAndSet(CfName, Key, OldValue, NewValue),
```
BusyJay, line 67: It should be mentioned that `CheckAndSet` is not compatible with `DeleteRange`.

text/2020-05-14-compare-and-set.md
```diff
@@ -0,0 +1,79 @@
...
+    ReadIndex = 9;
+    CheckAndSet = 10;
+}
+```
+
+### RawKVScheduler
+
+```
+pub struct RawLatches {
+    slots: Vec<RwLock<Latch>>,
```
BusyJay, line 51: I'm afraid `RwLock` may not be performant enough.

text/2020-05-14-compare-and-set.md
zhangjinpeng1987: Yes, I suggest using the existing latches in the scheduler.

text/2020-05-14-compare-and-set.md
```diff
@@ -0,0 +1,79 @@
...
+}
+```
+
+### RocksDB
+- Merge Operator, the Atomic Read-Modify-Write operation in RocksDB.
+
+
+## Alternatives
+
+- Use Txn Api, get and set
```
zhangjinpeng1987, line 77: Is the RawKV better in your scenario? For the performance reason?

huangnauh: In my scenario,  the client always change or get a single key.  Transaction or 2PC is too heavy.

zhangjinpeng1987:
@Little-Wallace Please confirm if this rfc is out of date?

Xavierj2800 (approved):

PR #56, opened 2020-09-03
propose a feature to factor key prefixes

gregwebs:


text/2020-09-03-factored-prefix.md
```diff
@@ -0,0 +1,71 @@
...
+If a V2 application wants to access the entire V2 key space ('y' first byte prefix), it must set the header "TIKV-PREFIX-ALL".
+This supports the use case of performing a transaction across any prefixes. The prefix feature does nothing to restrict access: this must be done via an auth feature.
+
+A configuration setting will be added to TiKV: "prefix-only". If this setting is set
+* No V1 applications are allowed. It is an error to not set "TIKV-PREFIX".
+
+
+## Drawbacks
+
+It is not possible to query both V1 and V2 applications: in other words to query the 'z' first byte prefix and the 'y' first byte prefix.
```
coocood, line 50: Theoretically, we can let `prefix-all` application access the `z` prefix, just need to send requests via different connections.

But in practice, the SQL syntax can not express key prefixes, and the optimizer doesn't know anything about key prefixes, it's hard to do analytical query across key prefixes.

coocood: If users really need to do analytical queries across key prefix,  they can connect to multiple TiDB servers for different prefixes, and insert the results in a temporary table under a certain prefix, then do the aggregation in that table.

gregwebs: yes, this can be done with this proposal by using two connections. Using 1 connection for a single transaction is possible but would require building an additional feature.

gregwebs: I updated the RFC to reserve a prefix of 0x01 for V1 applications and to signify that the 'z' first byte prefix will be used. This way a V1 TiDB application can be treated as V2 without re-writing the data.

coocood:
LGTM

text/2020-09-03-factored-prefix.md
```diff
@@ -0,0 +1,71 @@
...
+All keys are returned with the first byte prefix and factored prefix stripped.
+
+A factored prefix of 0x00 is not allowed. This helps avoid accidentally using null as a factored prefix.
+
+### API
+
+The factored prefix applies to all keys in a connection.
+So it is set as a GRPC header "TIKV-PREFIX" with the value set to the key prefix.
+
+### Backwards compatibility
```
zhangjinpeng1987, line 30: How about changes in PD?

text/2020-09-03-factored-prefix.md
```diff
@@ -0,0 +1,71 @@
+# Factored Key Prefix
+
+## Summary
+
+TiKV provides a facility where the client can set a key prefix that will be used for all keys in its requests.
+
+## Motivation
+
+TiKV users need to use key prefixes to avoid conflicts.
+In keeping with DRY we should not require a program to add the same prefix everywhere.
```
yiwu-arbug, line 10: what's DRY?

text/2020-09-03-factored-prefix.md
```diff
@@ -0,0 +1,71 @@
+# Factored Key Prefix
+
+## Summary
+
+TiKV provides a facility where the client can set a key prefix that will be used for all keys in its requests.
+
+## Motivation
```
yiwu-arbug, line 7: The motivation is not clear to me. Is it for multi-tenancy? What's the higher level use case look like with multiple prefixes?

text/2020-09-03-factored-prefix.md
```diff
@@ -0,0 +1,71 @@
...
+
+TiKV users need to use key prefixes to avoid conflicts.
+In keeping with DRY we should not require a program to add the same prefix everywhere.
+This will also reduce the payload sent to TiKV.
```
yiwu-arbug, line 11: If the prefix is small enough compare to the total payload, we may not see clear win with the approach.

text/2020-09-03-factored-prefix.md
```diff
@@ -0,0 +1,71 @@
...
+The factored prefix applies to all keys in a connection.
+So it is set as a GRPC header "TIKV-PREFIX" with the value set to the key prefix.
+
+### Backwards compatibility
+
+Currently applications do not formally claim a prefix and there is no way to know what prefixes they might use.
+We can refer to these applications as V1 (version 1) applications. We can refer to applications that use the prefix as V2 (version 2).
+We must avoid mixing the usage of V1 and V2 applications.
+
+V1 applications store their data under a 'z' first byte prefix
```
yiwu-arbug, line 36: how about keeping the 'z' and append the user prefix after the 'z'? that way we don't have a V2.

text/2020-09-03-factored-prefix.md
yiwu-arbug, line 7: If it is use to support multi-tenancy, how does it compare to separating users on different TiKV processes? With multiple processes, we have tools for resource isolation, like systemd to limit CPU, memory and IO, and TiKV configs to limit thread-pool and IO usage on application level. When we mixed data from different users in one single TiKV instance, those tools are not available or hard to implement. But of cause it might make it hard for users to share resources.

text/2020-09-03-factored-prefix.md
gregwebs: Don't Repeat Yourself.

text/2020-09-03-factored-prefix.md
gregwebs: It is not a goal so we can remove it.

text/2020-09-03-factored-prefix.md
gregwebs: This can create conflicts with an existing TiDB application. We would like for all users to be able to upgrade to using key prefixes.

text/2020-09-03-factored-prefix.md
gregwebs: This is better discussed in the key spaces RFC. But separate TiKV processes doesn't scale down well. The approach provided by these RFCs scales down to one region (100 MB).

text/2020-09-03-factored-prefix.md
yiwu-arbug: It doesn't seems to me upgrade ability is required for multi-tenancy. We can always start a new TiKV instance for multi-tenancy setup. Also if for a use case user already have prefixes in the data, he can seamlessly change to have key prefixes support described in this RFC, also without the need to have the headache to have a V2.

text/2020-09-03-factored-prefix.md
yiwu-arbug: yeah, seems that the crucial question is what's the smallest dataset per tenant we want to support.

text/2020-09-03-factored-prefix.md
gregwebs: I have started to do some implementation on this and there are some complications.
If we can make future TiDB prefixes configurable then the situation will be safer. That is, TiDB currently uses 'm' and 't', so a user can reserve these for TiDB when upgrading. If TiDB starts using a new prefix in a new release (lets say it is 'd'), that prefix could conflict with an existing application that is added in the mean time. However, if we make the prefix configurable then the user can configure that prefix so that it doesn't conflict when they upgrade.

gregwebs: So I think we can stick with using just the 'z' first byte prefix at the cost of some upgrade hassle for existing users.

gregwebs: Okay, sticking with the 'z' first byte prefix. The implementation is much simpler and I think this will work out okay for users.

text/2020-09-03-factored-prefix.md
gregwebs: We don't require changes to PD. Over time we may want to further incorporate prefixes into our data model and have them be used by PD.

gregwebs:
@coocood I removed the 'y' first byte prefix. The previous transparent upgrade was nice but made for a more complex implementation. I will test and document how to upgrade an existing installation to use key spaces.

gregwebs:
After hacking at this for a bit I am now questioning whether this is worth the effort. There are a lot of places where prefixes could now be used. But there are also already a lot of places that assume a full key (with the 'z' first byte prefix) is already being given.

My plan is to start on the key spaces and auth proposal and see if it is easier to only add prefixes on the client side.

BusyJay:
So should I review key space first?

coocood:
> After hacking at this for a bit I am now questioning whether this is worth the effort. There are a lot of places where prefixes could now be used. But there are also already a lot of places that assume a full key (with the 'z' first byte prefix) is already being given.
> 
> My plan is to start on the key spaces and auth proposal and see if it is easier to only add prefixes on the client side.

So on a tikv node, each keyspace have its own rocksdb instance?

gregwebs:
> So should I review key space first?
Yes. This proposal is not required.

> So on a tikv node, each keyspace have its own rocksdb instance?
The key space proposal is about logical organization of keys and does not make any changes about organizing storage other than a region should contain only a single key space.

PR #59, opened 2020-09-28
introduce memory trace

BusyJay:
Suspicious memory increase has been a big challenge for TiKV. This PR
introduces a way to analyze what is exhausting the memory easily.

### DIFF
--- a/text/2020-09-28-memory-trace.md
+++ b/text/2020-09-28-memory-trace.md
@@ -64,6 +64,15 @@ impl MemoryTraceManager {
 `register_provider` will store the provider and `snapshot` will create a trace
 using all the information from registered providers.
 
+### Output
+
+A trace is a rust struct, when dumping the result, it needs to be transformed
+into some readable format. For best readability, it can be output as a
+framegraph using the real name. It can also support the existing format from
+[minitrace][1] to utilize more existing UI front end.
+
+[1]: https://github.com/pingcap-incubator/minitrace-rust
+
 ### Implementation example
 
 #### Tracing memory used by peer fsm
--- a/text/2020-09-28-memory-trace.md
+++ b/text/2020-09-28-memory-trace.md
@@ -1,0 +1,122 @@
+# Fully instrumented memory trace
+
+## Summary
+
+Make TiKV fully instrumented to provide accurate and details memory trace.
+
+## Motivation
+
+Suspicious memory increase has been a big challenge for TiKV. We don't have a
+good tool to know who is exhausted the memory. With the help of profile tools,
+we can identify memory leaks that can be reproduced and grow rapidly. However,
+using profiling tools requires a recompilation and also has a performance
+impact. And profile tool can only provide information about allocations, for
+diagnostics, find out who is holding the memory is more useful.
+
+Another use case is to limit the memory usage of a query. This is very useful
+in implementing QoS and avoiding OOM.
+
+## Detailed design
+
+To calculate memory usage, we need to define a memory trace first.
+
+```rust
+pub struct MemoryTrace;
+
+impl MemoryTrace {
+    pub fn add_sub_trace(&mut self, name_id: u64) -> &mut MemoryTrace;
+    pub fn set_size(&mut self, size: usize);
+}
+```
+
+A memory trace is tree that records how much memory its children and itself
+uses. It doesn't need to match any function stacktrace, instead it should have
+logically meaningful layout. For example, memory usage should be divided into
+several components under the root scope: TiDB EndPoint, Transaction, Raft,
+gRPC etc. TiDB EndPoint can divide its children by queries, while Raft can
+divide memory by store and apply. Name are defined as number for better
+performance. In practice, it can be mapped to enumerates instead.
+
+`size` should be the memory the category holds. It should *own* the memory.
+
+To record a trace, we need to define providers.
+
+```rust
+trait MemoryTraceProvider {
+    fn trace(&mut self, dump: &mut MemoryTrace);
+}
+```
+
+The trace method will fill the trace with stats itself and its children have.
+
+To trigger all providers, we need a trace manager.
+
+```rust
+pub struct MemoryTraceManager;
+
+impl MemoryTraceManager {
+    pub fn snapshot(&self) -> MemoryTrace;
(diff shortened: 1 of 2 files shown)


text/2020-09-28-memory-trace.md
```diff
@@ -0,0 +1,131 @@
+# Fully instrumented memory trace
+
+## Summary
+
+Make TiKV fully instrumented to provide accurate and details memory trace.
+
+## Motivation
+
+Suspicious memory increase has been a big challenge for TiKV. We don't have a
```
gregwebs, line 9: Is this heap memory or does the stack run out as well?

text/2020-09-28-memory-trace.md
```diff
@@ -0,0 +1,131 @@
...
+
+Another use case is to limit the memory usage of a query. This is very useful
+in implementing QoS and avoiding OOM.
+
+## Detailed design
+
+To calculate memory usage, we need to define a memory trace first.
+
+```rust
+pub struct MemoryTrace;
```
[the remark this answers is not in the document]
BusyJay: Yes, implementation details are omitted.

text/2020-09-28-memory-trace.md
```diff
@@ -0,0 +1,131 @@
...
+## Drawbacks
+
+Instrumentation can make code complicated.
+
+## Alternatives
+
+We can still keep using the profile tool we used to do and delivering profile
+build by default. It can have uncontrolled performance impact and doesn't have
+a clear view on historical allocations.
+
```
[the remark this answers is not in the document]
BusyJay: The layouts are similar, it's easy to make them compatible.

text/2020-09-28-memory-trace.md
BusyJay: Heap memory.

text/2020-09-28-memory-trace.md
```diff
@@ -0,0 +1,131 @@
...
+## Unresolved questions
+
+How to ensure all memory allocations are traced?
```
breeswish, line 131: Can we have some more generic memory tracing, instead of manually implement one? I'm worried that a manually implemented one may *miss* actual memory leaks since these memory leaks are usually not noticed.

BusyJay: We already have it in TiKV. According to the experiences, it's hard to used and requires recompilation and restart (and extra environment variables).

Even all the limitations are solved without hurting performance hopefully, it's still limited in functionality. For example, it can also miss allocations as stats are sample. It can't provide precise memory analyze by components like what RFC does. It also can't be used in QoS.

However, it's still useful when some places are missed by MemoryTraceProvider. It can be used as the last resort for memory leaks.

gregwebs: Is the reason that it is hard to use because such a a mechanism doesn't properly annotate allocations? Could we annotate allocations in our code (for example wrap Box::new) and would that cover a lot of annotations or would we have many allocations from libraries?

BusyJay: What does "annotate" mean? `Box::new` will call `je_malloc` under the hook and be recorded into profile if sampled. It's hard to use because of recompilation and restart. Enabling profile for a long time may not be practical for online system.

gregwebs: I am wondering if we should replace usages of Box::new with our own wrapper that helps to record memory usage. This would help to ensure we don't miss allocations. But maybe there is no way to make this an improvement over the existing je_malloc handle?

BusyJay: Yes. In my opinion, we care what holds the memory rather than allocating, which can't be solved by tagging allocations.

gregwebs: But if we control tagging allocations we could tag them with information about who holds them?

BusyJay: For example, memory allocated in grpc can go through all the way grpc, transaction and raftstore. Ownership is unpredictable.

breeswish (approved):
I'm fine with the memory tracing. Worth mentioning that this RFC is focused on simplify the diagnostics of where the memory is hold (instead of allocated), which is the main difference compared to existing memory leak facilities.

gregwebs:
It would be great to see an example of related work: some other project where a similar approach was taken.

BusyJay:
The RFC is inspired by the chromium project, and we use a simplified model that only use a global singleton allocator.

The related work can be checked from https://chromium.googlesource.com/chromium/src/+/master/docs/memory-infra/.

text/2020-09-28-memory-trace.md
```diff
@@ -0,0 +1,131 @@
...
+## Motivation
+
+Suspicious memory increase has been a big challenge for TiKV. We don't have a
+good tool to know who is exhausted the memory. With the help of profile tools,
+we can identify memory leaks that can be reproduced and grow rapidly. However,
+using profiling tools requires a recompilation and also has a performance
+impact. And profile tool can only provide information about allocations, for
+diagnostics, find out who is holding the memory is more useful.
+
+Another use case is to limit the memory usage of a query. This is very useful
```
gregwebs, line 16: How would we use this to limit the memory usage of a query? It seems the proposal is to take occasional snapshots. So if a query held memory for long enough to show up in a snapshot we would limit it? What would the snapshot frequency be if we were to use this proactively rather than just for debug? What would the overhead be?

BusyJay: A query can check the current memory usage after scanning constant row count. The cost should be small enough as the memory needs to be collected can be calculated for every result row, which is just query the encoded byte array's length. And the memory usage should be able to be caught in time because memory usage for a row is tiny generally.

gregwebs: This seems like it might be implemented separately from this proposal? Or is the query asking for a snapshot of its own usage after every row count?

BusyJay: Stats collection is implemented in this proposal, but query and abort should be implemented in other proposal like QoS.

BusyJay:
A POC has been implemented as tikv/tikv#9019, a leak case has been found by the POC as tikv/tikv#10037.

5kbpers (approved):
LGTM

zhangjinpeng1987 (approved):
LGTM

PR #62, opened 2020-12-16
RFC: RBAC Support for PD

PhotonQuantum:


text/2020-12-16-pd-rbac.md
```diff
@@ -0,0 +1,150 @@
...
+Utilities requiring direct access to PD may gain limited access to it with a user
+created earlier by an administrator.
+
+## Detailed design
+
+### Authentication
+
+- PD stores `User`s in a KV database (etcd).
+  - Username: string
+  - Hash: string (sha256)
```
nrc, line 37: I assume this is the hash of the password? In which case sha256 is inappropriate. You should either salt the password and use multiple iterations of sha512, or use a hash algorithm specifically designed for passwords, e.g., Argon2d.

text/2020-12-16-pd-rbac.md
```diff
@@ -0,0 +1,150 @@
...
+  - Roles: a set of strings (role names)
+- A client (TiDB/TiKV/pd-ctl/…) must provide a pair of username and password to
+  gain access to PD APIs (including HTTP and gRPC endpoints).
+  - Clients using HTTP APIs should send requests with a basic authentication header.
```
nrc, line 41: Can users always use https to contact PD? Using http for this would be insecure.

text/2020-12-16-pd-rbac.md
```diff
@@ -0,0 +1,150 @@
...
+  - Clients using gRPC should set its username and password in the meta field.
```
nrc, line 42: Likewise, do we always use TLS for gRPC?

text/2020-12-16-pd-rbac.md
```diff
@@ -0,0 +1,150 @@
...
+outcome of a security breach.
+
+Utilities requiring direct access to PD may gain limited access to it with a user
+created earlier by an administrator.
+
+## Detailed design
+
+### Authentication
+
+- PD stores `User`s in a KV database (etcd).
```
nrc, line 35: Will this require an etcd access on every request? That seems like it could be a big performance hit (though I am just guessing).

text/2020-12-16-pd-rbac.md
```diff
@@ -0,0 +1,150 @@
...
+control on PD APIs.
+
+The idea of authenticating with certificates and its design of tokens is great though.
+We may combine it with the design proposed with this RFC in the future by making
+minor modifications to the administrative APIs and data models, and implementing
+a token issuing and refreshing mechanism.
+
+## Future Additions
+
+- Token-based authentication.
```
nrc, line 149: token-based authorization is considerably more secure than using username/password and IMO should be the only mechanism supported. There are fairly standard approaches which can be used, e.g., OAuth

text/2020-12-16-pd-rbac.md
PhotonQuantum: That's a good idea! We may enforce TLS when authentication is enabled.

text/2020-12-16-pd-rbac.md
PhotonQuantum: I believe an in-memory cache can mitigate the performance regression.

text/2020-12-16-pd-rbac.md
PhotonQuantum: Maybe different users will have different security models? I'm thinking about providing several auth methods, including password-based, token-based, and certificate-based ones.

text/2020-12-16-pd-rbac.md
PhotonQuantum: I'm worried about the potential performance penalty if we use key derivation algorithms.

nrc: That is a feature, not a bug - being able to hash the password quickly is a security risk. I think a token based scheme would be more performant?

text/2020-12-16-pd-rbac.md
nrc: From a security perspective, the more auth methods there are, the less secure your system is. Could we support a single, token-based auth method and provide token generation from username/password in the client?

text/2020-12-16-pd-rbac.md
```diff
@@ -0,0 +1,150 @@
...
+
+Here we propose adding RBAC to PD so that all requests to PD must be authenticated,
+and clients may only do actions they are explicitly authorized to perform.
+
+This RFC is a partial implementation of [tikv/8621](https://github.com/tikv/tikv/issues/8621)
+that describes a way to support RBAC-based data access control for TiKV and PD.
+
+## Motivation
+
+PD server exposes its APIs to the network via HTTP or gRPC endpoints.
```
zz-jason, line 13: are there any guidelines about when to use HTTP and gRPC?

PR #63, opened 2021-02-24
RFC: Coprocessor Plugin

andylokandy:
Add a general and pluggable coprocessor framework for RawKV mode.

[Rendered](https://github.com/andylokandy/rfcs/blob/plugin/text/0063-coprocessor-plugin.md)


text/2021-02-24-coprocessor-plugin.md
```diff
@@ -0,0 +1,145 @@
+# Coprocessor Plugin
+
+## Summary
+
+Add a general and plugable coprocessor framework for RawKV mode.
```
skyzh, line 5: plugable  -> pluggable

text/2021-02-24-coprocessor-plugin.md
```diff
@@ -0,0 +1,145 @@
...
+
+## Motivation
+
+TiKV is the storage component in the TiDB ecosystem, however, the distribution computation principle suggests that computation should be as close to the data source as possible. Therefore, TiKV has embedded a subset of the TiDB executor framework to push down some computation tasks when applicable.
+
+But TiKV's capability should be far beyond that, as many distributed components can be built on top of TiKV, such as cache, full text search engine, gragh database and NoSQL database. And same as TiDB, these product will also like to push down specific computation to TiKV, which requires the coprocessor to be customizable, aka plugable.
```
skyzh, line 11: gragh -> graph

text/2021-02-24-coprocessor-plugin.md
```diff
@@ -0,0 +1,145 @@
...
+
+## Detailed design
+
+### Dynamic vs statically
+
+Generally, there are two strategies to build a plugin framework: dynamically and statically, which means to load the plugin on startup or to embed in the binary on compilation.
+
+![Plugin Arch](../media/plugin-arch.png)
```
skyzh, line 19: `plugin-arch.png` seems to be an empty file.

text/2021-02-24-coprocessor-plugin.md
```diff
@@ -0,0 +1,145 @@
...
+
+    Previously, a proof-of-concept plugin experiment [[repo]](https://github.com/andylokandy/plugin) is done. As a result, we found that Rust's unstable ABI is a risk in safety. It's hard to guarantee that the TiKV and the plugin has absolutely the same ABI version, and is also hard to debug such a plugin.
+
+- Lua
+
+    TOO SLOW
+
+- eBPF
+
+    Berkeley Packet Filter has been experimented in Hackthon. It requires the plugin to be written in C, which is not capable to migrate `tidb_query` eventually. And also, eBPF is not turning-complete, which is unacceptable.
```
skyzh, line 49: Berkeley -> Berkerley, Hackthon -> Hackathon, turning-complete -> turing-complete

skyzh (commented):
I'm curious about what functionalities the coprocessor plugin should provide when we build this API in the following months.

skyzh:
> I'm curious about what functionalities the coprocessor plugin should provide when we build this API in the following months.

If I understand correctly, we will get a simple plugin that supports simple scanning and defines primitive types after implementing this RFC.

text/2021-02-24-coprocessor-plugin.md
andylokandy: It's `Berkeley`: https://en.wikipedia.org/wiki/Berkeley_Packet_Filter

text/2021-02-24-coprocessor-plugin.md
andylokandy: I can see it on https://github.com/tikv/rfcs/blob/c180a9a4e1b5b5515b31a7fb7d77844ef707b2d9/text/2021-02-24-coprocessor-plugin.md

andylokandy:
> I'm curious about what functionalities the coprocessor plugin should provide when we build this API in the following months.

> If I understand correctly, we will get a simple plugin that supports simple scanning and defines primitive types after implementing this RFC.

I think in merely 3 months we can only implement the framework itself. A plugin prototype for testing is good to have.

text/2021-02-24-coprocessor-plugin.md
```diff
@@ -0,0 +1,145 @@
...
+pub enum Error {
+    KeyNotInRegion { key: Key, region: Region },
+    Deadlock { key: Key },
+    // More
+}
+
+pub type Result<T> = std::result::Result<T, Error>;
+
+#[async_trait]
+pub trait RawTransaction: Send {
```
zhongzc, line 97: I prefer the abstraction level which contains `Snapshot` and `WriteBatch`. Then I don't need to consider whether values returned from read operations are newest or from a snapshot.
If former, what will it return when reading the key just written via `put`? Also, the term `Transaction` would cause misleading.
If latter, will it always refresh the snapshot at a lock is acquired?

andylokandy: I think `put()` should always require a lock.

andylokandy: And the `get()` underling should scan the write batch first and then the snapshot.

zhongzc: Can I read committed?

andylokandy: Do you mean RC isolation? I think RocksDB's snapshot won't allow you to do so.

zhongzc: So I cannot even implement a counter.

text/2021-02-24-coprocessor-plugin.md
```diff
@@ -0,0 +1,145 @@
...
+
+Transaction mode is way more complicated than the RawKV mode. The current picture of txn plugin is an 'interactive' component with the client, which means, the client and the coprocessor will talk to each others during the request being processed. This is especially for resolving locks and coordinating the transaction among multiple coprocessors in multiple TiKV nodes.
+
+### Static plugin
+
+Discussed in the section above.
+
+### Migrate `tidb_query`
+
+When static plugin and txn plugin is implemented, we can move the coprocessor specific to TiDB out of TiKV, and the repository of TiDB might be a good new home.
```
nrc, line 145: Have you evaluated if this is possible using the proposed framework?

text/2021-02-24-coprocessor-plugin.md
```diff
@@ -0,0 +1,145 @@
...
+| X Build very slow | ◯ Build fast |
+| X Hard to debug | ◯ Easy to debug |
+
+Ideally, we'd like to develop a plugin in dynamic mode, and eventually, to distribute the statically linked one in release. So there is an interesting research area that designing a plugin framework that can write the plugin once, and compile to the dynamic one and the static one with duplicate code. Anyway, this is an ambitious target but sort of out of scope, so we may explore the possibility when marching on.
+
+So initially, in this RFC, we'll only focus on the dynamic plugin framework that works in RawKV mode.
+
+### Web Assembly
+
+Web Assembly is chosen to host the dynamic plugin. There was alternatives like dynamic lib, lua and bpf. So in this section, I'll explain the tradeoff between them and why WASM is the most appropriate choice for TiKV:
```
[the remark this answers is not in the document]
andylokandy: > What metadata is required 

WASM plugin should provide its name, plugin version and the API version.

> How do we know a plugin has not been tampered with

No guarantee.

> how does a TiKV client know what plugins are being run? What happens if TiDB (or other client) is expecting a plugin to be present but it isn't?

The client should specify the name and also the version constraint of the plugin. The request will be rejected if that specified plugin isn't loaded.

> What if different TiKV nodes have different plugin versions?

It should be fine if that plugins of different versions satisfy the version constraint of client.

text/2021-02-24-coprocessor-plugin.md
```diff
@@ -0,0 +1,145 @@
...
+
+- WASM
+
+    Web Assembly has also been experimented in Hackathon. As the result, it made the performance score around 50% to 80% to the statically linked one. Good work for `Wasmer`! Besides, WASM can be written in many languages and has great safety guarantee.
+
+    The Rust binding for WASM plugin should be first-class supported.
+
+### Keyspace
+
+Keyspace[[RFC]](https://github.com/tikv/rfcs/pull/39)[[The most updated design doc]](https://docs.google.com/document/d/1x17-urAqToDo8TVXJroEHtc76fdssFaoANjSaNDhjKg/edit) is an incoming feature of TiKV that is highly related to coprocessor plugin. Keyspace determines whether a range of key should only be used in transaction mode or in RawKV mode. Since coprocessor works in either RawKV mode or txn mode, surely coprocessor plugin framework should aware of Keyspace. The details is TBD.
```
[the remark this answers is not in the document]
andylokandy: Yes, it does. I mean the plugin framework should aware of the keyspace, not the plugin itself.

text/2021-02-24-coprocessor-plugin.md
```diff
@@ -0,0 +1,145 @@
...
+
+    /// Returns when Raft message applied successfully.
+    async fn commit(self) -> Result<()>;
+}
+
+pub trait Coprocessor: Send + Sync {
+    fn on_raw_coprocessor_request(
+        &self,
+        region: Region,
+        request: Vec<u8>,
```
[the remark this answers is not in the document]
andylokandy: Yes, exactly. The plugin can use whatever it like to communicate with the client.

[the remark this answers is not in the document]
andylokandy: `tidb_query` has already been reading raw bytes from the coprocessor request.

text/2021-02-24-coprocessor-plugin.md
andylokandy: Not yet. This is the ultimate goal, but it will not block this RFC if this is not possible eventually. I think the most tricky part is the static plugin. Once we've figured it out, it will be quite trivial to do the migration.

text/2021-02-24-coprocessor-plugin.md
```diff
@@ -0,0 +1,145 @@
...
+    async fn delete_range(&mut self, key_range: Range<Key>) -> Result<()>;
+
+    /// Returns when Raft message applied successfully.
+    async fn commit(self) -> Result<()>;
+}
+
+pub trait Coprocessor: Send + Sync {
+    fn on_raw_coprocessor_request(
+        &self,
+        region: Region,
```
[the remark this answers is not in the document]
andylokandy: It provide the bounary of this request. For example, the request may says 'scan all documents that contains `state monad` in this region', then the coprocessor should be careful that not to scan beyond the region bounary or else `Err(KeyNotInRegion)` will be returned.

text/2021-02-24-coprocessor-plugin.md
```diff
@@ -0,0 +1,145 @@
...
+### Keyspace
+
+Keyspace[[RFC]](https://github.com/tikv/rfcs/pull/39)[[The most updated design doc]](https://docs.google.com/document/d/1x17-urAqToDo8TVXJroEHtc76fdssFaoANjSaNDhjKg/edit) is an incoming feature of TiKV that is highly related to coprocessor plugin. Keyspace determines whether a range of key should only be used in transaction mode or in RawKV mode. Since coprocessor works in either RawKV mode or txn mode, surely coprocessor plugin framework should aware of Keyspace. The details is TBD.
+
+### Multi-plugin
+
+Currently TiKV has only one coprocessor `tidb_query`. However, without further work on statically linked plugin and txn mode support, we can't strip it from official release. So, multiple coprocessor has to be supported. Basically, we may need to add a `gPRC` rpc for coprocessor v2 request, in which coprocessor name and version is given, so that TiKV will be able to dispatch the request to the proper coprocessor, as well as to reject the request on version mismatch.
+
+### API design
+
```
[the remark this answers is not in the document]
andylokandy: > What invariants can TiKV authors assume about plugins? What must plugins always be prevented from doing? Or must always do? Do plugins have access to data in a state which is not usually visible? Can plugins leave data in a state which TiKV cannot handle?

The goal is to make the plugin fool-proof. So the guarantee and invariants should be similar to the client.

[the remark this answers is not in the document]
andylokandy: > What are the performance implications of the API

It's merely a wrapper on the existing `WriteBatch` and `Snapshot` from the `Engine` trait, with additional memory locks which is the only performance overhead.

The memory locks is necessary because the plugin performs interactively, which make it impossible know all mutations in advance, so as to set the latch automatically like what `KvPut` does.

text/2021-02-24-coprocessor-plugin.md
```diff
@@ -0,0 +1,145 @@
...
+    async fn get(&self, key: Key) -> Result<Option<Value>>;
+    async fn scan(&self, key_range: Range<Key>) -> Vec<Value>;
+
+    async fn put(&mut self, key: Key, value: Value) -> Result<()>;
+    async fn delete(&mut self, key: Key) -> Result<()>;
+    async fn delete_range(&mut self, key_range: Range<Key>) -> Result<()>;
+
+    /// Returns when Raft message applied successfully.
+    async fn commit(self) -> Result<()>;
```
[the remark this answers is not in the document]
andylokandy: Updated in the doc.

text/2021-02-24-coprocessor-plugin.md
andylokandy: Updated the section

text/2021-02-24-coprocessor-plugin.md
```diff
@@ -0,0 +1,145 @@
+# Coprocessor Plugin
+
+## Summary
+
+Add a general and pluggable coprocessor framework for RawKV mode.
+
+## Motivation
```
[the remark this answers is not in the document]
andylokandy: Updated the doc

text/2021-02-24-coprocessor-plugin.md
```diff
@@ -0,0 +1,145 @@
...
+
+    The Rust binding for WASM plugin should be first-class supported.
+
+### Keyspace
+
+Keyspace[[RFC]](https://github.com/tikv/rfcs/pull/39)[[The most updated design doc]](https://docs.google.com/document/d/1x17-urAqToDo8TVXJroEHtc76fdssFaoANjSaNDhjKg/edit) is an incoming feature of TiKV that is highly related to coprocessor plugin. Keyspace determines whether a range of key should only be used in transaction mode or in RawKV mode. Since coprocessor works in either RawKV mode or txn mode, surely coprocessor plugin framework should aware of Keyspace. The details is TBD.
+
+### Multi-plugin
+
+Currently TiKV has only one coprocessor `tidb_query`. However, without further work on statically linked plugin and txn mode support, we can't strip it from official release. So, multiple coprocessor has to be supported. Basically, we may need to add a `gPRC` rpc for coprocessor v2 request, in which coprocessor name and version is given, so that TiKV will be able to dispatch the request to the proper coprocessor, as well as to reject the request on version mismatch.
```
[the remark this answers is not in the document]
andylokandy: Updated in doc

text/2021-02-24-coprocessor-plugin.md
```diff
@@ -0,0 +1,145 @@
...
+
+- eBPF
+
+    Berkeley Packet Filter has been experimented in Hackathon. It requires the plugin to be written in C, which is not capable to migrate `tidb_query` eventually. And also, eBPF is not turing-complete, which is unacceptable.
+
+- WASM
+
+    Web Assembly has also been experimented in Hackathon. As the result, it made the performance score around 50% to 80% to the statically linked one. Good work for `Wasmer`! Besides, WASM can be written in many languages and has great safety guarantee.
+
+    The Rust binding for WASM plugin should be first-class supported.
```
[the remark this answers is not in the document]
andylokandy: Updated in doc

text/2021-02-24-coprocessor-plugin.md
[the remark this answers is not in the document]
andylokandy: > why are the 'building up' methods async?

It's because every mutation will acquire the memory lock.

andylokandy:
@nrc `RawTransaction` is removed from this RFC. It deserves an individual RFC later.

text/2021-02-24-coprocessor-plugin.md
```diff
@@ -0,0 +1,161 @@
...
+| ◯ Easy to deploy | X Complexify the deploy process |
+| X Build the entire TiKV | ◯ Easy to build |
+| X Build very slow | ◯ Build fast |
+| X Hard to debug | ◯ Easy to debug |
+
+In this RFC, we'll only focus on the dynamic plugin framework that works in RawKV mode.
+
+### Plugin runtime
+
+The plugin runtime is a new component settling in `tikv::server::service::kv::Service`. It loads the dylib plugin, dispatches coprocessor request to the plugin, and proxy the API calls from plugin to the [`Storage`](https://tikv.github.io/doc/tikv/storage/struct.Storage.html).
```
zz-jason, line 49: could you elaborate more about the implementation of the plugin runtime? for example:

- how to implement a plugin under this framework?
- how to compile a plugin to a dynamic library?
- how to load the dynamic plugin library to the tikv-server process?
- is it possible to upgrade the dynamic plugin to a newer version without restarting the tikv-process?

text/2021-02-24-coprocessor-plugin.md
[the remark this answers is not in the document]
zz-jason: for the API spec, what's the expected performance reduction?

text/2021-02-24-coprocessor-plugin.md
```diff
@@ -0,0 +1,161 @@
...
+        &self,
+        region: Region,
+        request: Vec<u8>,
+        storage: Box<dyn RawStorage>,
+    ) -> Result<Vec<u8>>;
+```
+
+### Keyspace
+
+Keyspace[[RFC]](https://github.com/tikv/rfcs/pull/39)[[The most updated design doc]](https://docs.google.com/document/d/1x17-urAqToDo8TVXJroEHtc76fdssFaoANjSaNDhjKg/edit) is an incoming feature of TiKV that is highly related to coprocessor plugin. Keyspace determines whether a range of key should only be used in transaction mode or in RawKV mode. Since coprocessor works in either RawKV mode or txn mode, surely coprocessor plugin framework should aware of Keyspace. The details is TBD.
```
zz-jason, line 137: should we consider the coprocessor plugin after the key range feature is implemented?

text/2021-02-24-coprocessor-plugin.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Future work
```
zz-jason, line 139: could you elaborate more about the testing, for example:

- correctness testing, to ensure the plugin framework API is working as expected.
- performance testing, to ensure the performance is as expected.

text/2021-02-24-coprocessor-plugin.md
andylokandy: It would be the best if the keyspace is done. But it's also fine to implement coprocessor plugin in advance since we're assuming the user takes responsibility on the key range seperation.

text/2021-02-24-coprocessor-plugin.md
andylokandy: correctness testing is done by the example plugin, which, when plugged in, can be test against the test suit for it. performance testing is the same.

andylokandy:
@zz-jason 

> for the API spec, what's the expected performance reduction?

As we don't even have a coprocessor plugin before, there will be no *reduction*, and since coprocessor plugin moves the computation from client to the tikv, the performance should be improved comparing to the client.

text/2021-02-24-coprocessor-plugin.md
andylokandy: > how to implement a plugin under this framework?
> how to compile a plugin to a dynamic library?
> how to load the dynamic plugin library to the tikv-server process?

These are the main research goal ongoing in the LFX program, I could update the RFC after we're satisfied by a solution. However, here is an [poc project](https://github.com/andylokandy/plugin) that may answers these question for now.

> is it possible to upgrade the dynamic plugin to a newer version without restarting the tikv-process?

This is a good question, I think it's viable but I've though about it before. I'd try to achieve that during the program.

text/2021-02-24-coprocessor-plugin.md
```diff
@@ -0,0 +1,161 @@
...
+
+Add a general and pluggable coprocessor framework for RawKV mode.
+
+## Motivation
+
+TiKV is the storage component in the TiDB ecosystem, however, the distribution computation principle suggests that computation should be as close to the data source as possible. Therefore, TiKV has embedded a subset of the TiDB executor framework to push down some computation tasks when applicable.
+
+But TiKV's capability should be far beyond that, as many distributed components can be built on top of TiKV, such as cache, full text search engine, graph database and NoSQL database. And same as TiDB, these product will also like to push down specific computation to TiKV, which requires the coprocessor to be customizable, aka pluggable.
+
+For instance, a full-text seraching engine will persist the origin document and n-gram index on TiKV. It'll be a waste of resource if we read back and then update the index from a client. In contrary, the coprocessor plugin can generate the index from the origin document, and update the index inplace. What's more, the coprocessor plugin can perform index scan directly on TiKV.
```
BusyJay, line 13: s/In contrary/On the contrary/

text/2021-02-24-coprocessor-plugin.md
```diff
@@ -0,0 +1,161 @@
...
+
+The goals of the coprocessor plugin are:
+
+- Do what client can do (on single region)
```
BusyJay, line 17: Not necessary limited to one region. Though should be simple to start with single region first.

text/2021-02-24-coprocessor-plugin.md
```diff
@@ -0,0 +1,161 @@
...
+
+### Keyspace
+
+Keyspace[[RFC]](https://github.com/tikv/rfcs/pull/39)[[The most updated design doc]](https://docs.google.com/document/d/1x17-urAqToDo8TVXJroEHtc76fdssFaoANjSaNDhjKg/edit) is an incoming feature of TiKV that is highly related to coprocessor plugin. Keyspace determines whether a range of key should only be used in transaction mode or in RawKV mode. Since coprocessor works in either RawKV mode or txn mode, surely coprocessor plugin framework should aware of Keyspace. The details is TBD.
+
+## Future work
+
+### Official coprocessor
+
+Provide an official coprocessor plugin which defines common primitives like `bool`, `f32`, `UTF8String`, `Array<T>` along with codec for them, moreover, implements simple scanning expressions or even aggregations.
```
BusyJay, line 143: Why not just define them in SDK?

skyzh: First of all, encoding of `bool`, `f32`, `UTF8String`, `Array<T>` depends on the library which plugin uses (e.g. protobuf, Apache Arrow, etc.). At the same time, different plugins may process and interpret raw byte data from clients in different ways. From my perspective, it is the plugins' job to define how to interpret raw byte data, and what types they could process. With this design, SDK only needs to provide byte data, and the types should be specified in plugin itself.

Maybe someday we would find out that all plugins are using these same set of types, and lift those types from the plugin to the SDK?

andylokandy:
@BusyJay PTAL

andylokandy:
@BusyJay PTAL

andylokandy:
@BusyJay @skyzh @zz-jason PTAL, I'm going to merge it since it's implemented.

skyzh (approved):
LGTM

text/2021-02-24-coprocessor-plugin.md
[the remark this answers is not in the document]
andylokandy: There is no overhead on any existing functionalities.

text/2021-02-24-coprocessor-plugin.md
```diff
@@ -0,0 +1,241 @@
+# Coprocessor Plugin
+
+## Summary
+
+Add a general and pluggable coprocessor framework for RawKV mode.
```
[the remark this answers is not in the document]
andylokandy: That deserves another RFC IMO.

text/2021-02-24-coprocessor-plugin.md
```diff
@@ -0,0 +1,241 @@
...
+
+- Easy to use
+  - Out of box
+  - Easy to deploy
+
+- Robust
+  - Easy to debug
+  - Log support, metrics support
+
+## Detailed design
```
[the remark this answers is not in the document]
zz-jason: you can write the alternatives (not taken into implementation but meaningful for brainstorming and comparison) to the last sections, for example, the static mode for the plugin regrestering.

text/2021-02-24-coprocessor-plugin.md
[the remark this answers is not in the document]
zz-jason: could you illustrate more about the effort to implement this framework on transactional kv?

zz-jason:
> @zz-jason
> 
> > for the API spec, what's the expected performance reduction?
> 
> As we don't even have a coprocessor plugin before, there will be no _reduction_, and since the coprocessor plugin moves the computation from client to the tikv, the performance should be improved comparing to the client.

For a transactional plugin, you can test the performance on the current TiKV coprocessor. For a non-transactional plugin, you can benchmark the performance of the API call.

andylokandy:
> For a transactional plugin, you can test the performance on the current TiKV coprocessor. For a non-transactional plugin, you can benchmark the performance of the API call.

The plugin performance could vary on different workloads. For instance, a plugin that updates a key in place is expected to outperform the trivial client get and put, while the plugin may be slower on point get. Nevertheless, the actual benchmark should not be included in an RFC IMO, but it should be done in the following GA process.

zz-jason:
> The plugin performance could vary on different workloads. For instance, a plugin that updates a key in place is expected to outperform the trivial client get and put, while the plugin may be slower on point get. Nevertheless, the actual benchmark should not be included in an RFC IMO, but it should be done in the following GA process.

By "performance", I mean the performance of the framework itself, for example:

- the performance to load a custom plugin
- the performance of a function call on the plugin API

Not the end-to-end time to execute a workload on a custom plugin, which depends on the calculation logic inside the plugin.

text/2021-02-24-coprocessor-plugin.md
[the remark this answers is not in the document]
andylokandy: Almost done, but I ignored `performance analysis, is there any performance penalty to use coprocessor plugin` because no benchmark no talk, and also ignored `milestones and sub-tasks` since it has been implemented.

andylokandy:
> * the performance to load a custom plugin

Need benchmark; should be within milliseconds

> * the performance of a function call on the plugin API

The dynamic dispatch typically takes 2 nanoseconds.

text/0063-coprocessor-plugin.md
```diff
@@ -0,0 +1,323 @@
...
+    /// Returns [`Result::Ok]` if the key was successfully deleted.
+    async fn delete(&self, key: Key) -> PluginResult<()>;
+
+    /// Same as [`RawStorage::delete()`], but deletes multiple key-value pairs at once.
+    async fn batch_delete(&self, keys: Vec<Key>) -> PluginResult<()>;
+
+    /// Same as [`RawStorage::delete()`], but deletes multiple key-values pairs at once
+    /// given a `key_range`. All records with keys in `[key_range.start, key_range.end)`
+    /// will be deleted. The upper bound of the `key_range` is exclusive.
+    async fn delete_range(&self, key_range: Range<Key>) -> PluginResult<()>;
```
zz-jason, line 266: how about separating the supporting of the coprocessor plugin into 2 phases:
- the first phase is to support read-only APIs, enable TiKV users to add custom key-value codecs, data types (models), and calculate function and models.
- the second phase is to support write-related APIs, enable tiKV users to pushdown custom write operations, improve the write performance

text/0063-coprocessor-plugin.md
```diff
@@ -0,0 +1,323 @@
...
+message RawCoprocessorResponse {
+    errorpb.Error region_error = 1;
+    // Error message for cases like if no coprocessor with a matching name is found
+    // or on a version mismatch between plugin_api and the coprocessor.
+    string error = 2;
+    bytes data = 3;
+}
+```
+
+### Plugin API design
```
zz-jason, line 184: except for the API that a plugin needs to implement, on the user side, how to:

- manage plugins: observe existing plugins, register new plugins, deregistration existing plugins, plugin privileges (plugins can be added by different people for different purposes)
- utilize plugins in the TiKV client API: how to scan keys that satisfy a user-defined filer based on user-defined data types and key-value codecs?

skyzh:
Can we get this merged?

PR #64, opened 2021-04-20
An RFC requires tracking issue

tisonkun:
So far it is a bit awkward for a newcomer to understand how a feature gets implemented after it is approved and started.

We requires creating a tracking issue for the RFC and close when it is rejected or completed. Of course, updated on progress made.

What do you think?

cc @BusyJay @nolouch @HunDunDM @disksing  @andylokandy


andylokandy:
I suggest swapping step 1 and step 2. There is no need for a tracking issue to a very initialized RFC.

tisonkun:
> I suggest swapping step 1 and step 2. There is no need for a tracking issue to a very initialized RFC.

make sense. updated.

nrc:
> We requires creating a tracking issue for the RFC

When the RFC is accepted? That would make sense to track implementation. Since the RFC is a PR until it is accepted, there is no point having an issue as well as an RFC before it is accepted.

It would also make sense to let users file issues for things which they think should be RFCs, to give a space to discuss before taking the time/effort to write a full RFC.

tisonkun:
> > We requires creating a tracking issue for the RFC
> 
> When the RFC is accepted? That would make sense to track implementation. Since the RFC is a PR until it is accepted, there is no point having an issue as well as an RFC before it is accepted.

Yes as mentioned [here](https://github.com/tikv/rfcs/pull/64/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R24).

> It would also make sense to let users file issues for things which they think should be RFCs, to give a space to discuss before taking the time/effort to write a full RFC.

Yes, here we focus on the requirement for an effort requires an RFC. The pre-RFCs story will be updated later as we discuss about where it would take place. Hopefully we reach a consensus and add the information as [the rust-lang team mentions their ways](https://github.com/rust-lang/rfcs/blame/master/README.md#L90-L94)

README.md
```diff
@@ -12,16 +12,16 @@ stakeholders can be confident about the direction the project is evolving in.
...
 3. Submit a pull request.
 
 ## Timeline of an RFC
 
 1. An RFC is submitted as a PR.
 2. Discussion takes place, and the text is revised in response.
-3. The PR is merged or closed when at least two project maintainers reach
-   consensus.
+3. The PR is accepted or rejected when at least two project maintainers reach consensus.
+4. If accepted, create a tracking issue, refer it in the RFC, and finalize by merging.
```
BusyJay, line 24: We may need bot to create it automatically.

andylokandy: FYI, in the rust community, rfc is merged once it is accepted. The tracking issue can be created later.

tisonkun: It is a bit tricky because we don't have a central repository and either a tracking issue under

- tikv/tikv
- tikv/pd
- tikv/raft-rs

I don't see a good automatic way and think it can be a later discussion. The problem is we want a way to tracking how a feature gets implemented.

tisonkun: > FYI, in the rust community, rfc is merged once it is accepted. The tracking issue can be created later.

@andylokandy Yes it works. I write as is because we want a way to tracking how a feature gets implemented. And after merging the RFC, creating a tracking issue, and ask the contributor to update the RFC think a bit awkward.

nrc: When the RFC is ready to merge, the person merging it can create a tracking issue in the right repo, add the link to the PR, then merge it.

tisonkun: > When the RFC is ready to merge, the person merging it can create a tracking issue in the right repo, add the link to the PR, then merge it.

Yes. That is exactly what in my mind.

tisonkun:
@BusyJay @nrc do you guys committers can help this pull request merged or we should involve others for the following progress.

BusyJay (approved):
LGTM

README.md
```diff
@@ -12,16 +12,16 @@ stakeholders can be confident about the direction the project is evolving in.
...
 2. Write the document and fill in the blanks.
 3. Submit a pull request.
 
 ## Timeline of an RFC
 
 1. An RFC is submitted as a PR.
 2. Discussion takes place, and the text is revised in response.
-3. The PR is merged or closed when at least two project maintainers reach
-   consensus.
+3. The PR is accepted or rejected when at least two project maintainers reach consensus.
```
nrc, line 23: The previous language was chosen to be less confrontational

README.md
```diff
@@ -12,16 +12,16 @@ stakeholders can be confident about the direction the project is evolving in.
 
 ## How to submit an RFC
 
-1. Copy `template.md` into `text/YYYY-MM-DD-my-feature.md`.
+1. Copy `template.md` into `text/PRID-my-feature.md`.
```
nrc, line 15: This change seems unrelated to the rest of this PR

nrc (commented):
Changing to require a tracking issue seems good. I think it is better to pull out the change from date to PR number to a separate PR, since that has not had any discussion here.

README.md
tisonkun: I will revert these changes and let's discuss on another PR. Briefly I think it is easy for someone pronounces/refers `RFC-PRID` instead of `RFC-YYYY-MM-DD` or `RFC-my-feature`.

README.md
tisonkun: With `4` now a tracking issue is required on accepted case so that we cannot interchange these words. In the other way, a proposal is rejected by technical reason is not offensive in my opinion.

andylokandy (approved):

zhangjinpeng1987 (approved):
LGTM

tisonkun:
@BusyJay @zhangjinpeng1987 could you help on the merging process?

PR #65, opened 2021-04-23
RFC: Builtin GC worker

andylokandy:
## Summary

Move TiKV MVCC GC worker from TiDB into PD and make it general for use without TiDB.

## Motivation

GC worker is an important component for TiKV that deletes outdated MVCC data so as to not explode the storage. But currently, the GC worker is implemented in TiDB, which makes TiKV not usable without TiDB.

[Rendered](https://github.com/andylokandy/rfcs/blob/gc/text/2021-04-23-builtin-gc-worker.md)


tisonkun:
I've fixed ci for you at https://github.com/andylokandy/rfcs/pull/1 . Please take a look,

text/2021-04-23-builtin-gc-worker.md
```diff
@@ -0,0 +1,126 @@
...
+
+Three new concepts will be introduced by this proposal:
+
+1. `GC Barrier`
+
+    PD guarantees that the GC process will not affect the data with a timestamp later than any `GC Barrier`. TiKV users (e.g. TiDB / CDC / Client) can put GC Barriers into PD so as to make sure the data they are using will not be GC.
+
+2. `Transaction Safepoint`
+
+    TiKV and TiFlash guarantee that no read or write will happen on the data earlier than `Transaction Safepoint`. PD calculates `Transaction Safepoint Proposal` by `min(GC Barriers) - 1` and proposes it to TiKV and TiFlash via the `StoreHeartbeatResponse`. Therefore, TiKV and TiFlash will reject all requests with timestamp earlier than the `Transaction Safepoint`. In the next following `StoreHeartbeatRequest` TiKV and TiFlash will report their latest `Transaction Safepoint Status` to PD.
```
leiysky, line 34: Why do we need this concept? IMO, both TiFlash and TiKV shouldn't care about validity of timestamp.

andylokandy: This is for solving a very tricky corner case:

zhongzc: > `A region on StoreB having MVCC data only later than ts=100, anyhow transferred to StoreA via raft snapshot.`

This decision is made by PD. It has the choice that not to do this in the situation where Involved tikvs have different safepoints.

leiysky: After all, "it's PD's fault". 🤣

zhongzc: But I do agree stores should reject requests with stale ts. There should be something wrong to encounter these requests sent by some client. I think it's better to report this danger to the client.

leiysky: > But I do agree stores should reject requests with stale ts. There should be something wrong to encounter these requests sent by some client. I think it's better to report this danger to the client.

Rejection of request with stale ts can prevent extra resource consumption either.

text/2021-04-23-builtin-gc-worker.md
```diff
@@ -0,0 +1,131 @@
...
+
+Yet another tricky part is that TiDB has a shortcut to GC a range of keys after `DROP TABLE` or `DROP INDEX`: parallel with step 3 and step 4, TiDB will directly invoke `UnsafeDeleteRange` to TiKV which will delete the range of key-values regardless of locks or timestamp. To make it safe, TiDB will guarantee that the deleted range will never be used again.
+
+## Detailed design
+
+Three new concepts will be introduced by this proposal:
+
+1. `GC Barrier`
+
+    PD guarantees that the GC process will not affect the data with a timestamp later than any `GC Barrier`. TiKV users (e.g. TiDB / CDC / Client) can put GC Barriers into PD so as to make sure the data they are using will not be GC.
```
MyonKeminta, line 30: > PD guarantees that the GC process will not affect the data with a timestamp later than any `GC Barrier`.

Strictly speaking, *all snapshots with a TS after GC Barrier should be consistent*. Which means, not only the versions with a timestamp later than GC Barrier should be kept, but also the latest version before GC Barrier.

This description is also used when we are talking about safepoints.

text/2021-04-23-builtin-gc-worker.md
```diff
@@ -0,0 +1,131 @@
...
+
+2. `Transaction Safepoint`
+
+    TiKV and TiFlash guarantee that no read or write will happen on the data earlier than `Transaction Safepoint`. PD calculates `Transaction Safepoint Proposal` by `min(GC Barriers) - 1` and proposes it to TiKV and TiFlash via the `StoreHeartbeatResponse`. Therefore, TiKV and TiFlash will reject all requests with timestamp earlier than the `Transaction Safepoint`. In the next following `StoreHeartbeatRequest` TiKV and TiFlash will report their latest `Transaction Safepoint Status` to PD.
```
MyonKeminta, line 34: > TiKV and TiFlash guarantee that no read or write will happen on the data earlier than `Transaction Safepoint`.

When we are reading on a ts greater than the txn safepoint, a version before the safepoint should be read, if it's the latest version before the reading ts.
Consider:

TiKV and TiFlash guarantee that no read or write will happen on **a snapshot** earlier than `Transaction Safepoint`.

MyonKeminta, line 34: > PD calculates `Transaction Safepoint Proposal` by `min(GC Barriers) - 1`

min(GC Barrier - 1 for all GC Barriers, now - gc_lifetime)

text/2021-04-23-builtin-gc-worker.md
```diff
@@ -0,0 +1,131 @@
...
+
+3. `GC Safepoint`
+
+    PD calculates `GC Safepoint` by `min(Transaction Safepoint Status) - 1`, resolves locks earlier than `GC Safepoint`, then synchronizes it to TiKV and TiFlash via `StoreHeartbeatResponse`. And finally, TiKV and TiFlash are safe to delete MVCC data earlier than `GC Safepoint`.
```
MyonKeminta, line 38: What's the meaning of `Transaction Safepoint Status`?

text/2021-04-23-builtin-gc-worker.md
MyonKeminta: Maybe you'd better also mention that Transaction Saefpoint is guaranteed non-decreasing.

text/2021-04-23-builtin-gc-worker.md
```diff
@@ -0,0 +1,131 @@
...
++    uint64 txn_safepoint = 3;
++    uint64 gc_safepoint = 4;
++ }
+```
+
+### Backward compatibility
+
+#### Rolling update
+
+During the rolling update, the GC worker will not function and the new TiDB will not try to be elected as a GC leader. Once the rolling update is completed, the GC worker in PD will start to work.
```
MyonKeminta, line 117: How do you check if the rolling update is complete?

text/2021-04-23-builtin-gc-worker.md
```diff
@@ -0,0 +1,131 @@
...
+
+The RPC `GetGCSafePoint` and `UpdateServiceGCSafePoint`, which are used by the GC model before this proposal should still work for keep compatible with tools and they should be just a thin wrapper on top of the new `GetGCStatus` and `SetGCBarrier`.
```
MyonKeminta, line 119: You also need to ensure that service safepoints that are already set in old version won't be lost and will become the GC barrier after upgrading.

text/2021-04-23-builtin-gc-worker.md
andylokandy: I know it, maybe `snapshot ts` could make it clear.

text/2021-04-23-builtin-gc-worker.md
andylokandy: It is the txn safepoint TiKV has confirmed and applied.

text/2021-04-23-builtin-gc-worker.md
andylokandy: I think it's similar to async commit which has to be turned on manually. One problem is that  async commit switch is written in TiDB global variable which is not viable w/o TiDB. Maybe we have to add a pdpb RPC for that or to use etcd.

text/2021-04-23-builtin-gc-worker.md
andylokandy: Yes, that should not be difficult, but not possible to downgrade in place. I've to solution for that: 1. reuse the underling service safepoint storage for GC barrier 2. always store both service safepoint and GC barrier.

text/2021-04-23-builtin-gc-worker.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Motivation
+
+GC worker is an important component for TiKV that deletes outdated MVCC data so as to not explode the storage. But currently, the GC worker is implemented in TiDB, which makes TiKV not usable without TiDB.
+
+## Background
+
+According to the [documentation](https://docs.pingcap.com/tidb/stable/garbage-collection-overview) for the **current** GC worker, the GC worker processes the following steps in a GC cycle:
+
+1. Calculate the minimal timestamp among all living SQL sessions, earlier than which are safe to GC, called `Service Safepoint`.
```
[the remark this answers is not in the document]
youjiali1995: If the GC worker is in PD, TiDB is a service.

text/2021-04-23-builtin-gc-worker.md
andylokandy: updated

andylokandy: updated

text/2021-04-23-builtin-gc-worker.md
andylokandy: updated

text/2021-04-23-builtin-gc-worker.md
```diff
@@ -0,0 +1,131 @@
...
+2. [Upload](https://github.com/pingcap/kvproto/blob/8ecb5e46d7f5f7952a1a8d262b54f61dc8de1ef3/proto/pdpb.proto#L73) the `Service Safepoint` to PD, and get the minimal `Service Safepoint` among all Tools (e.g. CDC) from the response.
+3. Resolve all locks whose timestamp is earlier than the `Minimal Service Safepoint`.
+4. [Push](https://github.com/pingcap/kvproto/blob/8ecb5e46d7f5f7952a1a8d262b54f61dc8de1ef3/proto/pdpb.proto#L71) the `Minimal Service Safepoint` as `GCSafepoint` to PD.
+
+After the `GCSafepoint` being uploaded, TiKV will pull it from PD by interval, then TiKV will automatically delete all records earlier than `GCSafepoint`.
+
+Yet another tricky part is that TiDB has a shortcut to GC a range of keys after `DROP TABLE` or `DROP INDEX`: parallel with step 3 and step 4, TiDB will directly invoke `UnsafeDeleteRange` to TiKV which will delete the range of key-values regardless of locks or timestamp. To make it safe, TiDB will guarantee that the deleted range will never be used again.
```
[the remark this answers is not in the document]
andylokandy: updated

text/2021-04-23-builtin-gc-worker.md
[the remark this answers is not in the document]
andylokandy: updated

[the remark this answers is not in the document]
andylokandy: @youjiali1995 this section is talking about the current implementation.

text/2021-04-23-builtin-gc-worker.md
MyonKeminta: I'm afraid it would be over-complicated, and currently we didn't support automatically  enabling async commit on rolling  update due to the difficulty. Can you try to make them both able to work without breaking any consistency?

text/2021-04-23-builtin-gc-worker.md
MyonKeminta: I think reusing the service safepoint storage would be easier. If you want to write both the service safepoint keys and your new keys at the same time, you need to consider how to keep them consistent when PD leader is transferring, or when PD is rolling upgrading (which means there are old version PD nodes).

text/2021-04-23-builtin-gc-worker.md
```diff
@@ -12,26 +12,29 @@ GC worker is an important component for TiKV that deletes outdated MVCC data so
 
 According to the [documentation](https://docs.pingcap.com/tidb/stable/garbage-collection-overview) for the **current** GC worker, the GC worker processes the following steps in a GC cycle:
 
-1. Calculate the minimal timestamp among all living SQL sessions, earlier than which are safe to GC, called `Service Safepoint`.
-2. [Upload](https://github.com/pingcap/kvproto/blob/8ecb5e46d7f5f7952a1a8d262b54f61dc8de1ef3/proto/pdpb.proto#L73) the `Service Safepoint` to PD, and get the minimal `Service Safepoint` among all Tools (e.g. CDC) from the response.
-3. Resolve all locks whose timestamp is earlier than the `Minimal Service Safepoint`.
-4. [Push](https://github.com/pingcap/kvproto/blob/8ecb5e46d7f5f7952a1a8d262b54f61dc8de1ef3/proto/pdpb.proto#L71) the `Minimal Service Safepoint` as `GCSafepoint` to PD.
+1. Calculate the minimal timestamp among all living SQL sessions, earlier than which are safe to GC, called `Session Safepoint`.
+2. [Fetch](https://github.com/pingcap/kvproto/blob/8ecb5e46d7f5f7952a1a8d262b54f61dc8de1ef3/proto/pdpb.proto#L73) the `Service Safepoint` from PD, which is the minimal snapshot timestamp requirement among all tools (e.g. CDC, BR).
```
MyonKeminta, line 16: Actually it's not only *fetching*. TiDB it self also registers a service safepoing with id = `"gc_worker"`.

text/2021-04-23-builtin-gc-worker.md
andylokandy: Yeah, it is difficult, that's why I want to turned it on manually after rolling update. Once pd has turned builtin gc worker on, it will then reject UpdateGCSafepoint request.

MyonKeminta:
By the way, did you considered TiDB's old syncing mechanism for safepoints by reading and writing the key `/tidb/store/gcworker/saved_safe_point` in etcd? It's an approach to preventing transaction reading and writing before the safepoint. I think you need to consider its compatibility, and maybe you need to keep it compatible for the first major version.

text/2021-04-23-builtin-gc-worker.md
MyonKeminta: Emmm... personally I think it would not be a good idea if the user need to update the configuration manually, since GC should be as transparent as possible.

text/2021-04-23-builtin-gc-worker.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Background
+
+According to the [documentation](https://docs.pingcap.com/tidb/stable/garbage-collection-overview) for the **current** GC worker, the GC worker processes the following steps in a GC cycle:
+
+1. Calculate the minimal timestamp among all living SQL sessions, earlier than which are safe to GC, called `Session Safepoint`.
+2. [Fetch](https://github.com/pingcap/kvproto/blob/8ecb5e46d7f5f7952a1a8d262b54f61dc8de1ef3/proto/pdpb.proto#L73) the `Service Safepoint` from PD, which is the minimal snapshot timestamp requirement among all tools (e.g. CDC, BR).
+3. Calculate `GC Safepoint = min(Service Safepoint, Session Safepoint, now - gc_life_time)`.
+3. Resolve all locks whose timestamp is earlier than the `GC Safepoint`.
+4. [Push](https://github.com/pingcap/kvproto/blob/8ecb5e46d7f5f7952a1a8d262b54f61dc8de1ef3/proto/pdpb.proto#L71) the `GC Safepoint` to PD.
```
tisonkun, line 19:
```suggestion
4. Resolve all locks whose timestamp is earlier than the `GC Safepoint`.
5. [Push](https://github.com/pingcap/kvproto/blob/8ecb5e46d7f5f7952a1a8d262b54f61dc8de1ef3/proto/pdpb.proto#L71) the `GC Safepoint` to PD.
```

MyonKeminta:
@youjiali1995 @sticnarf PTAL at this proposal when you have time. Thanks!

andylokandy:
@MyonKeminta 

> By the way, did you considered TiDB's old syncing mechanism for safepoints by reading and writing the key /tidb/store/gcworker/saved_safe_point in etcd? It's an approach to preventing transaction reading and writing before the safepoint. I think you need to consider its compatibility, and maybe you need to keep it compatible for the first major version.

Do you mean that PD should set `/tidb/store/gcworker/saved_safe_point` even the builtin GC worker is running? For the concern of downgrading, I think it's acceptable.

text/2021-04-23-builtin-gc-worker.md
andylokandy: edited

text/2021-04-23-builtin-gc-worker.md
```diff
@@ -0,0 +1,169 @@
...
+
+After the `GCSafepoint` being uploaded, TiKV will pull it from PD by interval, then TiKV will automatically delete all records earlier than `GCSafepoint`.
+
+Yet another tricky part is that TiDB has a shortcut to GC a range of keys after `DROP TABLE` or `DROP INDEX`, etc: parallel with step 3 and step 4, TiDB will directly invoke `UnsafeDeleteRange` to TiKV which will delete the range of key-values regardless of locks or timestamp. To make it safe, TiDB will guarantee that the deleted range will never be used again.
+
+## Detailed design
+
+Four new concepts will be introduced by this proposal:
+
+1. `Safeline`
```
MyonKeminta, line 33: Personally I prefer still naming it *safepoint* since it's still a point in the timestamp axis, and it's a term that has been used for years by us, the uses and DBAs.

text/2021-04-23-builtin-gc-worker.md
```diff
@@ -0,0 +1,169 @@
...
+
+    Similar to `Safepoint`, `Safeline` tells after which timestamp is safe to read and before which is safe to GC, but it will no longer be globally equal among the whole key range but instead separated by key ranges so as to differently push safepoint forward for different key ranges. In this case, safepoint will be a special cased `Safepoint` who ranges across the whole key space.
+
+2. `GC Barrier`
+
+    PD guarantees that using a timestamp later than the `GC Barrier` to read and write on the range within the key range of `GC Barrier` will always be valid. TiKV users (e.g. TiDB / CDC / Client) can put GC Barriers into PD so as to make sure the data they are using will not be GC.
+
+    ![GC Barrier](../media/gc-barrier.png)
```
MyonKeminta, line 41: To ensure consistency, in the example in this pic, ResolveLocks should be done on all range above the `safeline`, even in the green area, otherwise it's possible that some unresolved locks' primaries' state is lost after GC. Therefore it would be hard to allow writes in the green area above the `safeline`. So I suggest r/w can be done below the `safeline`, and the green area above `safeline` is readonly.

text/2021-04-23-builtin-gc-worker.md
andylokandy: That's similar to the async commit switch which is a must-do. In detail, it can be a specified key in etcd.

text/2021-04-23-builtin-gc-worker.md
```diff
@@ -0,0 +1,149 @@
+# Builtin GC worker
+
+## Summary
+
+Move TiKV MVCC GC worker from TiDB into PD and make it general for use without TiDB.
+
+## Motivation
+
+GC worker is an important component for TiKV that deletes outdated MVCC data so as to not explode the storage. But currently, the GC worker is implemented in TiDB, which makes TiKV not usable without TiDB.
```
BusyJay, line 9: This is not true. Gc can be triggered within client.

BusyJay:
Why not put it into client? There is also a tikv proxy design, why not put it into the proxy? PD should be a lightweight component and not care about MVCC stuff.

text/2021-04-23-builtin-gc-worker.md
BusyJay: The mechanism described here adds extra complexity while not solve the problem completely. For example, an ongoing read right before transaction safepoint can be corrupted by unsafe delete range later. The behavior of reading data before safepoint is undefined. Client should make sure it's operating on the safe data, for example, making sure read timestamp is larger than the GC safepoint, by not advancing the GC Barrier.

BusyJay:
> @BusyJay I'd like to answer two of your comments together. The current implementation is unsound due to the corner case described in the comment above([#65 (comment)](https://github.com/tikv/rfcs/pull/65#discussion_r619682837)). To fix that, the concept of transaction safepoint has to be introduced, and, in fact, TiDB has implemented an implicit transaction safepoint by directly editing the etcd. Anyway, this solution requires a centralized component to synchronize a safepoint to all TiKV and be aware of TiKV responding that the safepoint is correctly received. So PD becomes the best one for the job.

I still don't see why GC should be moved to PD. A centralized component can be a just storage that provides atomicity access. It can be PD or TiKV. Resolving locks is not something PD should care.

andylokandy:
@BusyJay no, it's not only storage. The gc worker has to somehow know the transaction safepoint for each tikv

text/2021-04-23-builtin-gc-worker.md
```diff
@@ -0,0 +1,149 @@
...
+5. [Push](https://github.com/pingcap/kvproto/blob/8ecb5e46d7f5f7952a1a8d262b54f61dc8de1ef3/proto/pdpb.proto#L71) the `GC Safepoint` to PD.
+
+After the `GCSafepoint` being uploaded, TiKV will pull it from PD by interval, then TiKV will automatically delete all records earlier than `GCSafepoint`.
+
+Yet another tricky part is that TiDB has a shortcut to GC a range of keys after `DROP TABLE` or `DROP INDEX`, etc: parallel with step 3 and step 4, TiDB will directly invoke `UnsafeDeleteRange` to TiKV which will delete the range of key-values regardless of locks or timestamp. To make it safe, TiDB will guarantee that the deleted range will never be used again.
+
+## Detailed design
+
+Four new concepts will be introduced by this proposal:
+
```
tisonkun, line 28:
```suggestion

```

BusyJay:
> @BusyJay no, it's not only storage. The gc worker has to somehow know the transaction safepoint applied for each tikv, the best way is tikv reporting it in store heartbeat. If the gc worker is not in pd, pd has to expose 3 or 4 new RPC just for supporting the gc system. It would make no sense to me.

As what I have commented at https://github.com/tikv/rfcs/pull/65#discussion_r627665466, transaction safepoint is over-design, let client maintain its owned gc barrier is enough.

> Besides, what is the downside for pd to resolve lock? 

From design perspective, PD is meant to be routing and scheduling. Anything about transactions is out of its scope. From performance perspective, PD as the single node that maintains metadata and allocating TSO, it should be as lightweight as possible. Any unnecessary work can slow it down and hurt the cluster performance. From implementation perspective, things about MVCC and transactions are already implemented and maintained by TiDB, introduce them to PD just adds unnecessary complexity and extra maintainence.

> Is it worse than the stateless TiDB electing a GC leader to do the gc?

TiDB already has a similar leader like DDL owner and similar. I don't see what can go worse to add a GC leader. If TiDB is not used, then we can introduce TiKV proxy, which is another requirement that not just for GC, acting like TiDB to perform the GC operations.

andylokandy:
@BusyJay 

> As what I have commented at #65 (comment), transaction safepoint is over-design, let client maintain its owned gc barrier is enough.

We should guarantee that no **correctness** issue can happen by any API misusing. For example, the mixed raw and txn API usage has brought us many oncalls.

> From design perspective, PD is meant to be routing and scheduling. Anything about transactions is out of its scope. From performance perspective, PD as the single node that maintains metadata and allocating TSO, it should be as lightweight as possible. Any unnecessary work can slow it down and hurt the cluster performance. From implementation perspective, things about MVCC and transactions are already implemented and maintained by TiDB, introduce them to PD just adds unnecessary complexity and extra maintenance.

> TiDB already has a similar leader like DDL owner and similar. I don't see what can go worse to add a GC leader. If TiDB is not used, then we can introduce TiKV proxy, which is another requirement that not just for GC, acting like TiDB to perform the GC operations.

GC worker is similar to the deadlock detector, which should be done by TiKV itself at the very beginning. Let GC worker being in TiDB is totally a mistake. For sure, there is room for arguing whether the GC worker should be placed into TiKV just like the deadlock detector or being a standalone process or embedded into PD. Besides, I don't mean to directly embed it into PD but to link it into the same binary just like the TiDB dashboard.

tisonkun:
cc @Xuanwo you may be interested in this RFC.

text/2021-04-23-builtin-gc-worker.md
```diff
@@ -0,0 +1,149 @@
+# Builtin GC worker
```
tisonkun, line 1: You can update the file title to `0065-builtin-gc-worker.md`.

PR #66, opened 2021-04-27
prefer pull request id as RFC number

### LINKED ISSUE #22: Should we give each RFC an numeric identifier?

tisonkun:
Signed-off-by: tison <wander4096@gmail.com>

this closes #22 and supersedes #24 

cc @nrc @MyonKeminta


README.md
```diff
@@ -12,7 +12,7 @@ stakeholders can be confident about the direction the project is evolving in.
 
 ## How to submit an RFC
 
-1. Copy `template.md` into `text/YYYY-MM-DD-my-feature.md`.
+1. Copy `template.md` into `text/PRID-my-feature.md`.
```
MyonKeminta, line 15: One doesn't know the PRID before submitting the PR. You can refer to the way stated in my old PR.
By the way, I suggest include the propose time in each RFC.

MyonKeminta (commented):
Rest LGTM. Maybe I'm not the person who is qualified to make the decision though.

README.md
tisonkun: I think create a pull request and correspondingly change filename is trivial and thus we don't have to mention it clearly which looks redundant.

One may include the propose time in the RFC but I don't think it is a strong requirement. The time can be traced by the original pull request and generally we don't require an exact time but about how long ago, e.g., recent? months ago? years ago?

BusyJay (approved):

zhangjinpeng1987 (approved):
LGTM

PR #69, opened 2021-07-27
Add API V2 RFC

andylokandy:

## Motivation

`API V2` is a set of breaking changes that aim to solve serval issues with current RawKV (hereafter referred to as `API V1`):

1. `API V1` is not safe to use RawKV along with TxnKV.
2. `API V1` Raw TTL is controlled by Store configuration. Switching the configuration will cause data corruption in silence.
3. `API V1` Raw TTL is encoded into the value by appending 8-bytes UNIX timestamp to the end of the value, therefore it's hard to introduce other encode afterward.

By solving issue 1, TiDB will be able to support RawKV as the table's storage engine, which will enrich TiDB's use cases.

[Rendered](https://github.com/andylokandy/rfcs/blob/apiv2/text/0069-api-v2.md)


text/0069-api-v2.md
```diff
@@ -0,0 +1,106 @@
...
+Using TxnKV along with RawKV is safe in `API V2` mode.
+
+## Reference-level design
+
+### RawKV Key Encode
+
+Once the `API V2` is enabled, the key will be either start with either:
+
+1. `m` or `t`: TxnKV key. Used by TiDB.
+2. `x\0`: TxnKV key. The `\0` is a placeholder for the potential [Keyspace](https://github.com/tikv/rfcs/pull/39).
```
BusyJay, line 55: What able the compatibility with other systems like titan or juicefs?

andylokandy: They should choose `API V1` in client mode. If they want to support mixed TxnKV and RawKV, they should upgrade to `API V2` client.

BusyJay: I'm afraid they can't do that easily as data need to be reloaded.

andylokandy: No, they can't, but also i think they really need api v2.

BusyJay: Unless we always maintain two versions, which doesn't seem reasonable to me, otherwise they have to use V2 someday.

andylokandy: Can we drop the support for v1 api? If not,  I think maintainace effort is not the avoidable.

text/0069-api-v2.md
```diff
@@ -0,0 +1,106 @@
...
+
+- Run TTL compaction filter only on the keys starting with `r`.
+
+- Use the `API V2` Value encode in `RawStore`, `TTLStore` and `sst_importer`.
+
+If `stoarge.enable_api_v2` is not set:
+
+- Reject all requests with `enable_api_v2` in the context.
+
+### TiKV Clients
```
disksing, line 93: TiKV client needs to consider if it needs to decode keys when it loads a region from PD.

andylokandy: I know there is a bug the client may try to decode the start key and end key of a region, which will panic when the boundary is not decodable. But this is not related to the api v2 encode, so I didn't list it here.

disksing: It's not just a bug. It is a quite more complicated problem. Currently, based on whether TiKV is in txnkv or rawkv mode, the region split keys can be encoded or not encoded. When TiDB loads a region from PD, it should determine if it needs to decode region key based on it uses txnkv or rawkv.

text/0069-api-v2.md
```diff
@@ -0,0 +1,106 @@
...
+
+Provide two modes for users:
+
+- V2: Prepend `t\0` on TxnKV keys and prepend `r\0` on RawKV keys. Set `enable_api_v2` to `true` on TiKV gRPC's `Context`.
```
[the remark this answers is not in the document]
Connor1996: agree

text/0069-api-v2.md
```diff
@@ -0,0 +1,106 @@
...
+3. `r\0`: RawKV key.
+
+Note that in TxnKV, the key will be encoded by `Memory Comparable Encoding`. But since the `Memory Comparable Encoding` will not change the starting bytes but only add paddings, there won't be any overlap between RawKV and TxnKV.
+
+### RawKV Value Encode
+
+If the key is started with `r`, then the value can be either:
+
+1. `{0x0}{data}`
+2. `{0x1}{TTL expire timestamp}{data}`
```
Connor1996, line 65: why TTL is not appended after data

text/0069-api-v2.md
```diff
@@ -0,0 +1,106 @@
...
+
+### PD
+
+Add a new configuration `enable_api_v2` which can be configured by `pd-ctl`. When switching on or off, PD will synchronize the configuration to all TiKV. PD should scan keys to check that no non-TiDB data exists while switching `enable_api_v2`.
+
+### TiKV
+
+Add a new configuatrion `storage.enable_api_v2`. When enabled, `storage.enable_ttl` must also be enabled. TiKV will carry it's `storage.enable_api_v2` in `PutStore` request. PD will reject the `PutStore` if the switch mismatched.
```
Connor1996, line 73: Can we just remove `storage.enable_ttl`?

text/0069-api-v2.md
andylokandy: There is no difference. So I'd like to put metas together.

text/0069-api-v2.md
andylokandy: @zz-jason

text/0069-api-v2.md
andylokandy: I think the fix should change the region cache to not assume the split key is decodable.

text/0069-api-v2.md
andylokandy: I think it's not appropriate to remove it but we can make it enable by default.

text/0069-api-v2.md
skyzh: If I understand correctly, this encoding will make keys of continuous TTLs more likely to be stored in one region, instead of keys of the same user-key. Will this hurt scan performance?

skyzh (commented):
By the way, do we need to consider a new garbage collection logic apart from TxnKV for this design? e.g. how to handle TTL, how to remove old CAS result

andylokandy:
@skyzh that is value encoding, so the later put with ttl will overwrite the former.

text/0069-api-v2.md
```diff
@@ -0,0 +1,111 @@
...
+
+TiKV clients will have a new configuration choosing `API V1` or `API V2` mode on connecting to TiKV.
+
+TiKV will reject the request if the client mode is mismatched to the TiKV configuration. There is one exception: With TiKV in `API V2` mode, you can use `TxnClient` in `API V1` mode to access the TiDB data (key starting with `m` or `t`) which is invisible to the client in `API V2` mode.
+
+Note that in `API V2` mode, you can't specify CF in `RawClient`.
+
+Using TxnKV along with RawKV is safe in `API V2` mode.
+
+## Reference-level design
```
zz-jason, line 49: could you add a section to illustrate the compatibility issues, consider these scenarios:

- (client v1, TiKV API v2, TxnKV)
- (client v1, TiKV API v2, RawKV)
- (client v2, TiKV API v1, TxnKV)
- (client v2, TiKV API v1, RawKV)

the following information needs to be provided:

- whether the scenario is supported
- if not, is there a way to work around, and how to workaround?

text/0069-api-v2.md
```diff
@@ -0,0 +1,111 @@
...
+1. `{0x0}{data}`
+2. `{0x1}{TTL expire timestamp}{data}`
+
+### PD
+
+Add a new configuration `api_version` which can be configured to `1` or `2` by `pd-ctl`. When switching `api_version`, PD will synchronize the configuration to all TiKV. PD should scan keys to check that no non-TiDB data exists while switching `api_version`.
+
+### TiKV
+
+Add a new configuatrion `storage.api_version`. When enabled, `storage.enable_ttl` must also be enabled. TiKV will carry it's `storage.api_version` in `PutStore` request and PD will reject the `PutStore` if the version mismatched.
```
zz-jason, line 75: could you explain in detail about these things:

1. what's the current protocol to communicate between TiKV client and server?
2. what's the newly added field in the protocol, which place to put this field?

text/0069-api-v2.md
```diff
@@ -0,0 +1,111 @@
...
+
+### Upgrade
+
+Upgrade from `API V1` to `API V2` is a simple process:
+
+1. Update TiKV, TiDB, and PD to the version that supports `API V2`.
+2. Ensure that no non-TiDB data is in TiKV. Detele if any. (guaranteed by user)
+3. Ensure that no non-TiDB client is online. (guaranteed by user)
+4. Use `pd-ctl` to enable `API V2`.
+5. Restart all TiKV. (guaranteed by user)
```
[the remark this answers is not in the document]
andylokandy: 'no non-tidb data should exist' means 'only tidb data can exist' 😃

text/0069-api-v2.md
```diff
@@ -0,0 +1,111 @@
...
+## Reference-level design
+
+### Key Encode
+
+Once the `API V2` is enabled, the key will be starting with either:
+
+1. `m` or `t`: TxnKV key. Used by TiDB.
+2. `x\0`: TxnKV key. The `\0` is a placeholder for the potential [Keyspace](https://github.com/tikv/rfcs/pull/39).
+3. `r\0`: RawKV key.
+4. `a\0`: RawKV atomic key, on which you can use `CompareAndSwap`.
```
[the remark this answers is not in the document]
andylokandy: Atomic becomes an new mode.

text/0069-api-v2.md
```diff
@@ -0,0 +1,111 @@
...
+### RawKV Value Encode
+
+If the key is started with `r` or `a`, then the value can be either:
+
+1. `{0x0}{data}`
+2. `{0x1}{TTL expire timestamp}{data}`
+
+### PD
+
+Add a new configuration `api_version` which can be configured to `1` or `2` by `pd-ctl`. When switching `api_version`, PD will synchronize the configuration to all TiKV. PD should scan keys to check that no non-TiDB data exists while switching `api_version`.
```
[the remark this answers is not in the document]
andylokandy: It's in etcd and in tikv config file.

text/0069-api-v2.md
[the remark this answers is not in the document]
iosmanthus: The upgrade for RawKV users may require them to reload the data?

text/0069-api-v2.md
```diff
@@ -0,0 +1,111 @@
...
+
+1. Update TiKV, TiDB, and PD to the version that supports `API V2`.
+2. Ensure that no non-TiDB data is in TiKV. Detele if any. (guaranteed by user)
+3. Ensure that no non-TiDB client is online. (guaranteed by user)
+4. Use `pd-ctl` to enable `API V2`.
+5. Restart all TiKV. (guaranteed by user)
+
+### Downgrade
+
+Downgrade from `API V2` to `API V1` is also simple:
```
iosmanthus, line 32: What if the data in the storage is in the new format? Which means I can't even use the `api v1` to read the downgrade cluster?

andylokandy: no, you can't do that. I'm going to add a enforce check during the upgrade and downgrade.

text/0069-api-v2.md
andylokandy: updated

text/0069-api-v2.md
andylokandy: updated

text/0069-api-v2.md
andylokandy: updated. nothing to change in pd now.

text/0069-api-v2.md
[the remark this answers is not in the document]
andylokandy: yes

text/0069-api-v2.md
```diff
@@ -0,0 +1,135 @@
...
+## Reference-level design
+
+### Key Encode
+
+Once the `API V2` is enabled, the key will be starting with either:
+
+1. `m` or `t`: TxnKV key. Used by TiDB.
+2. `x{keyspace prefix}`: TxnKV key.
+3. `r{keyspace prefix}`: RawKV key.
+4. `a{keyspace prefix}`: RawKV atomic key, on which you can use `CompareAndSwap`.
```
zz-jason, line 74: is it acceptable to create a keyspace that contains keys from TxnKV, RawKV, and RawKV with atomic keys?

text/0069-api-v2.md
```diff
@@ -0,0 +1,135 @@
...
+
+The `{keyspace prefix}` is the [keyspace](https://github.com/tikv/rfcs/pull/39) prefix for seperating keys of different keyspace. It should be an vary-length integer whose highest bit of every byte denotes whether the next byte is still part of the integer. The client will fetch the prefix from PD by the keyspace name specified by the user when initializing the client, so it means that the keyspace prefix is valid during the session, in other words, the change on keyspace on PD will not take effect on running seesions.
```
zz-jason, line 76: could you describe more about the keyspace data stored in PD, for example, what are the keys and values represented for? how many keyspaces can be supported by storing keyspace names in PD?

text/0069-api-v2.md
zz-jason: +1

text/0069-api-v2.md
```diff
@@ -0,0 +1,135 @@
+# TiKV API V2
+
+## Motivation
```
zz-jason, line 3: we need a compatibility section to list the potential compatibility issues and the solution to solve them. For compatibility, the following things need to be considered:

1. compatibility with other components, like BR, TiCDC
2. compatibility with other features, like placement rules and other scheduling constraints and policies.

BTW, it's better to add a testing section to describe how to test these API changes on correctness and performance.

text/0069-api-v2.md
andylokandy: Refering https://github.com/tikv/rfcs/pull/69#issuecomment-891424135:

> If I understand correctly, this encoding will make keys of continuous TTLs more likely to be stored in one region, instead of keys of the same user-key. Will this hurt scan performance?

that is value encoding, so the later put with ttl will overwrite the former.

text/0069-api-v2.md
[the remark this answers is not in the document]
andylokandy: No. Let me clarify it in case of any misunderstanding. In this RFC, TiKV will have three modes, respectively RawKV, Raw Atomic, TxnKV, where each mode has its own client, i.e. RawClient, RawAtomicClient, TxnClient. Each client can't access data in another mode.

text/0069-api-v2.md
```diff
@@ -0,0 +1,135 @@
...
+In addition, keys will be contained in keyspaces, where the keys in different keyspace are totally independant. If keyspace is not specified, the keyspace 'default' will be used.
+
+The `API V2` is enabled by a switch on PD. Since it changes the storage encode, it will be not compatible to switch between `API V1` and `API V2` while there are non-TiDB data in TiKV. TiDB data is specially treated in order to not be affected by the change.
+
+### Upgrade
+
+Upgrade from `API V1` to `API V2` is a simple process:
+
+1. Update TiKV, TiDB, and PD to the version that supports `API V2`.
+2. Ensure that no non-TiDB data is in TiKV. Detele if any.
```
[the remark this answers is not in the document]
andylokandy: The [`Online Bulk Load for RawKV`](https://github.com/tikv/rfcs/pull/72) should help.

text/0069-api-v2.md
```diff
@@ -0,0 +1,135 @@
...
+3. Ensure that no non-TiDB client is online.
```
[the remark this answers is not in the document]
andylokandy: No, it can't.

text/0069-api-v2.md
```diff
@@ -0,0 +1,135 @@
...
+
+`API V2` is a set of breaking changes that aim to solve serval issues with current RawKV (hereafter referred to as `API V1`):
+
+1. RawKV is not safe to use along with TxnKV. By solving this, TiDB will be able to support RawKV as the table's storage engine, which will enrich TiDB's use cases.
+2. RawKV TTL is controlled by Store configuration. Switching the configuration will cause data corruption in silence.
+3. RawKV TTL is encoded into the value by appending 8-bytes UNIX timestamp to the end of the value, therefore it's hard to introduce other encode afterward.
+4. RawKV compare_and_swap is easy to make mistake.
+5. It could be nice if we can deploy multiple applications on one TiKV cluster.
+
+## Document-level design
```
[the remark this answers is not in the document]
andylokandy: > Referring https://github.com/rust-lang/rfcs/blob/master/0000-template.md#guide-level-explanation:

# Guide-level explanation
[guide-level-explanation]: #guide-level-explanation

Explain the proposal as if it was already included in the language and you were teaching it to another Rust programmer. That generally means:

- Introducing new named concepts.
- Explaining the feature largely in terms of examples.
- Explaining how Rust programmers should *think* about the feature, and how it should impact the way they use Rust. It should explain the impact as concretely as possible.
- If applicable, provide sample error messages, deprecation warnings, or migration guidance.
- If applicable, describe the differences between teaching this to existing Rust programmers and new Rust programmers.

For implementation-oriented RFCs (e.g. for compiler internals), this section should focus on how compiler contributors should think about the change, and give examples of its concrete impact. For policy RFCs, this section should provide an example-driven introduction to the policy, and explain its impact in concrete terms.

# Reference-level explanation
[reference-level-explanation]: #reference-level-explanation

This is the technical portion of the RFC. Explain the design in sufficient detail that:

- Its interaction with other features is clear.
- It is reasonably clear how the feature would be implemented.
- Corner cases are dissected by example.

The section should return to the examples given in the previous section, and explain more fully how the detailed proposal makes those examples work.

text/0069-api-v2.md
andylokandy: Yes! The keyspace won't assume the mode you use, instead, you will have access to the full key range in every mode in a keyspace.

text/0069-api-v2.md
andylokandy: done

text/0069-api-v2.md
```diff
@@ -0,0 +1,135 @@
...
+4. Use `pd-ctl` to enable `API V2`.
+5. Restart all TiKV.
+
+### Downgrade
+
+Downgrade from `API V2` to `API V1` is also simple:
+
+1. Ensure that no non-TiDB TiKV client is online.
+2. Ensure that no non-TiDB data is in TiKV. Detele if any.
+3. Use `pd-ctl` to disable `API V2`.
```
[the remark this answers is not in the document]
andylokandy: added in the PD section

text/0069-api-v2.md
```diff
@@ -0,0 +1,178 @@
...
+### Key Encode
+
+Once the `API V2` is enabled, the key will be starting with either:
+
+1. `m` or `t`: TxnKV key. Used by TiDB.
+2. `x{keyspace prefix id}`: TxnKV key.
+3. `r{keyspace prefix id}`: RawKV key.
+4. `a{keyspace prefix id}`: RawKV atomic key, on which you can use `CompareAndSwap`.
+
+The `{keyspace prefix id}` is the [keyspace](https://github.com/tikv/rfcs/pull/39) prefix for seperating keys of different keyspace. It should be an vary-length integer whose highest bit of every byte denotes whether the next byte is still part of the integer. The client will fetch the prefix from PD by the keyspace name specified by the user when initializing the client, so it means that the keyspace prefix is valid during the session, in other words, the change on keyspace on PD will not take effect on running seesions.
```
zz-jason, line 84: how to utilize key space in the tidb side?

text/0069-api-v2.md
```diff
@@ -0,0 +1,178 @@
...
+### RawKV Value Encode
+
+If the key is started with `r` or `a`, then the value can be either:
+
+1. `{0x0}{data}`
+2. `{0x1}{TTL expire timestamp}{data}`
+
+### PD
+
+Add a new http interface for adding, renaming, deleteing and querying the mapping from keyspace name to prefix id:
```
zz-jason, line 97: The keyspace name comes from user application, which makes it too sensitive to be leaked. We need to control the access privilege of this information. Is it better to use gRPC or other methods?

text/0069-api-v2.md
```diff
@@ -0,0 +1,178 @@
...
+
+```javascript
+// list all keyspaces
+GET /keyspaces
+[
+  {
+    name: "default",
+    prefix_id: "0"
```
zz-jason, line 105: 1. Should `prefix_id` be an integer?
2. How about `name` and `id`?

text/0069-api-v2.md
```diff
@@ -0,0 +1,178 @@
...
+// add new keyspace
+POST /keyspaces
+{
+  name: "mongodb",
+  // optional. if is null, PD will allocate an unused prefix id
+  prefix_id: "2"
+}
+
+// update a keyspace
+POST /keyspaces/{keyspace_name}
```
zz-jason, line 127: I prefer to not allow users to update the keyspace id since it's internal information. So maybe this API is only used to update the keyspace name?

text/0069-api-v2.md
```diff
@@ -0,0 +1,178 @@
...
+// delete a keyspace
+DELETE /keyspaces/{keyspace_name}
+```
+
+The keyspace is stored in etcd and should have no limitation on the numbers.
+
+### TiKV
+
+In TiKV config file, add a new configuration `storage.api_version`. When enabled, `storage.enable_ttl` must also be enabled.
```
zz-jason, line 136: what would happen if we enable `storage.api_version` and leave `storage.enable_ttl` disabled?

text/0069-api-v2.md
```diff
@@ -0,0 +1,178 @@
...
+
+In kvdb, add a store meta `api_version`. When the store meta mismatches the config`storage.enable_ttl`, it means that the user is switching the API version, then check no non-TiDB exist, and then save the new api version in store meta.
```
zz-jason, line 138:
```suggestion
In kvdb, add a store meta `api_version`. When the store meta mismatches the config `storage.enable_ttl`, it means that the user is switching the API version, then check no non-TiDB exist, and then save the new api version in store meta.
```

text/0069-api-v2.md
```diff
@@ -0,0 +1,178 @@
...
+
+In kvproto message `SSTMeta`, add `api_version`. Reject the SST file if the version is mismatched.
+
+In TiKV gRPC's context, add a field `api_version`.
+
+If `storage.api_version=2`:
+
+- Run TTL compaction filter only on the keys starting with `r`.
```
zz-jason, line 146: do we need to consider keys with prefix `i` as well?

text/0069-api-v2.md
```diff
@@ -0,0 +1,178 @@
...
+
+- Use the `API V2` Value encode in `RawStore`, `TTLStore` and `sst_importer`.
+
+- If the request's context has `api_version=2`:
+  - Reject the request except for it's a TxnKV request and the keys starting with `m` or `t`.
```
zz-jason, line 151: why rejecting these requests since the `storage.api_vesion` and the request's context has `api_version=2`?

text/0069-api-v2.md
```diff
@@ -0,0 +1,178 @@
...
+3. Use `pd-ctl` to disable `API V2`.
+4. Disable `API V2` in TiKV config file and restart TiKV.
+
+### CDC / BR
+
+Since all access to TiDB is unchanged during the upgrade, CDC and BR should work trivally on TiDB data.
+
+### Data migration
+
+It's reasonable to provide a way to import and export non-TiDB data in TiKV during the upgrade or downgrade. On TiKV before 4.0, the only way to do that is `scan` and `batch_put` on the client. After 4.0, TiKV start to support importing SST file into TxnKV, and after 5.1, importing on RawKV is also supported. You can find more information in [`RFC: Online Bulk Load for RawKV`](https://github.com/tikv/rfcs/pull/72).
```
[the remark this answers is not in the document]
andylokandy: > How to distinguish and scan all the non-tidb data in TiKV? If there is non-tidb data inside a tikv cluster 4.0

The TxnKV data start with `m` or `t` are considered as tidb data, otherwise are non-tidb data.

>  If there is non-tidb data inside a tikv cluster 4.0, what's the correct step to update the cluster to the latest version without data losing?

The user should migrate the non-tidb to new version tikv either by the new `SST import and export` feature or `get()` and `put()` through the client.

text/0069-api-v2.md
```diff
@@ -0,0 +1,178 @@
...
+Downgrade from `API V2` to `API V1` is also simple:
+
+1. Ensure that no non-TiDB TiKV client is online.
+2. Ensure that no non-TiDB data is in TiKV. Detele if any.
+3. Use `pd-ctl` to disable `API V2`.
+4. Disable `API V2` in TiKV config file and restart TiKV.
+
+### CDC / BR
+
+Since all access to TiDB is unchanged during the upgrade, CDC and BR should work trivally on TiDB data.
```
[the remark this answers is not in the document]
andylokandy: It means nothing to change because the API of TiKV is not changed for TiDB data.

text/0069-api-v2.md
```diff
@@ -0,0 +1,178 @@
...
+
+### Data migration
+
+It's reasonable to provide a way to import and export non-TiDB data in TiKV during the upgrade or downgrade. On TiKV before 4.0, the only way to do that is `scan` and `batch_put` on the client. After 4.0, TiKV start to support importing SST file into TxnKV, and after 5.1, importing on RawKV is also supported. You can find more information in [`RFC: Online Bulk Load for RawKV`](https://github.com/tikv/rfcs/pull/72).
+
+### Client Version
+
+TiKV clients will have a new configuration choosing `API V1` or `API V2` mode on connecting to TiKV.
```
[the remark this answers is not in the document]
andylokandy: The user is required to explicitly specify the version to use.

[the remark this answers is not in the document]
andylokandy: The alternative is to fetch the API version from PD.

text/0069-api-v2.md
andylokandy: TiDB doesn't have a corresponding keyspace because it uses the API V1 reserving for TiDB.

text/0069-api-v2.md
andylokandy: There is no difference between gRPC and HTTP on the aspect of security. The thing you might want is [RBAC Support for PD](https://github.com/tikv/rfcs/pull/62).

text/0069-api-v2.md
andylokandy: 1. good catch
2. I think `id` is not so appropriate because different keyspaces may share the same 'prefix' (aka. id).

text/0069-api-v2.md
andylokandy: The tikv will panic on start.

text/0069-api-v2.md
andylokandy: What is the `i` prefix for?

text/0069-api-v2.md
andylokandy: Fixed. It was a typo, I was meant to say "If the request's context has `api_version=1`".

text/0069-api-v2.md
andylokandy: I removed "update" entirely.

text/0069-api-v2.md
andylokandy: done

text/0069-api-v2.md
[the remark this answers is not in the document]
zz-jason: please check whether the word is a typo :joy:. how about 'work smoothly or 'work as usual'?

text/0069-api-v2.md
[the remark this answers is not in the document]
zz-jason: if users want to upgrade a tikv cluster to a newer version. should they first scan all the existing non-tidb data from tikv, store them to external storage (maybe a file), then upgrade the cluster, and finally put these non-tidb data from the external storage to the upgraded tikv cluster?

text/0069-api-v2.md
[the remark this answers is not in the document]
zz-jason: it's better to let old clients be able to connect the new version tikv cluster. consider that during the upgrade, an old tidb-server may need to connect to the new tikv cluster to provide services. setting the config to `v1` may be a better choice.

text/0069-api-v2.md
zz-jason: it stands for index

zz-jason: never mind, tidb only uses `m` and `t` prefixes, `_i` and `_r` are used as sub prefixes together with `t`.

text/0069-api-v2.md
```diff
@@ -0,0 +1,171 @@
...
+    id: "1"
+  }
+]
+
+// add new keyspace
+POST /keyspaces
+{
+  name: "mongodb",
+  // optional, pd will allocate one if not specified
+  id: "2"
```
zz-jason, line 118: should it be auto-generated by tikv, which can avoid possible coniflict?

andylokandy: The name and id are both required to be unique. Allowing specifying id so that user can recover a delete keyspace.

text/0069-api-v2.md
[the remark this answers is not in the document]
andylokandy: This is one of the approches.

text/0069-api-v2.md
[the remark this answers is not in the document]
andylokandy: During the upgrade, both server and client should be using API V1. The API version switch should take place after the upgrade has finished.

text/0069-api-v2.md
zz-jason: > recover a delete keyspace

when will we remove the data after deleting a keyspace?

andylokandy: No one but the user can delete the data. So the data will be left there if the user delete the keyspace without clearing them.

text/0069-api-v2.md
```diff
@@ -0,0 +1,171 @@
...
+DELETE /keyspaces/{keyspace_name}
+```
+
+The keyspace is stored in etcd and should have no limitation on the numbers.
+
+### TiKV
+
+In TiKV config file, add a new configuration `storage.api_version`. When enabled, `storage.enable_ttl` must also be enabled.
+
+In kvdb, add a store meta `api_version`. When the store meta mismatches the config `storage.enable_ttl`, it means that the user is switching the API version, then check no non-TiDB exist, and then save the new api version in store meta.
```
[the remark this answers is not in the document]
andylokandy: typo: non-TiDB data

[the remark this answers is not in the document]
andylokandy: Then tikv should panic on start. The user should turn the config switch back and delete the data.

gregwebs (commented):
Can you explicitly discuss changing the keyspace prefix during import? Data restore from backup or migration (for example merging data from one TiDB cluster into another).

text/0069-api-v2.md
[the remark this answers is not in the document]
zz-jason: you can put other reasonable approaches to this section as well.

text/0069-api-v2.md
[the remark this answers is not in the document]
zz-jason: > During the upgrade, both server and client should be using API V1. The API version switch should take place after the upgrade has finished.

@andylokandy I thought this is done during upgrading the cluster. Could you write this information explicitly in the **"Upgrade"** section?

> The server (set by the administrator) could specify whether the version must be set by the client or whether the server will fill in a default value.

agree with @gregwebs.

text/0069-api-v2.md
```diff
@@ -0,0 +1,171 @@
...
+
+Add a new http interface for adding, renaming, deleteing and querying the mapping from keyspace name to prefix id:
+
+```javascript
+// list all keyspaces
+GET /keyspaces
+[
+  {
+    name: "default",
+    id: "0"
```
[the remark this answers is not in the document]
zz-jason: +1, an optional field to describe the keyspace is useful. Lots of things can be added upon keyspace, for example:

- users and their access privileges
- placement policies to physically isolate logically different keyspaces

we can do it later since these functionalities don't break the underlying key-value codec, they only affect the keyspace metadata.

text/0069-api-v2.md
zz-jason: > No one but the user can delete the data. So the data will be left there if the user deletes the keyspace without clearing them.

@andylokandy after the keyspace is deleted, the key-values inside that key-space can not be able to be accessed anymore, Am I correct?

Could you elaborate on why not deleting the key-values inside that keyspace?

zz-jason (commented):
to improve the readability and expose the most concerning compatibility issues, how about reorganizing the content to:

```
## Motivation

keep the original content here.

## Detailed Design

### The new key-value codec after bringing the keyspace concept

introduce how the key is encoded and decoded for:
- TiDB data
- TxnKV Data
- RawKV data with CAS support
- RawKV data without CAS support

### Keyspace Management

- how to discover keyspaces
- how to add/add/rename a keyspace
- when a keyspace is deleted, whether and when to delete the data belonging to that keyspace

### How to safely enable and utilize keyspace

- how to enable keyspace after upgrading the cluster if there are mixed TiDB and non-TiDB data

## Compatibility issues and how to workaround

### Compatibility issues after upgrading without enabling keyspace

### Compatibility issues after upgrading and enabling keyspace

- compatibility issues related to TiDB migration and replication tools
- compatibility issues related to existing TiKV clients
- compatibility issues related to TiKV migration tools, like BR.
- compatibility issues 

## Test Plan

### What test do we have and can cover the changes introduced by this RFC

### What test do we need to add

```

text/0069-api-v2.md
```diff
@@ -0,0 +1,172 @@
...
+In addition, the user key in every mode will be encoded by `Memory Comparable Encoding` as TxnKV does in API V1.
+
+#### RawKV Value Encode
+
+If the key is started with `r` or `a`, then the value can be either:
+
+1. `{0x0}{data}`
+2. `{0x1}{TTL expire timestamp}{data}`
+
+### Keyspace Management
```
zz-jason, line 43: how about simplifying the HTTP methods to only use GET and PUT, and unifying the API to start with `/keyspace` like the following:

```shell
# list all keyspaces
GET "{pd_address:status}/keyspace/list"

# add keyspaces
curl -X POST -d "{json}" "{pd_address:status}/keyspace/add"

# delete keyspaces
curl -X POST -d "{json}" "{pd_address:status}/keyspace/delete"
```

And we also lack a method to update the name of an existing keyspace, how about adding the following HTTP API:

```shell
# update keyspaces
curl -X POST -d "{json}" "{pd_address:status}/keyspace/update"
```

after adding the update method, the whole CURD operations on keyspace are supported.

text/0069-api-v2.md
```diff
@@ -0,0 +1,172 @@
...
+// add new keyspace
+POST /keyspaces
+{
+  name: "mongodb",
+  // optional, pd will allocate one if not specified
+  id: "2"
+}
+
+// delete a keyspace
+DELETE /keyspaces/{keyspace_name}
```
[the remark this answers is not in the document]
andylokandy: No, the client no longer have access to the deleted keyspace.

andylokandy:
@zz-jason 

> we can leave this as an open question to solve in the future. at the initial version of keyspace, I prefer a strict privilege control that does not allow users to add keyspace whose id conflicts with an existing one. We can relax the control in the future if a specific requirement appears. What do you think?

If so, what to recover if you deleted a keyspace by accident?

text/0069-api-v2.md
```diff
@@ -0,0 +1,172 @@
...
+The `API V2` is enabled by a switch on PD. Since it changes the storage encode, it will be not compatible to switch between `API V1` and `API V2` while there are non-TiDB data in TiKV. TiDB data is specially treated in order to not be affected by the change.
+
+#### Key Encode
+
+Once the `API V2` is enabled, the key will be starting with either:
+
+1. `m` or `t`: TxnKV key. Used by TiDB.
+2. `x{keyspace prefix id}`: TxnKV key.
+3. `r{keyspace prefix id}`: RawKV key.
+4. `a{keyspace prefix id}`: RawKV atomic key, on which you can use `CompareAndSwap`.
```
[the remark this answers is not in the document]
andylokandy: > is it possible to add mvcc support for rawkv key-values, which can help us simplify the incremental data replication like TiCDC?

This is a big topic. Ignoring the bunch of reasons, I don't think mvcc without transaction control has any useful scenario.

> why not setting the {keyspace prefix id} as the common prefix for transactional and non-transactional key-values like the following:

Previously rawkv and txnkv use different encoding (MCE), and to avoid a region from spanning over two modes, PD should always split the keys between each modes. But now the rawkv uses MCE too in API V2, so there is no any reason to put mode prefix before the kepsapce id.

> achieve better performance when accessing both transactional and non-transaction key-values

This is not practically true. because in each request we can only access one region by one mode, we can not benefit from putting data in different modes into the same region.

>physically isolate the storage and computation on different key-spaces via placement rule

Reasonable.

zz-jason:
> Is it different from creating a new keyspace with a specified id?

Yes

andylokandy:
@zz-jason Does the flashback only recover the lastest deleted keyspace?

text/0069-api-v2.md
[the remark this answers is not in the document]
zz-jason: when the client is firstly connected to the cluster, it will record the mapping between keyspace name and id, how to notify the client that a keyspace is deleted and stop serving data access to that keyspace?

text/0069-api-v2.md
gregwebs: Not using REST makes things more complicated for me.

This kind of topic is a distraction for this RFC and any other with APIs. Instead we should have a consistent standard for how PD APIs are implemented (for now this would be determined by looking at existing APIs). If we want to change or formalize this standard, there should be a separate RFC.

text/0069-api-v2.md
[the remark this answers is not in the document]
andylokandy: The keyspace is valid throughout the session, which means we don't notify it.

text/0069-api-v2.md
zz-jason: yes. let's focused on the APIs for keyspace management at this RFC.

text/0069-api-v2.md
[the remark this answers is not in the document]
zz-jason: how about specially treating keys started with `m` and `t` as the first keyspace (keyspace 0), and using the following key-value encoding:

```
// for keyspace 0
{m|t}: prefix for keyspace 0, which is already used by TiDB

// for other keyspaces
k{keyspace id}x: prefix for transactional keyspaces
k{keyspace id}r: prefix for non-transactional keyspaces
k{keyspace id}a: prefix for non-transactional keyspaces with atmoic support
```

text/0069-api-v2.md
```diff
@@ -25,17 +25,31 @@ The `API V2` is enabled by a switch on PD. Since it changes the storage encode,
...
 
 The `{keyspace prefix id}` is the [keyspace](https://github.com/tikv/rfcs/pull/39) prefix for seperating keys of different keyspace. It should be an vary-length integer whose highest bit of every byte denotes whether the next byte is still part of the integer. The client will fetch the prefix from PD by the keyspace name specified by the user when initializing the client, so it means that the keyspace prefix is valid during the session, in other words, the change on keyspace on PD will not take effect on running seesions.
 
-In addition, the user key in every mode will be encoded by `Memory Comparable Encoding` as TxnKV does in API V1.
+Note that in TxnKV, the key will be encoded by `Memory Comparable Encoding`. But since the `Memory Comparable Encoding` will not change the starting bytes but only add paddings, there won't be any overlap between RawKV and TxnKV.
 
 #### RawKV Value Encode
 
-If the key is started with `r` or `a`, then the value can be either:
+If the key has RawKV prefix, then the value can be either:
```
zz-jason, line 38:
```suggestion
If the key has RawKV prefix, which are `r` and `a`, then the value can be either:
```

text/0069-api-v2.md
```diff
@@ -62,17 +76,25 @@ GET /keyspaces
...
 DELETE /keyspaces/{keyspace_name}
+{}
+
+// recover the deleted keyspace, if no name or id is provided, recover the lastest deleted keyspace
+POST /keyspaces?action=flashback
+{
+  // optional
+  name: "mongodb",
+  // optional
+  id: "mongodb",
```
zz-jason, line 91: should it be:

```suggestion
{
  // required
  name: "mongodb",
  // optional
  new_name: "mongodb",
```

text/0069-api-v2.md
```diff
@@ -0,0 +1,194 @@
...
+  {
+    name: "redis",
+    id: "1"
+  }
+]
+
+// add new keyspace
+POST /keyspaces
+{
+  name: "mongodb",
```
zz-jason, line 78: better to use a neutral name in the example.

text/0069-api-v2.md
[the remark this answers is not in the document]
zz-jason: without notification, how to ensure the data can not be accessed when the keyspace is deleted?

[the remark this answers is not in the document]
gregwebs: Notification would be a big change? For now clients can refresh their data? Would having a TTL for deletion help?

Even if a client is notified, what will stop them from using the deleted keyspace? We will have to follow this up with [auth](https://github.com/tikv/rfcs/pull/40) for keyspaces to enforce proper client behavior.

text/0069-api-v2.md
```diff
@@ -0,0 +1,194 @@
...
+2. `{0x1}{TTL expire timestamp}{data}`
+
+#### TTL in Titan
+
+Titan is a storage engine plugin for RocksDB that store the value of a kvpair in a separated blob file if the value is large enough. This brings a new challenge for RawKV TTL compaction filter where the compaction filter can not know the ttl without reading the blob file. So with Titan, the RawKV value encode is defined to be:
+
+##### Value
+
+1. `{0x0}`
+2. `{0x1}{TTL expire timestamp}`
```
sunxiaoguang, line 50: Is it possible to have other encoding schema later on? If not, we can probably use varint encoding to encode TTL lifetime and assume lifetime of 0 means no expiration. This way values are encoded  exactly the same way.

sunxiaoguang: It would be nice to have TTL support in TxnKV.

andylokandy: > Is it possible to have other encoding schema later on.

Yes, it's possible, just to use a new prefix byte in the value to indicate the new format.

andylokandy:
@zz-jason 
>we can leave this as an open question to solve in the future. at the initial version of keyspace, I prefer a strict privilege control that does not allow users to add keyspace whose id conflicts with an existing one. We can relax the control in the future if a specific requirement appears. What do you think?

Yeah, I agree.

text/0069-api-v2.md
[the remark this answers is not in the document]
andylokandy: foundationdb and Cassandra don't provide such a field for keyspace:
- https://apple.github.io/foundationdb/developer-guide.html#developer-guide-sub-keyspaces
- https://docs.datastax.com/en/cql-oss/3.3/cql/cql_reference/cqlCreateKeyspace.html

In addition, we should also provide an API for modifying the description, which adds more complexity.

text/0069-api-v2.md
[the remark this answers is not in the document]
andylokandy: updated. the client will refresh keyspace every 5min and return error if the keyspace is deleted.

text/0069-api-v2.md
```diff
@@ -0,0 +1,195 @@
...
+// recover the latest deleted keyspace
+POST /keyspaces?action=flashback
+{
+  new_name: "bar",
+}
+```
+
+The keyspaces are stored in etcd and has the no limitation on the id number.
+
+Note that when deleting a keyspace, the data in the keysapce will not be deleted automatically, the user should clean up the keyspace before deleting the kepspace. If a keyspace is deleted by accident, the user should be able to recover the keyspace by creating a keyspace using the same id as before. When a keyspace is deleted, the client should aware of the deletion in 5 mins, and after 5 mins, it guranetees that no client should be using the keyspace.
```
zz-jason, line 97:
```suggestion
1. Adding keyspace: newly added keyspaces can only be viable to new clients.
2. Deleting keyspace:
	1. Deleting keyspace only marks the metadata to inviable in PD, the data in the keyspace and metadata in PD will not be deleted automatically. Garbage collecting the data in deleted keyspaces may be introduced in the future since it increases the complexity of this RFC. To ensure no data is left, the user should clean up the keyspace before deleting the keyspace at present.
	2. Every client syncs the keyspace information with the PD leader every 5 minutes. When a keyspace is deleted, the client should be aware of the deletion in 5 minutes. The keyspace can not be accessed by any living client after 5 minutes.
3. Flashbacking keyspace:
	1. Flashbacking keyspace only affects the metadata, turning the keyspace name and id mapping to viable by clients.
	2. If there are multiple deleted keyspaces with the same name, only the last deleted keyspace is flashbacked. Users can flashback all these deleted keyspaces by calling the flashback API multiple times with different new keyspace names.
```

zz-jason: Please specifically describe the keyspace name and id mappings stored in PD. As far as I can see, it must contain the following field for each mapping:

- keyspace name: string
- keyspace id: int64
- created at: timestamp
- deleted at: timestamp, if set, the keyspace is not visiable to users
- flashbacked at: timstamp

text/0069-api-v2.md
```diff
@@ -0,0 +1,195 @@
...
+
+### How to safely enable API V2
+
+#### Upgrade
+
+Upgrade from `API V1` to `API V2` is a simple process:
+
+1. Update TiKV, TiDB, and PD to the version that supports `API V2`.
+2. Ensure that no non-TiDB data is in TiKV. Delete if any.
```
zz-jason, line 106:
```suggestion
2. Ensure that all the keys in TiKV are written by TiDB, which are prefixed with `m` or `t`. Delete if any. Users can perform 3 range scans on (-oo, m), (m, t), (t, +oo) to identify that.
```

Correct me if I'm wrong

text/0069-api-v2.md
```diff
@@ -0,0 +1,195 @@
...
+3. Ensure that no non-TiDB client is online. (manual process)
```
zz-jason, line 107:
```suggestion
3. Ensure that TiDB is the only client and user on TiKV. Needs to be manually verified.
```

text/0069-api-v2.md
```diff
@@ -0,0 +1,195 @@
...
+4. Use `pd-ctl` to enable `API V2`.
```
zz-jason, line 108: could you especially specify the command used by pd-ctl?

text/0069-api-v2.md
```diff
@@ -0,0 +1,195 @@
...
+5. Enable `API V2` in TiKV config file and restart TiKV.
```
zz-jason, line 109: this is a heavy step that breaks the existing session. could we enable API v2 without restarting the tikv-servers?

text/0069-api-v2.md
```diff
@@ -0,0 +1,195 @@
...
+
+#### Downgrade
+
+Downgrade from `API V2` to `API V1` is also simple:
+
+1. Ensure that no non-TiDB TiKV client is online.
```
zz-jason, line 115:
```suggestion
1. Ensure that TiDB is the only client and user on TiKV. Needs to be manually verified.
```

text/0069-api-v2.md
```diff
@@ -0,0 +1,195 @@
...
+2. Ensure that no non-TiDB data is in TiKV. Detele if any.
```
zz-jason, line 116:
```suggestion
2. Ensure that all the keys in TiKV are written by TiDB, which are prefixed with `m` or `t`. Delete if any. Users can perform 3 range scans on (-oo, m), (m, t), (t, +oo) to identify that.
```

text/0069-api-v2.md
```diff
@@ -0,0 +1,195 @@
...
+3. Use `pd-ctl` to disable `API V2`.
```
zz-jason, line 117: ditto

text/0069-api-v2.md
```diff
@@ -0,0 +1,195 @@
...
+
+#### Data migration
+
+It's reasonable to provide a way to import and export non-TiDB data in TiKV during the upgrade or downgrade. On TiKV before 4.0, the only way to do that is `scan` and `batch_put` on the client. After 4.0, TiKV start to support importing SST file into TxnKV, and after 5.1, importing on RawKV is also supported. You can find more information in [`RFC: Online Bulk Load for RawKV`](https://github.com/tikv/rfcs/pull/72).
+
+### Implementation Details
+
+#### PD
+
+Add the new APIs described [above](#Keyspace-Management).
```
zz-jason, line 128: what happens when we execute the command in pd-ctl to enable API v2?

text/0069-api-v2.md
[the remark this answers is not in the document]
gregwebs: Note that a FDB subspace is client-side only: there is no central registry. It is likely that some FDB users implement create a separate central registry (maybe just a document) that has a description.

I can see how without any real [auth system](https://github.com/tikv/rfcs/pull/40) in place there will always be a desire to mutate data because it always brings up questions relating to permission which are impossible to address.

gregwebs (approved):
I am really looking forward to this!

text/0069-api-v2.md
andylokandy: Then PD will open its keyspace APIs. but nothing will happen on TiKV unless we manually edit the tikv config and restart.

text/0069-api-v2.md
andylokandy: We have to do it

zz-jason: why?

text/0069-api-v2.md
```diff
@@ -0,0 +1,225 @@
...
+3. Flashbacking keyspace:
+  3.1. Flashbacking keyspace only affects the metadata, turning the keyspace name and id mapping to viable by clients.
+  3.2. If there are multiple deleted keyspaces with the same name, only the last deleted keyspace is flashbacked. Users can flashback all these deleted keyspaces by calling the flashback API multiple times with different new keyspace names.
+
+#### pd-ctl
+
+To enable API V2:
+
+```bash
+>> config set api-version 2
```
zz-jason, line 107: could you explain why we need to set this config in PD?

zz-jason (approved):
LGTM

text/0069-api-v2.md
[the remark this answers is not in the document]
sunxiaoguang: Other than CAS command, are RawKV and Raw Atomic sharing the same implementation?

text/0069-api-v2.md
sunxiaoguang: I also agree that current design is the most appropriate at time. Not reclaiming data when key space is deleted sounds a reasonable protection to user in case there is mistake during operation. If we really would like to improve this, we can add retain policy just like what K8S does for PVC and POD.

text/0069-api-v2.md
sunxiaoguang: Just have a second thought, how about we make this encoding works for V2 TxnKV?

sunxiaoguang:
> I am really looking forward to this!

+1

text/0069-api-v2.md
```diff
@@ -0,0 +1,225 @@
...
+  new_name: "bar",
+}
+```
+
+The keyspaces are stored in etcd and has the no limitation on the id number.
+
+1. Adding keyspace: newly added keyspaces can only be viable to new clients.
+2. Deleting keyspace:
+  2.1. Deleting keyspace only marks the metadata to inviable in PD, the data in the keyspace and metadata in PD will not be deleted automatically. Garbage collecting the data in deleted keyspaces may be introduced in the future since it increases the complexity of this RFC. To ensure no data is left, the user should clean up the keyspace before deleting the keyspace at present.
+  2.2. Every client syncs the keyspace information with the PD leader every 5 minutes. When a keyspace is deleted, the client should be aware of the deletion in 5 minutes. The keyspace can not be accessed by any living client after 5 minutes.
```
sunxiaoguang, line 97: How about make it configurable but deny extreme settings, 0 minute for example.

text/0069-api-v2.md
```diff
@@ -0,0 +1,225 @@
...
+#### Keyspace metadata
+
+```json
+{
+  name: string,
+  id: int64,
+  created_at: timestamp,
+  deleted_at: timestamp, // if set, the keyspace is not visiable to users
+  flashbacked_at: timstamp
+}
```
marsishandsome, line 128: Can we extend the metqdata？ e.g. add a new filed `default_TTL`, which means if users put a kv without ttl, the `defualt_TTL` will be used.

marsishandsome: LGTM

text/0069-api-v2.md
andylokandy: This is a big topic. The main problem is that it's hard to synchronize this setting across all components including the clients the servers and pd.

text/0069-api-v2.md
```diff
@@ -0,0 +1,242 @@
...
+The `API V2` is enabled by a switch on PD. Since it changes the storage encode, it will be not compatible to switch between `API V1` and `API V2` while there are non-TiDB data in TiKV. TiDB data is specially treated in order to not be affected by the change.
+
+#### Key Encode
+
+Once the `API V2` is enabled, the key will be starting with either:
+
+1. `m` or `t`: TxnKV key. Used by TiDB.
+2. `k{keyspace prefix id}x`: TxnKV key.
+3. `k{keyspace prefix id}r`: RawKV key.
+4. `k{keyspace prefix id}a`: RawKV atomic key, on which you can use `CompareAndSwap`.
```
BusyJay, line 30: Why need a standalone keyspace?

text/0069-api-v2.md
```diff
@@ -0,0 +1,242 @@
...
+
+The `{keyspace prefix id}` is the [keyspace](https://github.com/tikv/rfcs/pull/39) prefix for seperating keys of different keyspace. It should be an vary-length integer whose highest bit of every byte denotes whether the next byte is still part of the integer. The client will fetch the prefix from PD by the keyspace name specified by the user when initializing the client, so it means that the keyspace prefix is valid during the session, in other words, the change on keyspace on PD will not take effect on running seesions.
+
+Note that in TxnKV, the key will be encoded by `Memory Comparable Encoding`. But since the `Memory Comparable Encoding` will not change the starting bytes but only add paddings, there won't be any overlap between RawKV and TxnKV.
+
+#### RawKV Value Encode
+
+If the key has RawKV prefix, which are `k{keyspace id prefix}r` or `k{keyspace id prefix}a`, then the value can be either:
```
BusyJay, line 38: /cc @Connor1996

text/0069-api-v2.md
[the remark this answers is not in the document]
andylokandy: > Other than CAS command, are RawKV and Raw Atomic sharing the same implementation?

yes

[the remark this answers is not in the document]
sunxiaoguang: Cool

text/0069-api-v2.md
andylokandy: It's not a standalone keyspace. it's an individual mode to plain rawkv mode in a keyspace. This means the client would provide three kinds of clients to access the same keyspace (RawClient, RawAtomicClient, TxnClient), and data for each client is transparent to other clients.

text/0069-api-v2.md
zz-jason: we can hardcode a setting at present. optimize the mechanism to improve user experience in the future.

text/0069-api-v2.md
```diff
@@ -64,7 +64,18 @@ GET /keyspaces
 [
   {
     name: "default",
-    id: "0"
+    id: "0",
+    metadata: {
```
zz-jason, line 68: I'd like to rename it to "property" rather than "metadata". BTW, is there any guideline about what fields should be contained in this JSON object, what fields should be put outside this JSON object?

zz-jason: please also update the metadata design in PD after the guideline is settled.

andylokandy: > I'd like to rename it to "property" rather than "metadata".

Fair

> BTW, is there any guideline about what fields should be contained in this JSON object, what fields should be put outside this JSON object?

No. Every component can define a new config item. The component may print a warning if it meets an unrecognized property under its section (eg. TxnClient can warn the property `default-config.txn-client.abcd`).

zz-jason: is it possible to put "created_at" to this field?

text/0069-api-v2.md
BusyJay: Why need to separate the data physically? In theory, RawKV should be able to accessed atomically without changing the data.

text/0069-api-v2.md
andylokandy: If you're mentioning the `created_at` in top-level metadata, no, because it's only writable by PD. Otherwise, if you ask whether the user can set a custom `created_at` property, it's yes, the user is allowed to set whatever he wants.

text/0069-api-v2.md
Connor1996: +1

text/0069-api-v2.md
```diff
@@ -0,0 +1,242 @@
...
+Note that in TxnKV, the key will be encoded by `Memory Comparable Encoding`. But since the `Memory Comparable Encoding` will not change the starting bytes but only add paddings, there won't be any overlap between RawKV and TxnKV.
+
+#### RawKV Value Encode
+
+If the key has RawKV prefix, which are `k{keyspace id prefix}r` or `k{keyspace id prefix}a`, then the value can be either:
+
+1. `{0x0}{data}`
+2. `{0x1}{TTL expire timestamp}{data}`
+
+#### TTL in Titan
```
Connor1996, line 43: For Titan, it's better to support TTL in blob index which needs to change RocksDB. If not, no need to consider Titan encode specifically, it's transparent to the upper level. So the value encode should be the same as `RawKV` without Titan.

text/0069-api-v2.md
```diff
@@ -0,0 +1,242 @@
...
+The `{keyspace prefix id}` is the [keyspace](https://github.com/tikv/rfcs/pull/39) prefix for seperating keys of different keyspace. It should be an vary-length integer whose highest bit of every byte denotes whether the next byte is still part of the integer. The client will fetch the prefix from PD by the keyspace name specified by the user when initializing the client, so it means that the keyspace prefix is valid during the session, in other words, the change on keyspace on PD will not take effect on running seesions.
+
+Note that in TxnKV, the key will be encoded by `Memory Comparable Encoding`. But since the `Memory Comparable Encoding` will not change the starting bytes but only add paddings, there won't be any overlap between RawKV and TxnKV.
+
+#### RawKV Value Encode
+
+If the key has RawKV prefix, which are `k{keyspace id prefix}r` or `k{keyspace id prefix}a`, then the value can be either:
+
+1. `{0x0}{data}`
+2. `{0x1}{TTL expire timestamp}{data}`
```
Connor1996, line 41: still think better to encode as `{0x1}{data}{TTL expire timestamp}` as you can reuse the previous code.

text/0069-api-v2.md
gregwebs: The case of corruption is when both transaction and RawKV perform writes? So we can allow read-only RawKV access? We can state these access conditions in the metadata for the keyspace and we could actually enforce these access conditions in the clients now and to some extent enforce them in the servers, I am not familiar enough to say how much.

Without having some kind of enforcement there will be concerns about corrupting data.

andylokandy: The case is the limitation of the current implementation of CAS. 

Currently, TiKV archives CAS on a key by believing the user would tell TiKV whether the keys in `RawPut` and `RawDelete` requests may be `RawCAS` in order to acquire a latch during the processing. If the user gives wrong information that a key will not be CAS when `RawPut`, and sends `RawCAS` later, the `RawCAS` is not guaranteed.

There are three solutions to this problem:

1. Force `RawPut` and `RawDelete` on every key to acquiring a latch. This will significantly reduce the throughput when CAS is not used.

2. Force `RawPut` and `RawDelete` on `k{keyspace id}a` to acquiring a latch. (In this RFC)

3. Implement `RawCAS` in Raft apply. I'm still drafting an RFC for it.

andylokandy:
@Connor1996 @BusyJay I've removed the raw atomic prefix, PTAL.

Connor1996 (approved):
LGTM

zhongzc (approved):

text/0069-api-v2.md
```diff
@@ -0,0 +1,224 @@
...
+|              | V1 Server | V2 Server |
+| ------------ | --------- | --------- |
+| V1 RawClient | Raw Data  | Forbidden |
+| V1 TxnClient | Txn Data  | TiDB Data |
+| V2 RawClient | Forbidden | Raw Data  |
+| V2 TxnClient | Forbidden | Txn Data  |
+
+### CDC / BR
+
+Since all access to TiDB is unchanged during the upgrade, CDC and BR should work the same after upgrade/downgrade.
```
BusyJay, line 214: /cc @overvenus PTAL

BusyJay (commented):
There are still missing parts in this design, for example, like key codec in the TiKV read/write, the impact on performance, etc.

text/0069-api-v2.md
andylokandy: What do you mean by the `key codec in the TiKV read/write`?

BusyJay: For example, parsing keys or constructing keys in coprocessors.

andylokandy: coprocessor only read and write tidb data, and since tidb data will not change, coprocessor will not change either.

BusyJay: So tidb will not use API V2?

andylokandy: tidb uses api v1

PR #70, opened 2021-07-29
RFC: Record history hotspot information in TiDB_HOT_REGIONS_HISTORY 

qidi1:
Create a new table TIDB_HOT_REGIONS_HISTORY in the INFORMATION_SCHEMAT database to retrieve history hotspot information stored by PD periodically.

### DIFF
--- a/text/0070-pd-record-historical-hot-region.md
+++ b/text/0070-pd-record-historical-hot-region.md
@@ -6,17 +6,33 @@
 
 ## Summary
 
-Create a new table `TIDB_HOT_REGIONS_HISTORY` in the `INFORMATION_SCHEMAT` database to retrieve history hotspot information stored by PD periodically.  
+Create a new table `TIDB_HOT_REGIONS_HISTORY` in the `INFORMATION_SCHEMAT` schema to retrieve history hotspot regions stored by PD periodically.  
 
 ## Motivation
 
-TiDB has a memory table `TIDB_HOT_REGIONS` that provides information about hotspot regions. 
-But it only shows the current hotspot information. This leads to the fact that when DBAs want to query historical hotspot details, 
-they have no way to find the corresponding hotspot information.
+TiDB has a memory table `TIDB_HOT_REGIONS` that provides information about hotspot regions. But it only shows the current hotspot information. This leads to the fact that when DBAs want to query historical hotspot details, they have no way to find the corresponding hotspot information.
 
 ## Background
-According to the [documentation](https://docs.pingcap.com/tidb/stable/information-schema-tidb-hot-regions) for the current `TIDB_HOT_REGIONS` table, 
-the table provides information about hotspot regions shown in the following table: 
+According to the [documentation](https://docs.pingcap.com/tidb/stable/information-schema-tidb-hot-regions) for the current `TIDB_HOT_REGIONS` table,  it can only provides information about recent hotspot regions calculated by PD according to the heartbeat from tikv.  It is inconvenient to obtain hotspot regions of past time, and locate which store the region is. For ease of use, we can store extened hotspot region infomation in PD. The DBA can query hotspot regions within a specified period in such statement in TiDB:
+
+```SQL
+SElECT * FROM information_schema.tidb_hot_regions_history WHERE update_time>='2019-04-01 00:00:00' and update_time<='2019-04-01 00:01:00';
+```
+
+The result is shown below:
+
+```sql
++---------------------+---------+------------+----------+------------+----------+-----------+----------+---------+-------+------------+------------+------------+------------+
+| UPDATE_TIME         | DB_NAME | TABLE_NAME | TABLE_ID | INDEX_NAME | INDEX_ID | REGION_ID | STORE_ID | PEER_ID | TYPE  | HOT_DEGREE | FLOW_BYTES | KEY_RATE   | QUERY_RATE |
++---------------------+---------+------------+----------+------------+----------+-----------+----------+---------+-------+------------+------------+------------+------------+
+| 2019/04/01 00:00:00 | MYSQL   | USER       |        5 | PRIMARY    |        1 |         5 |        6 |      50 | READ  |         23 |  86.264363 |  70.001240 | 37.675487  |
++---------------------+---------+------------+----------+------------+----------+-----------+----------+---------+-------+------------+------------+------------+------------+
+```
+
+## Detailed design
+### Existing TIDB_HOT_REGIONS
+
+Before we introduce `TIDB_HOT_REGIONS_HISTORY`, let's see how `TIDB_HOT_REGIONS` works.
 
 ```SQL
 +----------------+-------------+------+------+---------+-------+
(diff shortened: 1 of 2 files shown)


BusyJay:
Interesting idea! But you should probably send the proposal to https://github.com/pingcap/tidb/issues.

text/0070-pd-record-historical-hot-region.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Motivation
+
+TiDB has a memory table `TIDB_HOT_REGIONS` that provides information about hotspot regions. 
+But it only shows the current hotspot information. This leads to the fact that when DBAs want to query historical hotspot details, 
+they have no way to find the corresponding hotspot information.
+
+## Background
+According to the [documentation](https://docs.pingcap.com/tidb/stable/information-schema-tidb-hot-regions) for the current `TIDB_HOT_REGIONS` table, 
+the table provides information about hotspot regions shown in the following table:
```
Yisaer, line 19: In Background, I think we should provide  what `TiDB_HOT_REGIONS_HISTORY` will look like and in which specific cases or problems, we can use `TiDB_HOT_REGIONS_HISTORY` to solve them and how.

text/0070-pd-record-historical-hot-region.md
```diff
@@ -0,0 +1,163 @@
...
+      HotDegree      int       `json:"hot_degree"`
+      ByteRate       float64   `json:"flow_bytes"`
+      KeyRate        float64   `json:"flow_keys"`
+      QueryRate      float64   `json:"flow_query"`
+      AntiCount      int       `json:"anti_count"`
+      LastUpdateTime time.Time `json:"last_update_time"`
+    }
+    ```
+1. Parse response and fetch the start key and end key of the hot region from region cache or PD by `REGION_ID` to calculate the region's corresponding schema information like：`TABLE_ID`, `TABLE_NAME`, `DB_NAME`, `INDEX_ID`, `INDEX_NAME`.
+1. Return the hotspot information to the upper call.
```
Yisaer, line 58: These should be moved to detailed design, and I think it is enough to mention that the hot regions data is fetched from PD circlely instead expose the detailed api design here.

Yisaer: I also suggestion to provide common examples for querying `HotPeerStatShow` in daily investigament.

text/0070-pd-record-historical-hot-region.md
```diff
@@ -0,0 +1,163 @@
...
+
+### Design in TiDB
+
+1. Add memory table `TIDB_HOT_REGIONS_HISTORY`：
+
+   Create a new memory table `TIDB_HOT_REGIONS_HISTORY` to the existing `INFORMATION_SCHEMA` database.
+
+1. Add pull function：
+
+   Pull  and combine the data from all  PD servers,  and add fields like `table_id`, `index_id`, `db_name`, `table_name`, `index_name` according to the `start_key` and `end_key`  of the hotspot region.
```
Yisaer, line 130: What's gc policy for this table?

text/0070-pd-record-historical-hot-region.md
```diff
@@ -0,0 +1,163 @@
...
+   There are two config options need to be add in PD’s `config.go`:
+
+   * `HisHotRegionSaveInterval`:  time interval for pd to record hotspot region information, default: 15 minutes.
+   * `HisHotRegionTTL`: maximum hold day for his hot region, default: 30 days.
+
+### Design in TiDB
+
+1. Add memory table `TIDB_HOT_REGIONS_HISTORY`：
+
+   Create a new memory table `TIDB_HOT_REGIONS_HISTORY` to the existing `INFORMATION_SCHEMA` database.
```
Yisaer, line 126: I think the gc policy should consider memory consuming. Still I think memory table may not be a good choice especially in big cluster cases, the hot regions count might be large and each tidb-server maintained `TIDB_HOT_REGIONS_HISTORY` while we just only need 1 replication.

text/0070-pd-record-historical-hot-region.md
IcePigZDB: Add a example sql and result row in Background section.

text/0070-pd-record-historical-hot-region.md
IcePigZDB: Move `TIDB_HOT_REGIONS` workflow to Detailed design section and remove the api.

IcePigZDB: It seems to be a more commom way to query `HotPeerStatShow` by `pd-ctl`. Does `TIDB_HOT_REGIONS_HISTORY` need to support `pd-ctl` in the future, if so, the data will take some time to transmit.

text/0070-pd-record-historical-hot-region.md
qidi1: add GC policy in pd

text/0070-pd-record-historical-hot-region.md
qidi1: it not table in memory,All data will be stored in pd's leveldb for persistent storage.In TiDB, TIDB_HOT_REGIONS_HISTORY is a virtual table which construct by data in pd

text/0070-pd-record-historical-hot-region.md
```diff
@@ -0,0 +1,202 @@
...
+  | Record Length (B)| Time Interval (Min)  | Data Size Per Day (MB)   | Data Size Per Month (MB)   |
+  | ---------------- | -------------------- | ------------------------ | -------------------------- |
+  | 328              | 5                    | 90.08789063              | 2702.636719                |
+  | 328              | 10                   | 45.04394531              | 1351.318359                |
+  | 328              | 15                   | 30.02929688              | 900.8789063                |
+
+### Design in PD
+1. Timing write：
+
+     The leader of PD will periodically encrypt  `START_KEY` and `END_KEY`  in data,and write hotspot region data into `LevelDB`.The write interval can be configured. The write fieldes are: `REGION_ID`, `TYPE`,  `HOT_DEGREE`, `FLOW_BYTES`, `KEY_RATE`, `QUERY_RATE`, `STORE_ID`, `PEER_ID`, `UPDATE_TIME`, `START_KEY`,  `END_KEY`.
```
skyzh, line 125: Does PD currently uses `LevelDB`? Maybe we should not add more dependency, especially these requiring cgo, in pd...

text/0070-pd-record-historical-hot-region.md
```diff
@@ -0,0 +1,202 @@
...
+* TiDB can not add fields for deleted Table or Schema.
+
+## Alternatives
+
+Write to TiDB like a normal table.
+* Reuse complete push-down function.
+* Data is written into TiKV to provide disaster tolerance.
+* The `Owner` election solves the problem of who pulls data from PD.
+* Can not support scenarios using TiKV independently.
+
```
skyzh, line 182: What about creating a new tool to periodically pull data from pd and insert them into TiDB or other log services, instead of making this into pd?

zz-jason (commented):
a better place for this RFC is the TiDB repo: https://github.com/pingcap/tidb/tree/master/docs/design

text/0070-pd-record-historical-hot-region.md
IcePigZDB: Yes,`LevelDB` is used  to store region info, [code](https://github.com/tikv/pd/blob/20b00a7c4933e307c5423da0c5047250259e19e9/server/core/region_storage.go#L60).

text/0070-pd-record-historical-hot-region.md
IcePigZDB: If the tool is in TiDB, it is similay with alternative we mention in `Alternatives`, one disadvantage is that it can not support pure TiKV scenario.  As for independent tool like [ecosystem-tool](https://docs.pingcap.com/zh/tidb/stable/ecosystem-tool-user-guide),  we'll talk more about that.

qidi1:
Move to the TiDB repo: https://github.com/pingcap/tidb/tree/master/docs/design

PR #74, opened 2021-09-18
rawkv bulk load: add description for pause merge

peng1999:
Part of https://github.com/tikv/client-java/issues/268


text/0072-online-bulk-load-for-rawkv.md
```diff
@@ -147,6 +147,15 @@ There are some usecases:
 1. Load parquet from S3 to RawKV: Location=S3, FileFormat=Parquet, Encoder=TableEncoder, IngestAPI=RawKV
 2. Load CSV from HDFS to TxnKV: Location=HDFS, FileFormat=CSV, Encoder=TiDBEncoder, IngestAPI=TxnKV
 
+### Import mode & pause merge checker
+
+When ingesting, TiKV should be switched to import mode. Also, merge to empty regions created by repartition operation should be prevented. To achieve this, a PD REST API to pause checkers (including merge checker) should be added. Pause-checker API should be called periodically to keep checker in paused status.
```
marsishandsome, line 152:
```suggestion
When ingesting, TiKV should be switched to import mode. Also, merge the empty regions split by Spark should be prevented. To achieve this, a PD REST API to pause checkers (including merge checker) should be added. Pause-checker API should be called periodically to keep checker in paused status.
```

marsishandsome (approved):
LGTM

text/0072-online-bulk-load-for-rawkv.md
```diff
@@ -147,6 +147,15 @@ There are some usecases:
...
 
+### Import mode & pause merge checker
+
+When ingesting, TiKV should be switched to import mode. Also, merge the empty regions split by Spark should be prevented. To achieve this, a PD REST API to pause checkers (including merge checker) should be added. Pause-checker API should be called periodically to keep checker in paused status.
+
+| URL | body | |
+|---|:---:|---|
+| POST /pd/api/v1/checker/{name} | `{ "delay": <seconds> }` | Pause or resume a checker |
+| GET /pd/api/v1/checker/{name} | -- | Check if a checker is paused |
+
```
andylokandy, line 158: Rather than pausing the whole cluster, is it better to pause a list of given regions that we are going to import? Considering the keyspace scenario, one keyspace may not want to pause mergin g just because the other keyspace is importing.

peng1999: Maybe related: tikv/tikv#8368

This would be better but needs more design.

marsishandsome: @andylokandy agree with you.
We can use `Store import mode` and `pd check` as milestone 1.
And update to `fine-grain control version` in milestone 2.

PR #75, opened 2021-09-29
Update API V2

andylokandy:

### DIFF
--- a/text/0069-api-v2.md
+++ b/text/0069-api-v2.md
@@ -174,6 +174,8 @@ If `storage.api_version=2`:
 
 - Use the new value encode in `RawStore`, `TTLStore` and `sst_importer`.
 
+- Only allow RawKV to access `default` CF.
+
 - If the request's context has `api_version=1`:
   - Reject the request unless it's a TxnKV request and the keys starting with `m` or `t`.
 
--- a/text/0069-api-v2.md
+++ b/text/0069-api-v2.md
@@ -23,17 +23,19 @@ The `API V2` is enabled by a switch on PD. Since it changes the storage encode, 
 
 Once the `API V2` is enabled, the key will be starting with either:
 
-1. `m` or `t`: TxnKV key. Used by TiDB.
-2. `k{keyspace prefix id}x`: TxnKV key.
-3. `k{keyspace prefix id}r`: RawKV key.
+1. `m` and `t`: TxnKV key that will only be used by TiDB.
+2. `x{keyspace prefix id}`: TxnKV key.
+3. `r{keyspace prefix id}`: RawKV key.
 
 The `{keyspace prefix id}` is the [keyspace](https://github.com/tikv/rfcs/pull/39) prefix for seperating keys of different keyspace. It should be an vary-length integer whose highest bit of every byte denotes whether the next byte is still part of the integer. The client will fetch the prefix from PD by the keyspace name specified by the user when initializing the client, so it means that the keyspace prefix is valid during the session, in other words, the change on keyspace on PD will not take effect on running seesions.
+
+`x`, `r` are mode prefixes that indicates the mode the key is belonging to. The mode prefix is ahead of the keyspace id so that TiKV and clients are able to quickly check the mode without parsing the keyspace id in varint format.
 
 Note that in TxnKV, the key will be encoded by `Memory Comparable Encoding`. But since the `Memory Comparable Encoding` will not change the starting bytes but only add paddings, there won't be any overlap between RawKV and TxnKV.
 
 #### RawKV Value Encode
 
-If the key has RawKV prefix, which is `k{keyspace id prefix}r`, then the value can be either:
+If the key has RawKV prefix `r`, then the value can be either:
 
 1. `{0x0}{data}`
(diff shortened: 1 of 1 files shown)


andylokandy:
@zz-jason PTAL

text/0069-api-v2.md
```diff
@@ -23,17 +23,17 @@ The `API V2` is enabled by a switch on PD. Since it changes the storage encode,
 
 Once the `API V2` is enabled, the key will be starting with either:
 
-1. `m` or `t`: TxnKV key. Used by TiDB.
-2. `k{keyspace prefix id}x`: TxnKV key.
-3. `k{keyspace prefix id}r`: RawKV key.
+1. `m` and `t`: TxnKV key that will only be used by TiDB.
+2. `x{keyspace prefix id}`: TxnKV key.
+3. `r{keyspace prefix id}`: RawKV key.
```
zz-jason, line 28: could you elaborate on why this is better in the RFC?

text/0069-api-v2.md
```diff
@@ -93,6 +93,7 @@ The keyspaces are stored in etcd and has the no limitation on the id number.
 3. Flashbacking keyspace:
   3.1. Flashbacking keyspace only affects the metadata, turning the keyspace name and id mapping to viable by clients.
   3.2. If there are multiple deleted keyspaces with the same name, only the last deleted keyspace is flashbacked. Users can flashback all these deleted keyspaces by calling the flashback API multiple times with different new keyspace names.
+4. Reserving keyspace name: "default".
```
zz-jason, line 96: could you elaborate on what's 'default' keyspace is in the RFC?

andylokandy: After consideration, I've removed it.

text/0069-api-v2.md
andylokandy: done

text/0069-api-v2.md
```diff
@@ -23,21 +23,33 @@ The `API V2` is enabled by a switch on PD. Since it changes the storage encode,
...
 
 #### RawKV Value Encode
 
-If the key has RawKV prefix, which is `k{keyspace id prefix}r`, then the value can be either:
+If the key has RawKV prefix `r`, then the value can be either:
 
 1. `{0x0}{data}`
 2. `{0x1}{TTL expire timestamp}{data}`
 
+### Raw CF
```
BusyJay, line 43: I don't think it's a good idea. First, we can detect whether it's a RawKV keys by its prefix; second, new CF means extra maintenance; third, if we add a new type besides RawKV and TxnKV, are we going to add new CFs?

andylokandy: > we can detect whether it's a RawKV keys by its prefix

This is not enough to separate compaction and compaction filter workload, and TxnKv uses memory comparable encoding while RawKV doesn't which makes the keyspace prefix unstable.

> new CF means extra maintenance

do you have a concrete idea on the extra maintenance effort (comparing to mixing two different read/write paths into the same default cf)? maybe I can reduce that effort somehow.

> if we add a new type besides RawKV and TxnKV, are we going to add new CFs?

Exactly. This is also what versioned kv was doing: https://github.com/tikv/tikv/pull/7607

BusyJay: > TxnKv uses memory comparable encoding while RawKV doesn't which makes the keyspace prefix unstable

I don't get it. Isn't RawKV always starts with `r` and `TxnKv` always starts with `x`?

> ...concrete example on the extra maintenance effort..

For example:
1. different CFs needs to be correctly handled during split check;
2. different CF options need to be maintained for different access pattern for each CF;
3. flush also needs to handle different CFs;
4. if new type adds new CF, it will introduce more and more resources and be hard to optimize, especially for the potential new architecture. It's a nightmare to deal with more and more memtables.

> This is also what versioned kv was doing

It's removed now. :)

andylokandy: > I don't get it. Isn't RawKV always starts with r and TxnKv always starts with x?

My mistake, this is the problem of `k{keyspace id}r` which I have changed in this PR.

> different CFs needs to be correctly handled during split check;

I've read through the split check but I didn't see any special on cf. The only place where cf is used is the `LARGE_CFS`  when creating the `MergedIterator`. As long as `raw` is in `LARGE_CFS`, I think the split checker should work as expected.

> different CF options need to be maintained for different access pattern for each CF

I think this should rather support adding raw cf because it gives us a fine grind of RawKV access patterns without trade-offs on TxnKV.

> flush also needs to handle different CFs;

I think I should have converted it in the PR https://github.com/tikv/tikv/pull/11031

> if new type adds new CF, it will introduce more and more resources and be hard to optimize, especially for the potential new architecture. It's a nightmare to deal with more and more memtables.

The first is that we are not going to add more modes in foreseeable future; the second is that introducing individual block cache and memtable from TxnKV and RawKV helps avoid them affecting one another.

> It's removed now. :)

It's me that removed it. But not because it introduced a new cf but because it is simply deprecated and no one is working on it.

andylokandy:
Closing this one. The changes and discussion are moved to https://github.com/tikv/rfcs/pull/79

PR #79, opened 2021-10-26
Update RFC of API v2 according to latest codes

pingyu:
Signed-off-by: pingyu <yuping@pingcap.com>

Update RFC of API v2 according to latest codes.

Rendered: https://github.com/pingyu/tikv-rfcs/blob/rfc_api_v2_pd/text/0069-api-v2.md


andylokandy:
@pingyu I've update the rfc, PTAL

text/0069-api-v2.md
```diff
@@ -105,29 +108,20 @@ To enable API V2, which also enables the keyspace API:
...
+  {
+     name: string,
+     id: int64,
+     created_at: timestamp
+  }
+]
 
->> config keyspaces delete-property <keyspace name> <property-path>
->> config keyspaces flashback <new keyspace name>
+>> config keyspaces create <keyspace name> [--id <Keyspace ID>]
```
pingyu, line 120: Is there any restriction on the provided `Keyspace ID` ? For example, what if the ID is the same as another Keyspace ?

text/0069-api-v2.md
```diff
@@ -105,29 +108,20 @@ To enable API V2, which also enables the keyspace API:
...
+>> config keyspaces delete <keyspace name>
+>> config keyspaces flashback [<new keyspace name>]
```
pingyu, line 122: Suggest describe how `flashback` is implemented.
Besides, `flashback` will delete the entry in `/keyspaces_deleted` (won't it ?), but PD doesn't provide such an API.

andylokandy: updated, PTAL

text/0069-api-v2.md
andylokandy: updated

pingyu: What if two concurrent create requests with different name but same id ?

AFAIK etcd provides [transaction API](https://etcd.io/docs/v3.5/learning/api/#transaction) similar to CAS. It seems to be difficult to solve such a race condition.

text/0069-api-v2.md
pingyu: It is possible to be conflicted with another concurrent create request with 1) same name, or 2) different name but same id.

Etcd's transaction API can help to deal with the same name condition. But it would be difficult to solve the same id case.

andylokandy: I think we allow multiple keyspace names referring to a same id.

pingyu: I think it would be better not to allow this.

Users would assume that two keyspaces are different if they have different names. Especially when two keyspaces have same id silently caused by the race condition. The two keyspaces will corrupt each other.

pingyu:
@iosmanthus @haojinming @zeminzhou PTAL, thanks~

haojinming (approved):
LGTM

ystaticy:
LGTM

text/0069-api-v2.md
```diff
@@ -222,3 +206,12 @@ Read `api_version` in kvdb and decode data using the corresponding version.
 Upgrade to the latest TiKV Go Client and use `V1` mode.
 
 ## Unresolved questions
+
+*TBD*
+
+[Change Data Capture]: https://en.wikipedia.org/wiki/Change_data_capture
+[Memory Comparable Encoding]: https://github.com/facebook/mysql-5.6/wiki/MyRocks-record-format#memcomparable-format
+[Causal Consistency]: https://en.wikipedia.org/wiki/Causal_consistency
+[happened before]: https://en.wikipedia.org/wiki/Happened-before
```
sunxiaoguang, line 215:
```suggestion
[Happened before]: https://en.wikipedia.org/wiki/Happened-before
```

pingyu:
@sunxiaoguang All comments are addressed (see [here](https://github.com/tikv/rfcs/pull/79/files/ca867b26647fa5a3c8b0c7cfb4d4ce8ef86571ff..dc977ba6dadb908408200e28cfa485f656c80f75). The force-push is to fix DCO error), PTAL, thanks~

sunxiaoguang (approved):

iosmanthus (approved):
LGTM

PR #80, opened 2021-11-22
Add RFC: RawKV Stale Read

andylokandy:

## Summary

Allow eventually consistency on RawKV.

## Motivation

Eventually consistency is acceptable in some scenarios when using RawKV. In this case, RawKV is allowed to break linear consistency and therefore improve the cluster-wise throughput.

[Rendered](https://github.com/andylokandy/rfcs/blob/staleread/text/0080-rawkv-stale-read.md)


BusyJay:
What's the guarantees stale read provides? It seems there is no guarantee at all, client may read a key's value and then read no keys and then read the keys again even if there is no changes to the key.

andylokandy:
@BusyJay This is exactly what eventually consistency means. To address these problems, the client should select the store with a satisfied RPO so that the freshness will not regress.

text/0080-rawkv-stale-read.md
```diff
@@ -0,0 +1,61 @@
...
+### Implementation details
+
+1. In kvproto, add `direct_read` to `RaftRequestHeader`:
+
+    ```diff
+        // raft_cmdpb.proto
+        message RaftRequestHeader {
+            ...
+            bool replica_read = 8;
+    +       bool direct_read = 9;
```
BusyJay, line 50: We already have the `STALE_READ` write batch flags, the new field may not be necessary.

PR #82, opened 2021-11-29
make region size dynamic

BusyJay:


text/0082-dynamic-size-region.md
```diff
@@ -0,0 +1,129 @@
...
+of 96MiB, we set the max region size to 10GiB (before compression). To get around the problem of
+concurrency, we introduce new concept bucket that is smaller ranges within a region. To make hotspot be
+recoginized and scheduled quickly, we choose the make the size of a region dynamic.
+
+![dynamic size](../media/dynamic-size.png)
+
+### Dynamic size
+
+The hotter a region is, the smaller its size becomes. To make it simplified, we choose 512MiB for hot regions
+and 10GiB for cold regions. So there are two types of split, hotspot split and general size split. Hotspot
```
feitian124, line 49: > To make it simplified, we choose 512MiB for hot regions
and 10GiB for cold regions.

here is for simplify the description of the idea in this document, or simplify the implements?
512Mib is fixed? how much performance increase expected by reduce hot reigion size to 1/2 ?

text/0082-dynamic-size-region.md
```diff
@@ -0,0 +1,129 @@
...
+split is triggered by PD. General size split is triggered by TiKV itself.
+
+10GiB is just an example, it's allowed to change to a bigger or smaller value.
+
+### Bucket
+
+A region is split into several buckets logically. We will collect query stats by buckets and report the bucket
+to PD. For hotspot regions, buckets are split by scan. For cold regions, buckets are split by approximate
```
feitian124, line 57: does split buckets mean split region? what is your mean `by scan`. could you make bucket concept more clear?

text/0082-dynamic-size-region.md
BusyJay: Simplify implementation. I don't think it's 1/2, but rather 5%. It's configurable. Using a small size is to make it easy to schedule around and balance quickly.

text/0082-dynamic-size-region.md
BusyJay: I'm not sure why do you think it's related to split region. As described in text and image, buckets are logical segments of a region range.

text/0082-dynamic-size-region.md
feitian124: sorry i misread it. 5% and configurable sounds reasonable.

text/0082-dynamic-size-region.md
feitian124: as per your description, bucket is a logical concept. 
no matter how you split buckets, they still in the same region. unless the owner region split, i am not sure how can a `physical region` splits into `logic buckets` improve performance.

text/0082-dynamic-size-region.md
```diff
@@ -0,0 +1,129 @@
...
+    the leader, serving a range scan can trigger at most 10GiB IO. To warm it up, we can extend
+    the existing pre transfer leader mechanism. Leader will send its bucket statics to follower
+    in pre transfer leader, follower can do a pre scan for hot buckets before accepting
+    transferring leader.
+
+### Flow report
+
+Buckets statistics are reported in a standalone stream. They will be reported more frequently
+than original region heartbeats, which is 60s. The new stream is expected to be reported every
+10 seconds when have changes.
```
5kbpers, line 105: Consider reporting asynchronously to free up raftstore CPU usage.

BusyJay: What do you mean by "asynchronously"? It's reported in PD worker, which will access a sharing state. raftstore will not be involved.

text/0082-dynamic-size-region.md
```diff
@@ -0,0 +1,129 @@
...
+concurrency, we introduce new concept bucket that is smaller ranges within a region. To make hotspot be
+recoginized and scheduled quickly, we choose the make the size of a region dynamic.
+
+![dynamic size](../media/dynamic-size.png)
+
+### Dynamic size
+
+The hotter a region is, the smaller its size becomes. To make it simplified, we choose 512MiB for hot regions
+and 10GiB for cold regions. So there are two types of split, hotspot split and general size split. Hotspot
+split is triggered by PD. General size split is triggered by TiKV itself.
```
bufferflies, line 50: will it include hotspot and region size??

BusyJay: What does "it" mean?

bufferflies: it means PD.

text/0082-dynamic-size-region.md
```diff
@@ -0,0 +1,129 @@
...
+    presplit + scatter.
+- Because followers don’t serve read, so their data is usually not in cache. After transferring
+    the leader, serving a range scan can trigger at most 10GiB IO. To warm it up, we can extend
+    the existing pre transfer leader mechanism. Leader will send its bucket statics to follower
+    in pre transfer leader, follower can do a pre scan for hot buckets before accepting
+    transferring leader.
+
+### Flow report
+
+Buckets statistics are reported in a standalone stream. They will be reported more frequently
```
bufferflies, line 103: In past, TiKV seperates  read and write in different channel. should it  uniform them in one channel .

text/0082-dynamic-size-region.md
BusyJay: No, PD only needs to do hotspot split.

text/0082-dynamic-size-region.md
BusyJay: Either way is OK. It depends on the implementation difficulty.

text/0082-dynamic-size-region.md
```diff
@@ -0,0 +1,129 @@
...
+regions and nodes. For normal read requests, TiKV will need to split its range into smaller buckets
+according to statics to increase concurrency. When TiDB wants to do a scan, it sends the RPC once,
+and TiKV will split the requests into smaller concurrent jobs.
+
+In the past design, follower read has also been discussed to offload works from leader. But TiDB
+can’t predict the progress of a follower, so latency can also become unpredictable. It’s more
+controllable to let PD split hotspots and schedule to get balance.
+
+Now that a region can be in 10GiB, a full scan can make the response exceed the limit of gRPC,
+which is 4GiB. So instead of unary, we need to use server streaming RPC to return the response as
```
hicqu, line 86: > we need to user server streaming RPC
Is it confirmed or just a candidate solution? An alternative is make clients send ranges instead of Region, maybe we can discuss about which is better.

BusyJay: It's a candidate solution. I prefer client sends a unary request with the whole range. It's simple and efficient. Otherwise, TiDB may waste CPU at batch RPCs. A similar problem can be found here: https://github.com/tikv/client-go/issues/342.

text/0082-dynamic-size-region.md
```diff
@@ -0,0 +1,129 @@
...
+### Replication
+
+A large region can take minutes to be replicated, PD should make cost estimation on actual size
+instead of fixed value. To avoid logs being truncated during sending a snapshot, we should set
+the log gc limit based on the actual size too. Raft engine should be GA before dynamic size
+region, so we don't need to worry about the extra write amplifications.
+
+There are ways to optimize the replications, sending a sub-LSM tree for example. But these designs
+may be complicated. And we rely on further designs like separating LSM tree to optimize the cost
+for all.
```
hicqu, line 116: Ingesting a large snapshot may cause a large compaction, so that the SST file can stay at L0 too long time and slow down or stop write requests. To resolve the problem, we can split the large region into several SST files.

text/0082-dynamic-size-region.md
```diff
@@ -0,0 +1,129 @@
...
+    in pre transfer leader, follower can do a pre scan for hot buckets before accepting
+    transferring leader.
+
+### Flow report
+
+Buckets statistics are reported in a standalone stream. They will be reported more frequently
+than original region heartbeats, which is 60s. The new stream is expected to be reported every
+10 seconds when have changes.
+
+### Replication
```
hicqu, line 107: Seems with TiKV's customed SstPartitioner, a large region can cause a large SST file? It can increase write amplification, so I think we also need to update the customed SstPartitioner.

BusyJay: Why SstPartitioner can cause large SST file? I think the SST file size is controlled by `target_file_size_base` configuration.

hicqu: Ah I forget this.

text/0082-dynamic-size-region.md
BusyJay: Yes, I forget it. Now it's added.

solotzg:
What's the conclusion about https://github.com/tikv/pd/issues/4326? Will PD split region by placement rules or not? @disksing

BusyJay:
In the long term, I think the answer is Yes, especially when all system without tiflash have less regions and perform better. But in the short term, we can hold the task to avoid too much work.

text/0082-dynamic-size-region.md
coocood: But streaming API is deprecated in TiDB

coocood: It would be much easier to retry if TiDB split the request into buckets, each bucket can be sent concurrently.

If there is leader change, a single bucket request can be send to the new leader, the previous succeed bucket requests don't need to retry.

text/0082-dynamic-size-region.md
```diff
@@ -0,0 +1,136 @@
...
+and 10GiB for cold regions. So there are two types of split, hotspot split and general size split. Hotspot
+split is triggered by PD. General size split is triggered by TiKV itself.
+
+10GiB is just an example, it's allowed to change to a bigger or smaller value.
+
+### Bucket
+
+A region is split into several buckets logically. We will collect query stats by buckets and report the bucket
+to PD. For hotspot regions, buckets are split by scan. For cold regions, buckets are split by approximate
+range size. Every bucket should have the size about 128MiB. Their ranges and stats are all reported to PD. PD
```
coocood, line 58: If all the bucket stats are reported to PD, it would consume a lot of memory.
We only need to report bucket stats when read/write workload exceed a threshold.

So for most of the regions, we only need to report the range of buckets.
The total memory on PD would be greatly reduced.

text/0082-dynamic-size-region.md
```diff
@@ -0,0 +1,136 @@
...
+
+The hotter a region is, the smaller its size becomes. To make it simplified, we choose 512MiB for hot regions
+and 10GiB for cold regions. So there are two types of split, hotspot split and general size split. Hotspot
+split is triggered by PD. General size split is triggered by TiKV itself.
+
+10GiB is just an example, it's allowed to change to a bigger or smaller value.
+
+### Bucket
+
+A region is split into several buckets logically. We will collect query stats by buckets and report the bucket
```
coocood, line 56: Do we need to introduce a bucket version field in `metapb::Region`?
So each time we update the buckets, the bucket version increases.

text/0082-dynamic-size-region.md
```diff
@@ -0,0 +1,136 @@
...
+Hibernate regions can be deprecated, and lease renew by heartbeats can be enabled by default for
+all regions. This can bring significant improvement for read tail latency.
+
+After the region becomes larger, the chance to hit 1PC becomes higher. And the RPC count in a
+transaction can also be reduced. However, the hotspot chance is also increased. For writes, we rely
+on PD to split the hotspot in time and increase concurrency. In addition, we can also use unorder
+apply to make single region apply logs faster. Since unorder apply is a standalone feature, I’m not
+going into details here.
+
+For read hotspots, split should be triggered by PD, which can utilize the global statics from all
```
coocood, line 76: s/statics/stats

text/0082-dynamic-size-region.md
nolouch, line 56: Currently the query stat is collect by region-level, such as the range scan [start_key, end_key], may cover bucket 1 to10 in this region, how to split it by buckets?  
If the scan will divide into sub-scan in buckets, and we only know the stats in buckets. It will make us cannot know the split cost.  for example,  the scan request in this region may split into two scan requests or only one scan request(region-level), it's decided by the split key whether in the range scan.

text/0082-dynamic-size-region.md
BusyJay: The bucket size is about 128MiB, which means it can consume as much as the memory current small regions use. As far as I know, it's not an issue yet.

text/0082-dynamic-size-region.md
BusyJay: No, buckets are not shared by regions. It's a peer's own property. So `metapb::Region` should not be modified. I'm not sure if version is useful, as final consistency is acceptable.

text/0082-dynamic-size-region.md
```diff
@@ -0,0 +1,136 @@
...
+To reduce the count of regions, we can set the size of region to a larger value. More precisely, instead
+of 96MiB, we set the max region size to 10GiB (before compression). To get around the problem of
+concurrency, we introduce new concept bucket that is smaller ranges within a region. To make hotspot be
+recoginized and scheduled quickly, we choose the make the size of a region dynamic.
+
+![dynamic size](../media/dynamic-size.png)
+
+### Dynamic size
+
+The hotter a region is, the smaller its size becomes. To make it simplified, we choose 512MiB for hot regions
```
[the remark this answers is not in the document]
BusyJay: >  512 is too big if the hotspot is small tables

It's configurable. As long as only a small number of hotspots are maintained, the size of hotspot doesn't matter.

> will the split size set a lower limit?

Yes, and it's configured as 512MiB for now.

text/0082-dynamic-size-region.md
```diff
@@ -0,0 +1,136 @@
...
+split is triggered by PD. General size split is triggered by TiKV itself.
+
+10GiB is just an example, it's allowed to change to a bigger or smaller value.
+
+### Bucket
+
+A region is split into several buckets logically. We will collect query stats by buckets and report the bucket
+to PD. For hotspot regions, buckets are split by scan. For cold regions, buckets are split by approximate
+range size. Every bucket should have the size about 128MiB. Their ranges and stats are all reported to PD. PD
+can detect hot buckets from all regions and decide whether need to split new hotspots and schedules them to
```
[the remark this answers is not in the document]
BusyJay: > I think this part need more supplements

It's covered at https://github.com/tikv/rfcs/pull/82/files#diff-ec76198abf750cd69e94e0e14564aea21f6f2de06d0bf85829d6d68969f36d20R107-R110.

> ...from the boundaries of buckets

Yes, the split key is chosen from the boundaries of buckets. As long as hotspots are not split into two accidentally, I think it's not worse than query statistics. And query statistics is unpredictable and changed frequently, makes buckets hard to maintain.

text/0082-dynamic-size-region.md
BusyJay: Scan request will be split into buckets and access statistics are collected at bucket level.

I'm not sure what does "split cost" mean.

coocood: If the buckets is not included in `metapb::Region` then there should be a new type `metapb::RegionBuckets`.
TiDB can query this info from PD, and use them to build requests.

text/0082-dynamic-size-region.md
BusyJay: If the request is split in TiKV side, snapshot is retrieved once, so there is no such case that retry some of the buckets due to leader change.

text/0082-dynamic-size-region.md
```diff
@@ -0,0 +1,136 @@
...
+Increasing the size only may not have significant improvement on all problems, it can introduce
+higher replication cost than before.
+
+## Alternatives
+
+For the problems on cost of too many regions, we can also optimize TiKV itself to make it run
+faster and capable to handle more regions. But the dynamic design make it possible to adjust
+current architecture and introduce more significant changes in the future.
+
+## Unresolved questions
```
rleungx, line 136: What about the compatibility?

text/0082-dynamic-size-region.md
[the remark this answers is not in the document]
BusyJay: It's unspecified. Either hibernate or awake feel good to me.

text/0082-dynamic-size-region.md
BusyJay: Yes, exact the key and statistics collection.

text/0082-dynamic-size-region.md
rleungx: When a region becomes hot, does the bucket need to be re-split since it uses a different split policy?

text/0082-dynamic-size-region.md
```diff
@@ -0,0 +1,136 @@
...
+We have hibernate raft (#23) to get around the most significant problem 2. But it's not always
+working as expected:
+
+- Workload with random access can still activate a lot of regions;
+- Hibernated region needs extra time to renew its lease, which lead to high tail latency;
+- It complicates the design of CDC/BR, which requires access all replicas instead of just leader now;
+- GC will still wake up all regions and cause periodical high usage.
+
+There are two source of region creations: size/keys split and table split. Table split is disabled by
+default in TiKV. Even if it splits a lot of regions, small one will be merged later.
```
rleungx, line 30: We may need more details about merge, e.g. should we try to move a smaller region instead of a larger region.

text/0082-dynamic-size-region.md
BusyJay: Yes, works most like current region scan split.

text/0082-dynamic-size-region.md
BusyJay: It depends on the emergency. If it's hotspot, then it should be split first and then moved. Otherwise, no need to split unless impossible to get size balance.

text/0082-dynamic-size-region.md
coocood: If TiKV is slow, and TiDB encounter frequently time out, retry the whole region request would make the situation much worse.

BusyJay: Then TiDB can change the behavior of retrying, it can know what ranges are missing easily.

text/0082-dynamic-size-region.md
5kbpers: Got, that is exactly what I mean.

coocood (approved):

text/0082-dynamic-size-region.md
BusyJay: I left unary vs streaming as an unresolved question and will make decision by evaluation on maintainability and performance.

text/0082-dynamic-size-region.md
```diff
@@ -61,6 +61,20 @@ balance.
 
 ![dynamic size buckets](../media/dynamic-size-buckets.png)
 
+A new bucket metadata will be added to kvproto:
+```
+message Buckets {
```
5kbpers, line 66: Should we allocate an ID for each bucket to reduce the size of requests from clients?

5kbpers: The ID could consist of term and version of the region and a number.

BusyJay: > ...reduce the size of requests from clients?

I don't get it. Client should send the range it wants to scan, so it has to be precisely start key and end key. How can ID reduce the size?

Bucket is supposed to be changed independently from region. For example, client can re-use last overlapping buckets even a region split.

5kbpers: It can optimize for queries that need to scan the whole bucket.

BusyJay: Because bucket keys are generated by even split, so the range of a query is probably different from the bucket edges.

text/0082-dynamic-size-region.md
```diff
@@ -0,0 +1,189 @@
...
+several smaller files. Each snapshot files should not be larger than 128MiB.
+
+There are ways to optimize the replications, sending a sub-LSM tree for example. But these designs
+may be complicated. And we rely on further designs like separating LSM tree to optimize the cost
+for all.
+
+### Compatibility
+
+As buckets don't modify existing metadata, so it's backward compatible. When upgrading from small
+regions, PD may trigger a lot of merge to get a large region size. This procedure should be made
```
rleungx, line 151: If we have some regions with heavy requests in different stores, we don't need to merge it until it cools down. But the calculation way of statistics of the bucket may be different.

text/0082-dynamic-size-region.md
nolouch: split will make RPC increase, in some scenarios we need to make a trade-off.

text/0082-dynamic-size-region.md
```diff
@@ -0,0 +1,189 @@
...
+    in pre transfer leader, follower can do a pre scan for hot buckets before accepting
+    transferring leader.
+
+### Flow report
+
+Buckets statistics are reported in a standalone stream. They will be reported more frequently
+than original region heartbeats, which is 60s. The new stream is expected to be reported every
+10 seconds when have changes.
+
+PD will maintain top N hot spot buckets and split all of those buckets that exceed P of
```
nolouch, line 130: I think 
- The Bucket needs to have a unique identifier. and it should like region split, use right derive. otherwise PD cannot know it is a stable hot bucket or it is a traffic peak.
- In most cases, it can be split from the boundary key of the bucket. but still need make load-base split key strategy or other strategy by PD can be accept.

nolouch (commented):
As the Region is bigger, the access pattern in one region will more complicated. especially on read-write pattern, it will make hot region have read-write conflicts. I think how to better split and help scheduling to solve the problem of read-write conflict is a more complicated and challenging problem that we will encounter in the follow-up.

text/0082-dynamic-size-region.md
BusyJay: > ... it should like region split, use right derive

This can be very complicated. How about handling load trace on PD side? For example, if the buckets are [a, d), [d, f) in the past, and the new version is [a, c), [c, e), [e, f). Then history H[a, d) should be inherited by both [a, c) and [c, e), history H[d, f) is inherited by both [c, e) and [e, f). The overlapped histories should be sum up. And the inherited history statistic can be optionally multiply by a factor like 0.8 to make process smoothly.

> ...make load-base split key strategy or other strategy by PD...

It's up to PD to split by any keys. It's recommended to choose bucket boundary.

text/0082-dynamic-size-region.md
BusyJay: I don't think we need to consider the RPC count for now. But I agree there might be many choices on choosing split keys. I don't think we can get a best solution without a lot of experiments. The RFC avoid describing concrete algorithm on purpose to not make false assumptions.

text/0082-dynamic-size-region.md
nolouch: I think this method also complicated. Do you mean that the bucket boundary may be different every time? [a, d)[d ,f) change to [a, c), [c, e), [e, f); d is gone.

BusyJay: It can change, but it only changes when there are many updates.

text/0082-dynamic-size-region.md
```diff
@@ -0,0 +1,203 @@
...
+
+![dynamic size buckets](../media/dynamic-size-buckets.png)
+
+A new bucket metadata will be added to kvproto:
+```
+message Buckets {
+    uint64 region_id = 1;
+    uint64 version = 2; // A hint indicate if keys have changed.
+    repeated bytes keys = 3;
+    repeated uint64 read = 4;
```
nolouch, line 70: How about using a separate message inside here,  we may add other statics.
```
repeated BucketStats stats = 3 
```

nolouch (commented):
lgtm

5kbpers (approved):
LGTM

tonyxuqqi:
LGTM

text/0082-dynamic-size-region.md
tonyxuqqi: @coocood Good point. Do you know what's the roughly memory consumption per region in PD today?
For 128MB bucket and 1PB total database size, it's about 800 million buckets.

tonyxuqqi: @BusyJay For bucket stats report, will both leader and follower need to report their buckets? And I guess it's very likely for the same region, the bucket range may be different. If only leader reports the buckets stats, then PD will have to add the follower buckets stats when calculating the machine level stats, or report machine level stats separately.

text/0082-dynamic-size-region.md
```diff
@@ -0,0 +1,203 @@
...
+to PD. For hot spot regions, buckets are split by scan. For cold regions, buckets are split by approximate
+range size. Every bucket should have the size about 128MiB. Their ranges and stats are all reported to PD. PD
+can detect hot buckets from all regions and decide whether need to split new hot spots and schedules them to
+balance.
+
+![dynamic size buckets](../media/dynamic-size-buckets.png)
+
+A new bucket metadata will be added to kvproto:
+```
+message Buckets {
```
tonyxuqqi, line 66: Should it be renamed as BucketStats?
Also where's the key range definition?

text/0082-dynamic-size-region.md
```diff
@@ -0,0 +1,203 @@
...
+region, so we don't need to worry about the extra write amplifications.
+
+To avoid ingesting a large snapshot cause large compaction, we need to split the snapshot into
+several smaller files. Each snapshot files should not be larger than 128MiB.
+
+There are ways to optimize the replications, sending a sub-LSM tree for example. But these designs
+may be complicated. And we rely on further designs like separating LSM tree to optimize the cost
+for all.
+
+### Compatibility
```
tonyxuqqi, line 158: If we implement the feature in the way that buckets are just optional and optimization for big region size, then by its nature it's backward compatible. 
And there's nothing prevent user from using small region,  so this should be well supported in the long term as well.  For small to medium size cluster, it's likely user will keep the small region size.

text/0082-dynamic-size-region.md
```diff
@@ -0,0 +1,203 @@
...
+
+### Dynamic size
+
+The hotter a region is, the smaller its size becomes. To make it simplified, we choose 512MiB for hot regions
+and 10GiB for cold regions. So there are two types of split, hot spot split and general size split. Hot spot
+split is triggered by PD. General size split is triggered by TiKV itself.
+
+10GiB is just an example, it's allowed to change to a bigger or smaller value.
+
+### Bucket
```
tonyxuqqi, line 54: Since we measure the read/write load at bucket level, which means it's about 128MB size. How can we further split a hot bucket? Today we have load-based split, which may cut the small region by the access pattern.

text/0082-dynamic-size-region.md
nolouch: Currently, the small region will report statistics on both the leader and follower. the TopN(1000) region of this node will be reported not all.  may buckets can keep the same behavior.

text/0082-dynamic-size-region.md
```diff
@@ -0,0 +1,129 @@
...
+of 96MiB, we set the max region size to 10GiB (before compression). To get around the problem of
+concurrency, we introduce new concept bucket that is smaller ranges within a region. To make hotspot be
+recoginized and scheduled quickly, we choose the make the size of a region dynamic.
+
+![dynamic size](../media/dynamic-size.png)
+
+### Dynamic size
+
+The hotter a region is, the smaller its size becomes. To make it simplified, we choose 512MiB for hot regions
+and 10GiB for cold regions. So there are two types of split, hotspot split and general size split. Hotspot
```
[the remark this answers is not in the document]
BusyJay: Yes, it's mentioned that they are configurable.

text/0082-dynamic-size-region.md
```diff
@@ -0,0 +1,129 @@
...
+apply to make single region apply logs faster. Since unorder apply is a standalone feature, I’m not
+going into details here.
+
+For read hotspots, split should be triggered by PD, which can utilize the global statics from all
+regions and nodes. For normal read requests, TiKV will need to split its range into smaller buckets
+according to statics to increase concurrency. When TiDB wants to do a scan, it sends the RPC once,
+and TiKV will split the requests into smaller concurrent jobs.
+
+In the past design, follower read has also been discussed to offload works from leader. But TiDB
+can’t predict the progress of a follower, so latency can also become unpredictable. It’s more
```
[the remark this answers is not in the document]
BusyJay: I think it can, but it doesn't.

text/0082-dynamic-size-region.md
BusyJay: It's defined as the third field `keys`.

text/0082-dynamic-size-region.md
BusyJay: > ...the bucket range may be different.

It doesn't matter. Only leader's buckets are used and follower's stats will be merged into leaders using similar algorithm described in flow report section.

> the TopN(1000) hot region's stats of this node will be reported not all.

I'm worried about accuracy and message layout. I prefer to report all metrics and do filter on the PD side. Let's see how it performs first.

text/0082-dynamic-size-region.md
BusyJay: I don't have a clear idea now. I will leave it as an unresolved question. Maybe we can split the bucket further. For example, besides controlling the size of a bucket, also require the minimal count of a bucket. So even a region becomes small, it can still be split further.

hicqu (approved):

PR #83, opened 2021-12-26
RFC: Maintain incremental analyze result

Little-Wallace:
Signed-off-by: Little-Wallace <bupt2013211450@gmail.com>

### DIFF
--- a/text/0083-incremental-analyze.md
+++ b/text/0083-incremental-analyze.md
@@ -1,0 +1,43 @@
+# RFC: Incremental analyze table
+
+
+## Summary
+
+This proposal introduce a method to store some result of analyze request from TiDB in TiKV Storage, so that we do not calculate the whole table when we run `analyze table` next time.
+
+## Motivation
+
+The initial motivation was to reduce the impact of `analyze table` on the cluster. And then I found that most of the calculations are not needed.
+
+- In produce environment, most of regions will not be update in a long time.
+- The analyze request will cost a lot of CPU resource but the result data of it is very small (In most cases, no more than 0.1% of the total data).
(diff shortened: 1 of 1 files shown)


text/0083-incremental-analyze.md
```diff
@@ -0,0 +1,43 @@
...
+    metapb.Region region = 1;
+    uint64 file_size = 2;
+    repeated KeyValue data = 3;
+    uint64 version = 4;
+    SnapshotMeta meta = 5;
+    StatisticCache statistic_cache = 6;
+}
+```
+
+* When a TiKV receives an analyze request it will check the cache and applied index of this region. If the cache data exists and `applied_index` of the cache equals to the current `applied_index` of this region, it means that the region has never write any user data or split.
```
PatrickNicholas, line 36: > If the cache data exists and `applied_index` of the cache equals to the current `applied_index` of this region, it means that the region has never write any user data or split.

So if any member changes, this cache will be invalied?

Little-Wallace: Member changes? Do you mean that TiKV add one peer for this region?

PatrickNicholas: yes

PatrickNicholas: s/member changes/member changed

Little-Wallace: Conf change would not make cache invalid.

Little-Wallace: I will try some trick method to avoid some conf change(include member changed or leader transfer) making the cache invalid.

PR #85, opened 2021-12-28
improve the robust of scheduler

bufferflies:


text/0083-Improve-the-robust-of-balance-scheduler.md
```diff
@@ -0,0 +1,65 @@
...
+- RFC PR: [https://github.com/tikv/rfcs/pull/85](https://github.com/tikv/rfcs/pull/83)
+- Tracking Issue: [https://github.com/tikv/pd/issues/](https://github.com/tikv/pd/issues/4428)
+
+## Summary
+
+Make scheduler more robust for dynamic region size.
+
+## Motivation
+
+We have observed many different situations when the region size is different. The major drawback comes from this aspects:
```
tonyxuqqi, line 12: typo: this->these

text/0083-Improve-the-robust-of-balance-scheduler.md
```diff
@@ -0,0 +1,65 @@
...
+
+1. Balance region scheduler pick source store in order of store's score, the second store will be picked after the first store has not met some filter or retry times exceed fixed value, this problem is also exist in target pick strategy.
+2. Operator has an import effect on region leader, and the leader is responsible in the operator life cycle. But the region leader will not be limited by any filter.
```
tonyxuqqi, line 15: what is import effect?

text/0083-Improve-the-robust-of-balance-scheduler.md
```diff
@@ -0,0 +1,65 @@
...
+3. There are some factor that influence execution time of operator such as region size, IO limit, cpu load. PD needs to be more flexible to manage operator's life not fixed config.
```
tonyxuqqi, line 16: factor--> factors
operator's life --> operator's timeout threshold

text/0083-Improve-the-robust-of-balance-scheduler.md
```diff
@@ -0,0 +1,65 @@
...
+4. PD should know some global config about TiKV like `region-max-size`, `region-report-interval`. This config should synchronize with PD.
+
+## Detailed design
+
+### Store pick strategy
+
+It can arrange all the store based on label, like TiKV and TiFlash and allow low score group has more chance to scheduler. But the first score region should has highest priority to be selected.
```
tonyxuqqi, line 23: Please provide more detail about label and how low score group can have more chance to schedule

text/0083-Improve-the-robust-of-balance-scheduler.md
```diff
@@ -0,0 +1,65 @@
...
+
+#### Consider Influence to leader
+
+Normally, one operator is made of region, source store and target store, the key works finished by region leader such as snapshot generate, snapshot send. It is not friendly to the leader if majority operator is add follow.
+
```
tonyxuqqi, line 28: "It is not friendly to the leader if majority operator is add follow"---could you explain a bit more detail regarding this?  Because for a region leader, the add follower operator should be up to 1 or 2. Or you mean the whole store.

text/0083-Improve-the-robust-of-balance-scheduler.md
```diff
@@ -0,0 +1,65 @@
...
+### Operator control
+
+#### Store limit cost
+
+Different size region occupy tokens should be different. Maybe can use this formula:
+
+![](https://latex.codecogs.com/gif.image?\dpi{200}&space;\bg_white&space;Influence=\sum_{i=0}^{j}step_{i}.Influence&space;\newline&space;Cost&space;=&space;200*ln{\frac{region_{size}}{100KiB}})
+
+Cost equals 200 if operator influence is 1Mb or equals 600 if operator influence is 1gb.
```
tonyxuqqi, line 39: 1Mb--->1MB
1gb --->1GB
ln should be log10? Otherwise 200*ln1MB/100KB won't be 200.
Why use formula log regionsize/100KB, this makes little difference when region size is 500MB and 1GB, for example--but the actual cost difference of 500MB and 1GB is much bigger.

text/0083-Improve-the-robust-of-balance-scheduler.md
```diff
@@ -0,0 +1,65 @@
...
+
+#### Operator life cycle
+
+The operator life cycle can divide into some stages: create, executing(started), complete. PD will check operator stage by region heart beats and cancel operator if one operator‘s running time exceed the fixed value(10m).
```
tonyxuqqi, line 43: 10m may not be enough. 
Why not make it configurable. 

What if Tikv crashed, the heartbeat request may not carry the operator info anymore, how will PD handle it?

text/0083-Improve-the-robust-of-balance-scheduler.md
```diff
@@ -0,0 +1,65 @@
...
+
+It will be better if we can calculate every step expecting execute duration by major factor includes region size, IO limit and operator concurrency like this:
```
tonyxuqqi, line 45: There's no per snapshot limit and thus each snapshot's speed cannot be just 100MB/6.
Also  snapshot generator duration cannot be ignore in single RocksDB instance version, as we have to scan to get the region's snapshot. 
I think the total time threshold should be pretty conservative, probably 1hr at least.

text/0083-Improve-the-robust-of-balance-scheduler.md
```diff
@@ -0,0 +1,65 @@
...
+
+![](https://latex.codecogs.com/gif.image?\dpi{200}&space;\bg_white&space;V=\frac{io_limit}{sending_{count}+receiving_{count}}=\frac{100Mb/s}{3+3}=16.7Mb/s\newline&space;T_{transfer}=\frac{10Gb}{16.7Mb/s}=598s\newline&space;T_{total}=T_{generator}+T_{transfer}+T_{apply})
+
+The snapshot generator duration can ignore because it doesn't need to scan. The apply snapshot duration will finish in minute level if it needs to load hot buckets.
+
+### Sync global config
+
+There are some global config that all components need to synchronize like `region-max-size`, `io-limit`. Using ETCD api to implement global config may be a good idea like [this](<[https://github.com/pingcap/tidb/pull/31010/files](https://github.com/pingcap/tidb/pull/31010/files)>)
```
tonyxuqqi, line 53: My concern is that such configuration is likely same for the whole cluster. Does it worth to ask every tikv report these values?
Even if some TiKv changes the value, then which region size value PD will use for calculating the formula above then?
To me, the region size should be cluster level config.

text/0083-Improve-the-robust-of-balance-scheduler.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Drawbacks
+
+## Alternatives
+
+Removing peer may not influence the cluster performance, it can be replace by leader store limit.
+
+Canceling operator can depends on TiKV not by PD, but TiKV should notify PD after canceled one operator.
+
+## Questions
```
tonyxuqqi, line 63: We noticed that when scale-out new node, it's much faster to move the data over if the new node is not the leader until the data is moved over. But of course in some scenarios, we hope the new node can act as leader ASAP. So it will be better to have an option to enable both scenarios. 

For scale-in an old node, for current implementation, is transfer leaders the first step before moving data?

text/0083-Improve-the-robust-of-balance-scheduler.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Summary
+
+Make scheduler more robust for dynamic region size.
+
+## Motivation
+
+We have observed many different situations when the region size is different. The major drawback comes from this aspects:
+
+1. Balance region scheduler pick source store in order of store's score, the second store will be picked after the first store has not met some filter or retry times exceed fixed value, this problem is also exist in target pick strategy.
```
rleungx, line 14:
```suggestion
1. Balance region scheduler pick source store in order of store's score, the second store will be picked after the first store has not met some filter or retry times exceed fixed value, this problem also exists in target pick strategy.
```

text/0083-Improve-the-robust-of-balance-scheduler.md
```diff
@@ -0,0 +1,65 @@
+# Improve the robust of balance scheduler
```
rleungx, line 1: It seems this RFC only talks about the balance region scheduler. Maybe you need to change the title.

text/0083-Improve-the-robust-of-balance-scheduler.md
rleungx, line 16: Also
```suggestion
3. There are some factor that influence execution time of operator such as region size, IO limit, CPU load. PD needs to be more flexible to manage operator's life not fixed config.
```

text/0083-Improve-the-robust-of-balance-scheduler.md
```diff
@@ -0,0 +1,65 @@
...
+
+### Store pick strategy
+
+It can arrange all the store based on label, like TiKV and TiFlash and allow low score group has more chance to scheduler. But the first score region should has highest priority to be selected.
+
+#### Consider Influence to leader
+
+Normally, one operator is made of region, source store and target store, the key works finished by region leader such as snapshot generate, snapshot send. It is not friendly to the leader if majority operator is add follow.
+
+It will add new store limit as new limit type to decrease leader loads of every store.
```
rleungx, line 29: Can you provide more details about how should we use this new limit type to decrease the load?

text/0083-Improve-the-robust-of-balance-scheduler.md
```diff
@@ -0,0 +1,65 @@
...
+
+### Sync global config
+
+There are some global config that all components need to synchronize like `region-max-size`, `io-limit`. Using ETCD api to implement global config may be a good idea like [this](<[https://github.com/pingcap/tidb/pull/31010/files](https://github.com/pingcap/tidb/pull/31010/files)>)
+
+## Drawbacks
+
+## Alternatives
+
+Removing peer may not influence the cluster performance, it can be replace by leader store limit.
```
rleungx, line 59:
```suggestion
Removing peer may not influence the cluster performance, it can be replaced by leader store limit.
```

text/0083-Improve-the-robust-of-balance-scheduler.md
bufferflies: In past, scale-in an node will evict leader first. 
The new region peers act as leader should depend on configs in  different scenarios.

text/0083-Improve-the-robust-of-balance-scheduler.md
bufferflies: In the PR 31010, TiKV doesn't need to report pd and watch this config.

text/0083-Improve-the-robust-of-balance-scheduler.md
bufferflies: I disagree this threshold should be conservative. If one region has two region peer and needs one peer, it will wait one hour to try another target store  if the origin target store is down or other reason.

text/0083-Improve-the-robust-of-balance-scheduler.md
bufferflies: I means that region leade generates and sends snapshot will occupy cpu and io resources.

text/0085-Improve-the-robust-of-balance-scheduler.md
```diff
@@ -0,0 +1,63 @@
...
+
+It will be better if we can calculate every step expecting execute duration by major factor includes region size, IO limit and operator concurrency like this:
+
+![](https://latex.codecogs.com/gif.image?\dpi{200}&space;\bg_white&space;V=\frac{io_limit}{sending_{count}+receiving_{count}}=\frac{100Mb/s}{3+3}=16.7Mb/s\newline&space;T_{transfer}=\frac{10Gb}{16.7Mb/s}=598s\newline&space;T_{total}=T_{generator}+T_{transfer}+T_{apply})
+
+The snapshot generator duration can be ignored because it doesn't need to scan. The applying snapshot duration will finish at minute level if it needs to load hot buckets.
+
+### Sync global config
+
+There are some global config that all components need to synchronize like `region-max-size`, `io-limit`. Using ETCD api to implement global config may be a good idea like [this](<[https://github.com/pingcap/tidb/pull/31010/files](https://github.com/pingcap/tidb/pull/31010/files)>)
```
disksing, line 51: Do you mean sync the config from TiKV to PD, or from PD to TiKV?

PR #86, opened 2022-01-12
RawKV Change Data Capture

pingyu:
This proposal introduces the technical design of RawKV Change Data Capture.
([Rendered](https://github.com/pingyu/tikv-rfcs/blob/rawkv-cdc/text/0083-rawkv-change-data-capture.md)).

Related issue: tikv/tikv#11745


text/0083-rawkv-cross-cluster-replication.md
```diff
@@ -0,0 +1,164 @@
...
+
+## Summary
+
+This proposal introduces the technical design of RawKV Cross Cluster Replication.
+
+## Motivation
+
+Customers are deploying TiKV clusters as non-transaction Key-Value (RawKV) storage. But the lack of *Cross Cluster Replication* is a obstacle to deploy disaster recovery clusters for providing more highly available services.
+
+The ability of cross cluster replication will help TiKV be more adoptable to industries such as Banking and Finance.
```
overvenus, line 14: Could you list key challenges of the use case "Banking" plus "Cross Cluster Replication"? It would be helpful for understanding the design in the RFC.

Here is my initial thought,
* How to address high latency issue between clusters?
* How do we define the consistency of raw kv?

pingyu: @overvenus I add a chapter about "Key Challenges & Solutions", and specify that we will implement RawKV features in TiCDC as new modules for isolation with original TiDB related logics.
PTAL, thanks~

pingyu:
@BusyJay @Little-Wallace PTAL, thanks~

text/0083-rawkv-cross-cluster-replication.md
```diff
@@ -0,0 +1,183 @@
...
+
+##### 2.2.1 Physical Part of Timestamp
+
+Physical part of HLC is acquired from TSO of [PD](https://github.com/tikv/pd).
+
+TSO is a global monotonically increasing timestamp, which help the generation of timestamp be independent to local clock of machine, and be immune to issues such as reverse between machine reboot.
+
+Physical part is refreshed by repeatedly acquiring TSO in a period of `500ms`, to keep it being closed the real world time. And it can tolerate fault of TSO no longer than `30s`, to keep time-related metrics such as RPO reasonable.
+
+Besides, on startup, physical part must be initialized by a successful TSO.
```
zz-jason, line 72: what happens if it failed to acquire ts from PD?

text/0083-rawkv-cross-cluster-replication.md
```diff
@@ -0,0 +1,183 @@
...
+To make deletions be captured as change data, we turn physical deletion to logical deletion, i.e., set a *deleted* flag, other than physically delete the entry.
+
+The *garbage collection* of deleted data is implemented by setting TTL to 25 hours after, and *GC-ed* by compaction filter. At the same time, start timestamp (i.e. `start-ts`) of **TiKV-CDC** replication task should not be earlier than 24 hours before.
+
+### 4. Encoding
+
+Timestamp & deleted flag are encoded as following:
+
+```
+r{keyspace id}{user key}{MCE Padding}{^timestamp:u64}: {user value}{expire ts}{meta flags}
```
zz-jason, line 105: should the `keyspace id` part be removed?

text/0083-rawkv-cross-cluster-replication.md
```diff
@@ -0,0 +1,183 @@
...
+  * More read latency in tens of microseconds.
+  * A little more write overhead caused by [MCE Padding](https://github.com/facebook/mysql-5.6/wiki/MyRocks-record-format#memcomparable-format).
+  * An extra Garbage Collection is required (used with TxnKV).
+  * Possible more space occupied (related to update pattern).
+
+### 5. Incremental Scan
+
+On replication task resumed after paused, we scan *KVDB* to get the change data during pause by extracting timestamp encoded in key. The scan will be heavy as we actually scan all the keys, so we should avoid scanning as possible, and limit scan rate.
+
+*Incremental Scan in SQL / TxnKV scenario scans write_cf to get change data scope, which is much cheaper. So an alternative to reduce cost for RawKV is writing timestamp to write_cf other than encoded in key, but will introduce an additional write for each put request. We tend to the lower delay of writing now.*
```
zz-jason, line 125: what's the solution for writing `timestamp`?

text/0083-rawkv-cross-cluster-replication.md
```diff
@@ -0,0 +1,183 @@
...
+
+* Sink:
+  
+  * A new *TiKV sink* to batch put data into downstream TiKV cluster, with configurable concurrency, batch size, and backoff parameters.
+
+## Prototype
+
+We have developed a [prototype](https://github.com/pingyu/tikv/issues/1) to verify feasibility of this proposal. All of the correctness validation cases are passed. And the benchmark results are as follows:
+
+| Case                          | gPRC P99 Duration (ms) | gPRC Avg Duration (ms) |
```
zz-jason, line 159: should we update the test result based on encoding timestamp into the key part?

text/0083-rawkv-cross-cluster-replication.md
pingyu: The raw write operations (raw_put, raw_del, etc.) will fail.

text/0083-rawkv-cross-cluster-replication.md
pingyu: No. Why it should be removed ?

text/0083-rawkv-cross-cluster-replication.md
pingyu: It would be: in addition to `{key}:{value}` in default_cf, write an entry `{key}:{timestamp}` to write_cf, to indicated the latest change timestamp of the key.

text/0083-rawkv-cross-cluster-replication.md
pingyu: Ah yes. We will update the result when we finish the benchmark this week.

text/0083-rawkv-cross-cluster-replication.md
```diff
@@ -0,0 +1,183 @@
...
+
+Customers are deploying TiKV clusters as non-transaction Key-Value (RawKV) storage. But the lack of *Cross Cluster Replication* is a obstacle to deploy disaster recovery clusters for providing more highly available services.
+
+The ability of cross cluster replication will help TiKV to be more adoptable to industries such as Banking and Finance.
+
+## Key Challenges & Solutions
+
+#### Capturing Change
+
+As there is no timestamp in RawKV entries, it is not possible to capture which and when data is changed. So in this proposal, we add timestamp to data. See "2. Add Timestamp to Data" chapter.
```
andylokandy, line 20:
```suggestion
As there is no timestamp in RawKV entries, it is not possible to capture which and when data is changed. So in this proposal, we add timestamp to data. See "(Add Timestamp to Data)[#2. Add Timestamp to Data]" chapter.
```

text/0083-rawkv-cross-cluster-replication.md
```diff
@@ -0,0 +1,183 @@
...
+
+#### Consistence
```
andylokandy, line 22:
```suggestion
#### Consistency
```

text/0083-rawkv-cross-cluster-replication.md
```diff
@@ -0,0 +1,183 @@
...
+
+**TiKV-CDC**, a fork of [TiCDC](https://github.com/pingcap/tiflow), is the tool for replicating incremental data of TiKV, which gets change data from TiKV's observer component, with high performance, high available, and scalable.
+
+Besides, **TiKV-CDC** can also provide the ability to connect to other components, e.g. Message Queues, which will help TiKV integrate with customers' data systems, and further extend the applicable scenarios of TiKV.
+
+![rawkv-cdc](../media/rawkv-cdc.png)
+
+### 2. Add Timestamp to Data
+
+As a kind of [Change Data Capture](https://en.wikipedia.org/wiki/Change_data_capture), timestamp (or version) is necessary to indicate which data is changed, and when. While RawKV doesn't have such a thing before, we need to add timestamp to the RawKV data.
```
andylokandy, line 48:
```suggestion
As a kind of [Change Data Capture](https://en.wikipedia.org/wiki/Change_data_capture), timestamp (or version) is necessary to indicate which and when data is changed.
```

text/0083-rawkv-cross-cluster-replication.md
```diff
@@ -0,0 +1,183 @@
...
+
+#### 2.1 Requirement
+
+Among requests of a key, *Order of Timestamp* must be consistent with sequence of data flush to disk in TiKV.
```
andylokandy, line 52:
```suggestion
Among requests of a key, the timestamp must be monotonic with the sequence of data flushed to disk in TiKV.
```

text/0083-rawkv-cross-cluster-replication.md
```diff
@@ -0,0 +1,183 @@
...
+
+We use [HLC](https://cse.buffalo.edu/tech-reports/2014-04.pdf) method to generate timestamp, which not only is easy to use (by being closed to real world time), but also can capture causality.
+
+##### 2.2.1 Physical Part of Timestamp
+
+Physical part of HLC is acquired from TSO of [PD](https://github.com/tikv/pd).
+
+TSO is a global monotonically increasing timestamp, which help the generation of timestamp be independent to local clock of machine, and be immune to issues such as reverse between machine reboot.
+
+Physical part is refreshed by repeatedly acquiring TSO in a period of `500ms`, to keep it being closed the real world time. And it can tolerate fault of TSO no longer than `30s`, to keep time-related metrics such as RPO reasonable.
```
andylokandy, line 70: Why `30s`? And will raw write fails if TSO is offline over 30s?

text/0083-rawkv-cross-cluster-replication.md
andylokandy: Even without Raw CDC, TiKV can't startup when PD is out of service.

text/0083-rawkv-cross-cluster-replication.md
```diff
@@ -0,0 +1,183 @@
...
+
+Besides, on startup, physical part must be initialized by a successful TSO.
+
+##### 2.2.2 Logical Part of Timestamp
+
+Logical part is advanced on every write request.
+
+Moreover, logical part is advanced on *Leader Transfer*. As TiKV a distributed system, every store in TiKV cluster has a instance of timestamp generator. As every key is existed in only one region and only one peer (the leader) in a region can write, it's safe to generate timestamp locally in each store, except for leader transfer.
```
andylokandy, line 78:
```suggestion
Moreover, the logical part is advanced on *Leader Transfer*. As TiKV is a distributed system, every store in the TiKV cluster has an instance of timestamp generator. As every key exists in only one region and only one peer (the leader) in the region can write, it's safe to generate timestamps locally in each store, except for leader transfer.
```

text/0083-rawkv-cross-cluster-replication.md
```diff
@@ -0,0 +1,183 @@
...
+
+On leader transfer, the timestamp generated by other store would be bigger than the store where new leader located, which will violate the causal consistency requirement. We solve this by this method: Every peer observes raft message applied, and keep the maximum timestamp of the region (called `Region-Max-TS`). On become leader, this peer advances timestamp generator of store to be bigger than `Region-Max-TS`.
```
andylokandy, line 80:
```suggestion
On leader transfer, the timestamp generated by another store could be larger than the store where the new leader is located, which will violate the causal consistency requirement. We solve this by this method: Every peer observes the raft message applied and keeps the maximum timestamp of the region (called `Region-Max-TS`). To become leader, this peer advances the timestamp generator of the store to be bigger than `Region-Max-TS`. 
```

text/0083-rawkv-cross-cluster-replication.md
```diff
@@ -0,0 +1,183 @@
...
+
+Logical part is advanced on every write request.
+
+Moreover, logical part is advanced on *Leader Transfer*. As TiKV a distributed system, every store in TiKV cluster has a instance of timestamp generator. As every key is existed in only one region and only one peer (the leader) in a region can write, it's safe to generate timestamp locally in each store, except for leader transfer.
+
+On leader transfer, the timestamp generated by other store would be bigger than the store where new leader located, which will violate the causal consistency requirement. We solve this by this method: Every peer observes raft message applied, and keep the maximum timestamp of the region (called `Region-Max-TS`). On become leader, this peer advances timestamp generator of store to be bigger than `Region-Max-TS`. 
+
+Besides, other region changes should be carefully handled:
+
+* On region merge: The `Region-Max-TS` of the region is the bigger one of the original two regions.
```
andylokandy, line 84:
```suggestion
* On region merge: The `Region-Max-TS` of the region is the max of the two original regions.
```

text/0083-rawkv-cross-cluster-replication.md
andylokandy: Ah, it's because we've decided not to include keyspace in API V2.

text/0083-rawkv-cross-cluster-replication.md
zz-jason: ok. it would be better to add iit to the RFC for disambiguation.

text/0083-rawkv-cross-cluster-replication.md
pingyu: Er... OK, so what's the latest proposal of API V2 ?

text/0083-rawkv-cross-cluster-replication.md
pingyu: En... I think this is not a serious decision. Seems that offline of TSO doesn't break the correctness, except some time-related metrics.
Maybe we can drop this restriction, and report failure metrics. And add some tests on this scenario.

Meanwhile, TiKV-CDC is highly dependent to PD. When PD is offline, the replication is not working.

andylokandy: > TiKV-CDC is highly dependent to PD. When PD is offline, the replication is not working. 

Yes, but the write on the main TiKV may still work.

> Maybe we can drop this restriction, and report failure metrics.

Agree

pingyu:
@zz-jason @andylokandy 
Comments addressed. PTAL, thanks~

text/0083-rawkv-cross-cluster-replication.md
```diff
@@ -67,9 +67,11 @@ Physical part of HLC is acquired from TSO of [PD](https://github.com/tikv/pd).
 
 TSO is a global monotonically increasing timestamp, which help the generation of timestamp be independent to local clock of machine, and be immune to issues such as reverse between machine reboot.
 
-Physical part is refreshed by repeatedly acquiring TSO in a period of `500ms`, to keep it being closed the real world time. And it can tolerate fault of TSO no longer than `30s`, to keep time-related metrics such as RPO reasonable.
+Physical part is refreshed by repeatedly acquiring TSO in a period of `500ms`, to keep it being closed the real world time.
 
-Besides, on startup, physical part must be initialized by a successful TSO.
+On startup, physical part must be initialized by a successful TSO, to ensure that HLC is larger than last running. RawKV write operations will fail if the initialization has not been completed. (TiKV also can't startup when PD is out of service).
+
+After that, HLC can tolerate fault of TSO. RawKV operations are normal, but time-related metrics such as RPO is not reasonable. Meanwhile, as TiKV-CDC being highly dependent to PD (for cluster and task management, checkpoint, etc.), the replication will not be working.
```
andylokandy, line 74: What exactly is the "After **that**"? From the context, it is the TSO failure on startup but I think it should be TSO failure after startup. And I think the time-related metrics may not be affected unless the clock shift is crazy enough...

pingyu: It is "After successful initialization". I'm sorry for the not clear enough.

We do not advance HLC by local clock when TSO is offline, which will cause the time-related metrics problem.

Using local clock on fault of TSO need some other logics to ensure causality, such as saving maximum HLC on storage to avoid timestamp reverse between startup.

I do not include the "Falling Back to Local Clock" in this proposal for simplicity, as the PD offline is rare, and when it was offline it would cause more severe issues than the problem of time-related metrics.

andylokandy: > Using local clock on fault of TSO need some other logics to ensure causality, such as saving maximum HLC on storage to avoid timestamp reverse between startup.

If I understand it right, the `RegionMaxTs` and `StoreMaxTs` should contain both physical time and logical time, thus we always know what is the latest physical time and therefore safe to use the local clock.

pingyu: Yes it is safe during running.
But unsafe will happen on restart. If the local clock shift backward between TiKV stop & restart, it's possible to get a smaller timestamp than last running.

pingyu: CockroachDB has a description on this restart issue [here](https://github.com/cockroachdb/cockroach/blob/d5ee674ec410f384df9ffb6f737e9b4e2c922589/pkg/util/hlc/doc.go#L152-L160). We can refer to it.

andylokandy: > Yes it is safe during running.

Yes, that's my point. As long as the TiKV doesn't restart, the metrics of RPO and write operation should not affect by PD failure.

andylokandy (approved):
rest LGTM

text/0083-rawkv-cross-cluster-replication.md
```diff
@@ -0,0 +1,185 @@
...
+
+In general, if request `a` ["happened before"](https://en.wikipedia.org/wiki/Happened-before) `b`, then `Timestamp(a) < Timestamp(b)`. As to cross cluster replication, we provide [Causal Consistency](https://en.wikipedia.org/wiki/Causal_consistency) by keeping order of the timestamp the same as sequence of data flush to disk in TiKV. Downstream systems apply data according to the timestamp order, will always lead to the consistent result.
+
+At the same time, as RawKV doesn't provide cross-rows transaction and snapshot isolation, we treat requests of different keys as concurrent, to reduce complexity and improve performance.
+
+#### 2.2 Timestamp Generation
+
+Timestamp is generated by TiKV internally, to get a better overall performance and client compatibility.
+
+We use [HLC](https://cse.buffalo.edu/tech-reports/2014-04.pdf) method to generate timestamp, which not only is easy to use (by being closed to real world time), but also can capture causality.
```
zz-jason, line 62: could you add move info about how many bits are physical and logical timestamp occupies?

text/0083-rawkv-cross-cluster-replication.md
```diff
@@ -0,0 +1,185 @@
...
+
+##### 2.2.1 Physical Part of Timestamp
+
+Physical part of HLC is acquired from TSO of [PD](https://github.com/tikv/pd).
+
+TSO is a global monotonically increasing timestamp, which help the generation of timestamp be independent to local clock of machine, and be immune to issues such as reverse between machine reboot.
+
+Physical part is refreshed by repeatedly acquiring TSO in a period of `500ms`, to keep it being closed the real world time.
```
zz-jason, line 70: is the period `500ms` configurable?

text/0083-rawkv-cross-cluster-replication.md
zz-jason: it would be better if you can explain the reason with an example. For example:

In a 3 AZ deployment:
- PD leader is located in AZ1
- TiKV region leaders are uniformly distributed in AZ1 - AZ3
- There is a partial network partition that happens: AZ1 can not communicate with AZ2, but they can both communicate with AZ3, TiKV Client (for example, client-java) can connect to all the three AZs

Would the write requests to AZ2 TiKV Region be blocked because AZ2 can not sync physical timestamp with PD leader in AZ1?

text/0083-rawkv-cross-cluster-replication.md
```diff
@@ -0,0 +1,185 @@
...
+
+On startup, physical part must be initialized by a successful TSO, to ensure that HLC is larger than last running. RawKV write operations will fail if the initialization has not been completed. (TiKV also can't startup when PD is out of service).
+
+After that, HLC can tolerate fault of TSO. RawKV operations are normal, but time-related metrics such as RPO is not reasonable. Meanwhile, as TiKV-CDC being highly dependent to PD (for cluster and task management, checkpoint, etc.), the replication will not be working.
+
+##### 2.2.2 Logical Part of Timestamp
+
+Logical part is advanced on every write request.
+
+Moreover, the logical part is advanced on *Leader Transfer*. As TiKV is a distributed system, every store in the TiKV cluster has an instance of timestamp generator. As every key exists in only one region and only one peer (the leader) in the region can write, it's safe to generate timestamps locally in each store, except for leader transfer.
```
zz-jason, line 80: > the logical part is advanced on *Leader Transfer*

when assigning a timestamp for a key-value put by a client, the physical part is generated by the local store's machine time, but how to get the logical part? I'm a little confused about this sentence, could you explain with some examples?

text/0083-rawkv-cross-cluster-replication.md
```diff
@@ -0,0 +1,185 @@
...
+
+* On region merge: The `Region-Max-TS` of the region is the max of the two original regions.
+
+* On region split: The `Region-Max-TS` of the new region is copied from the original one.
+
+#### 2.3 Resolved Timestamp
+
+**TiKV-CDC** using *Resolved Timestamp* (or `resolved-ts`) to indicate the largest transaction timestamp of the replication task, to help downstream avoid getting partial data of a transaction. Downstream streaming system can also use `resolved-ts` as [*watermark*](https://www.oreilly.com/radar/the-world-beyond-batch-streaming-102/).
+
+As no transaction support, in RawKV scenario we don't need scanning locks in *lock_cf*. But we should notify `resolved-ts` event generation that the minimum timestamp of data that on the way from *proposed* to *applied* (by [`resolver.track_lock`](https://github.com/tikv/tikv/blob/v5.0.4-20211201/components/resolved_ts/src/resolver.rs#L47-L57)).
```
zz-jason, line 94: > But we should notify `resolved-ts` event generation that the minimum timestamp of data that on the way from *proposed* to *applied* (by [`resolver.track_lock`](https://github.com/tikv/tikv/blob/v5.0.4-20211201/components/resolved_ts/src/resolver.rs#L47-L57)).

I'm a little confused, does it mean we need to record the minimal timestamp of data that is proposed but not applied?

text/0083-rawkv-cross-cluster-replication.md
```diff
@@ -0,0 +1,185 @@
...
+
+### 3. Deletion
+
+To make deletions be captured as change data, we turn physical deletion to logical deletion, i.e., set a *deleted* flag, other than physically delete the entry.
+
+The *garbage collection* of deleted data is implemented by setting TTL to 25 hours after, and *GC-ed* by compaction filter. At the same time, start timestamp (i.e. `start-ts`) of **TiKV-CDC** replication task should not be earlier than 24 hours before.
```
zz-jason, line 100: > The *garbage collection* of deleted data is implemented by setting TTL to 25 hours after.

could you add some explanation on why setting it to 25 hours instead of 1 second?

text/0083-rawkv-cross-cluster-replication.md
pingyu: Done, PTAL~

text/0083-rawkv-cross-cluster-replication.md
pingyu: Done, PTAL~

text/0083-rawkv-cross-cluster-replication.md
pingyu: I improve the words here. PTAL~

text/0083-rawkv-cross-cluster-replication.md
pingyu: OK. Done, PTAL~

text/0083-rawkv-cross-cluster-replication.md
pingyu: Yes. I add more words for this, PTAL~

text/0083-rawkv-cross-cluster-replication.md
pingyu: I add a diagram and some words for this. PTAL~

text/0083-rawkv-cross-cluster-replication.md
pingyu: Yes. I add a diagram and improve some words, PTAL~

text/0083-rawkv-cross-cluster-replication.md
pingyu: Sure. Done, PTAL~

pingyu:
@zeminzhou @haojinming PTAL, thanks~

text/0083-rawkv-change-data-capture.md
```diff
@@ -0,0 +1,152 @@
...
+
+## Detailed Design
+
+### TiKV
+
+#### Resolved-ts
+
+We will use [memory locks] to trace the "inflight" RawKV writes: acquire a lock before scheduling the writes, and release the lock after (asynchronously) send response to client.
+
+Then, CDC worker in TiKV is able to get the **resolved-ts** as `min(timestamp)-1` of all locks. If there is no lock, use current TSO as **resolved-ts**.
```
haojinming, line 81: As concurrency manager [lock the ts before key write](https://github.com/tikv/tikv/blob/a80ab9880d6c25fe9b962e582846360dd229783b/src/storage/mod.rs#L1921-L1925). The `resolved-ts` use the lock ts is enough, no need `-1`.

text/0083-rawkv-change-data-capture.md
```diff
@@ -0,0 +1,152 @@
...
+
+This proposal introduces the technical design of **RawKV Change Data Capture** (abbr. **RawKV CDC**).
+
+## Motivation
+
+Customers are deploying TiKV clusters as non-transactional key-value (**RawKV**) storage for application router or other scenarios. When **RawKV** being more and more widely adopted, [Change Data Capture] become a necessary feature to meet the high availability requirement of industries (*e.g.* banking and finance, to deploy disaster recovery clusters).
+
+*(Now we will focus on the scenario of **Cross Cluster Replication**, i.e. replication between main and recovery TiKV clusters. Design for other scenarios will be supplemented later.)*
+
+In this proposal we introduces **TiKV-CDC**, a new component for **RawKV CDC**, which captures changed data from upstream/main TiKV cluster, and replicates to downstream/recovery TiKV cluster in real time.
```
haojinming, line 16:
```suggestion
In this proposal we introduce **TiKV-CDC**, a new component for **RawKV CDC**, which captures changed data from upstream/main TiKV cluster, and replicates to downstream/recovery TiKV cluster in real time.
```

text/0083-rawkv-change-data-capture.md
```diff
@@ -0,0 +1,152 @@
...
+- RFC PR: https://github.com/tikv/rfcs/pull/86
+- Tracking Issue: https://github.com/tikv/tikv/issues/11965
+
+## Summary
+
+This proposal introduces the technical design of **RawKV Change Data Capture** (abbr. **RawKV CDC**).
+
+## Motivation
+
+Customers are deploying TiKV clusters as non-transactional key-value (**RawKV**) storage for application router or other scenarios. When **RawKV** being more and more widely adopted, [Change Data Capture] become a necessary feature to meet the high availability requirement of industries (*e.g.* banking and finance, to deploy disaster recovery clusters).
```
haojinming, line 12:
```suggestion
Customers are deploying TiKV clusters as non-transactional key-value (**RawKV**) storage for application router or other scenarios. With **RawKV** being more and more widely adopted, [Change Data Capture] becomes a necessary feature to meet the high availability requirement of industries (*e.g.* banking and finance, to deploy disaster recovery clusters).
```

text/0083-rawkv-change-data-capture.md
```diff
@@ -0,0 +1,152 @@
...
+
+Data in downstream cluster is expected to be consistent with upstream cluster. To be consistent:
+
+**1st**, as **RawKV** does not provide transaction semantic, **RawKV CDC** will not guarantee the order of writes for different keys between upstream and downstream cluster.
+
+**2nd**, [TiKV API V2] provides causality consistence, that if two writes (of the same key) `A` [Happened Before] `B`, then `Timestamp(A)` < `Timestamp(B)`. **RawKV CDC** guarantee the same causality consistence as well.
+
+**3rd**, [TiKV API V2] does not guarantee that a **RawKV** entry with smaller timestamp must be observed first. So **TiKV-CDC** will sort entires by timestamp before send to downstream.
+
+But how **TiKV-CDC** knows it will not receive any even earlier entry from TiKV after the previous batch of data had been sent to downstream ? So we need **Resolved Timestamp** (abbr. **resolved-ts**). **Resolved-ts** is similar to [Watermark], by which TiKV indicating that all entries earlier have been observed. When **TiKV-CDC** receives a `resolved-ts`, all sorted entries with `timestamp <= resolved-ts` are safe to write to downstream.
```
haojinming, line 36:
```suggestion
But how **TiKV-CDC** knows it will not receive any even earlier entry from TiKV after the previous batch of data had been sent to downstream ? So we need **Resolved Timestamp** (abbr. **resolved-ts**). **Resolved-ts** is similar to [Watermark], by which TiKV indicats that all entries earlier have been observed. When **TiKV-CDC** receives a `resolved-ts`, all sorted entries with `timestamp <= resolved-ts` are safe to write to downstream.
```

text/0083-rawkv-change-data-capture.md
```diff
@@ -0,0 +1,152 @@
...
+
+### Consistency and Resolved Timestamp
+
+Data in downstream cluster is expected to be consistent with upstream cluster. To be consistent:
+
+**1st**, as **RawKV** does not provide transaction semantic, **RawKV CDC** will not guarantee the order of writes for different keys between upstream and downstream cluster.
+
+**2nd**, [TiKV API V2] provides causality consistence, that if two writes (of the same key) `A` [Happened Before] `B`, then `Timestamp(A)` < `Timestamp(B)`. **RawKV CDC** guarantee the same causality consistence as well.
+
+**3rd**, [TiKV API V2] does not guarantee that a **RawKV** entry with smaller timestamp must be observed first. So **TiKV-CDC** will sort entires by timestamp before send to downstream.
```
haojinming, line 34:
```suggestion
**3rd**, [TiKV API V2] does not guarantee that a **RawKV** entry with smaller timestamp must be observed first. So **TiKV-CDC** will sort entries by timestamp before send to downstream.
```

text/0083-rawkv-change-data-capture.md
```diff
@@ -0,0 +1,152 @@
...
+
+For availability, we utilize a **Checkpoint** mechanism. **Checkpoint** is the latest **resolved-ts** that has been written to downstream successfully, and persist in [ETCD]. When **TiKV-CDC** cluster starts after planned or unexpected stop, replication tasks will resume from the **Checkpoint**.
+
+**Checkpoint** will be affected by **Garbage Collection**. **Checkpoint** requires that all versions since **Checkpoint** are kept, otherwise **TiKV-CDC** would miss some changes. So **TiKV-CDC** will set a **Service Safe Point** on timestamp the same as **Checkpoint**, with **TTL** defaults to `24` hours. The **TTL** means that **TiKV-CDC** and replicate tasks should be resumed in no more than **TTL**, otherwise the **Checkpoint** will be invalid.
+
+Finally, in this proposal, one replication task will be processed in only one **Capture**, to make it right more easily. Splitting a task to sub-tasks will get better performance and scalability, but we must be caution to avoid any overlap or hole among sub-tasks, otherwise the correctness will be violated. Sub-tasks will be considered later.
```
haojinming, line 56:
```suggestion
Finally, in this proposal, one replication task will be processed in only one **Capture**, to make it right more easily. Splitting a task to sub-tasks will get better performance and scalability, but we must be cautious to avoid any overlap or hole among sub-tasks, otherwise the correctness will be violated. Sub-tasks will be considered later.
```

text/0083-rawkv-change-data-capture.md
```diff
@@ -0,0 +1,152 @@
...
+Customers require that [Recovery Point Objective] (abbr. **RPO**) should be no more than tens of seconds. In **RawKV CDC**, [RPO] is measured by the **Checkpoint Lag** metric, the interval between **Checkpoint** and **TSO** at the time.
+
+Main factors affecting **Checkpoint Lag** includes:
+
+- TiKV internal:
+  - **Resolved-ts Lag**: the interval between **resolved-ts** and **TSO**. Which approximately equals to the longest duration of all RawKV writes at that moment. In most circumstances, it should be no more than several seconds.
+- Data transmission on network:
+  - **Network latency** cross cities would be `30` milliseconds or even more. But as we are using the high-performance [gRPC] as underlying implementation of networking, and writing to downstream by batch, we don't concern with network latency now.
+- **TiKV-CDC** internal:
+  - **Sorter** would be the bottleneck, as it 1) process data from all regions, 2) holds too many data if **resolved-ts** or downstream is blocked, 3) sorts data in disks when data is too big to fit in memory.
```
haojinming, line 69:
```suggestion
  - **Sorter** would be the bottleneck, as it 1) process data from all regions, 2) holds too much data if **resolved-ts** or downstream is blocked, 3) sorts data in disks when data is too big to fit in memory.
```

pingyu:
@haojinming All comments are addressed. PTAL~

haojinming (approved):
LGTM

zeminzhou (approved):
LGTM～

sunxiaoguang (approved):

sunxiaoguang (approved):

PR #87, opened 2022-01-13
PD: supports multiple level meta data space

zhangjinpeng1987:
Signed-off-by: zhangjinpeng1987 <zhangjinpeng@pingcap.com>

PD supports multiple level meta data space.

### DIFF
--- a/text/0083-multi-level-meta-data-space.md
+++ b/text/0083-multi-level-meta-data-space.md
@@ -1,0 +1,52 @@
+# Multi-level Meta Data Space
+
+## Summary
+
+PD can handle multiple users' meta data, each user only see its own meta data.
+
+## Motivation
+
+PD is the mata data center of TiKV Cluster, the meta data contains TiKV instances' addr, user data range(region)
+information, the number of replicas of data, etc.
+
+But PD currently only can handle the key space [min-key, max-key], it means that PD only can handle one user's 
+meta data, it is not possible when multiple users want share the same PD cluster as their meta system. There are
+two different scenarios that require PD can handle multiple user' meta data:
+
+1. Multiple TiKV Cluster share the same PD cluster. Because the minimal demplyment of a TiKV Cluster is 3 TiKV 3 PD,
+but it is not cost-effect if every small cluster has 3 dedicated meta data node.
+2. There are Multiple tenant in the same TiKV Cluster, each tenant has it own meta data, each tenant's key range can
+contains any key in the range of [min-key, max-key].
+
+## Detailed design
+
+Change the meta data from
+```
+{meta data}
+```
+to
+```
+user1 {meta data}
+user2 {meta data}
+user3 {meta data}
+...
+```
+
+### Compatibility
+
+When upgrade from old version, all legacy meta data belongs to the default meta data space.
+```
+{meta data}
+```
+```
+default {meta data}
+user1 {meta data}
+user2 {meta data}
+...
+```
+
+## Alternatives
+
+In the multi-tenant scenario, tenant can add a {tenant-id} prefix for each data key, but tenant-id
+is a meta data esstionally, each data key has a tenant-id prefix may cost more disk space & memory
+in TiKV.


text/0083-multi-level-meta-data-space.md
```diff
@@ -0,0 +1,52 @@
...
+```
+default {meta data}
+user1 {meta data}
+user2 {meta data}
+...
+```
+
+## Alternatives
+
+In the multi-tenant scenario, tenant can add a {tenant-id} prefix for each data key, but tenant-id
```
nolouch, line 50: Is the tenant only for TiKV users or also for TiDB users?

text/0083-multi-level-meta-data-space.md
```diff
@@ -0,0 +1,52 @@
...
+PD is the mata data center of TiKV Cluster, the meta data contains TiKV instances' addr, user data range(region)
+information, the number of replicas of data, etc.
+
+But PD currently only can handle the key space [min-key, max-key], it means that PD only can handle one user's 
+meta data, it is not possible when multiple users want share the same PD cluster as their meta system. There are
+two different scenarios that require PD can handle multiple user' meta data:
+
+1. Multiple TiKV Cluster share the same PD cluster. Because the minimal demplyment of a TiKV Cluster is 3 TiKV 3 PD,
+but it is not cost-effect if every small cluster has 3 dedicated meta data node.
+2. There are Multiple tenant in the same TiKV Cluster, each tenant has it own meta data, each tenant's key range can
```
nolouch, line 18: Is the [keyspace](https://github.com/tikv/rfcs/blob/7311fc02b21d3e5dc10165192f60cdc4610e98a3/text/0069-api-v2.md#keyspace-management) in API v2 match this?

text/0083-multi-level-meta-data-space.md
```diff
@@ -0,0 +1,52 @@
...
+default {meta data}
+user1 {meta data}
+user2 {meta data}
+...
+```
+
+## Alternatives
+
+In the multi-tenant scenario, tenant can add a {tenant-id} prefix for each data key, but tenant-id
+is a meta data esstionally, each data key has a tenant-id prefix may cost more disk space & memory
```
BusyJay, line 51: Any perf stats to show the cost?

text/0083-multi-level-meta-data-space.md
```diff
@@ -0,0 +1,52 @@
...
+information, the number of replicas of data, etc.
+
+But PD currently only can handle the key space [min-key, max-key], it means that PD only can handle one user's 
+meta data, it is not possible when multiple users want share the same PD cluster as their meta system. There are
+two different scenarios that require PD can handle multiple user' meta data:
+
+1. Multiple TiKV Cluster share the same PD cluster. Because the minimal demplyment of a TiKV Cluster is 3 TiKV 3 PD,
+but it is not cost-effect if every small cluster has 3 dedicated meta data node.
+2. There are Multiple tenant in the same TiKV Cluster, each tenant has it own meta data, each tenant's key range can
+contains any key in the range of [min-key, max-key].
```
BusyJay, line 19: To make it practical, every APIs need to be accept a `user` prefix. And each users' data can't be stored in the same rocksdb obviously. This also requires PD have knowledge about the underlying storage engine and avoid scheduling replicas from different users to the same storage engine. And TiKV needs to split all memory meta to different users. For example, the index of range becomes `HashMap<UserKey, BTreeMap<Vec<u8>, u64>>`.

In my opinion, using prefix is more straightforward and simpler.

text/0083-multi-level-meta-data-space.md
zhangjinpeng1987: both.

text/0083-multi-level-meta-data-space.md
zhangjinpeng1987: No, v2 API can not satisfy multiple TiDB tenants.

zhangjinpeng1987: When there are multiple TiDB tenants, each TiDB should has its own `ddl-owner`, `gc-safepoint` and other meta data, these meta data should be stored in PD separately.

text/0083-multi-level-meta-data-space.md
BusyJay: Using prefix can also achieve the same improvement. The difference of using prefix and using a separate explicit meta is that PD/TiKV/TiDB needs to take good care about meta in the later case.

text/0083-multi-level-meta-data-space.md
zhangjinpeng1987: The insert QPS of having prefix has 4% regression compare with no prefix.

zhangjinpeng1987: The bigger key size will consume more raftlog or wal and comparing CPU.

BusyJay: What prefix is used for testing? Note a two byte prefix can support 32768 tenants already.

zhangjinpeng1987:
Another using scenario: multiple tidb cluster share the same pd cluster to reduce the overhead of PD in the TiDB Cloud.

PR #90, opened 2022-03-22
RFC:TiKV RawKV MVCC GC

ystaticy:
## Summary
Move TiKV MVCC GC worker from TiDB into a group of independent GC worker node role and implement a new GC process in TiKV for RawKV.

## Motivation
GC worker is an important component for TiKV that deletes outdated MVCC data so as to not explode the storage. But currently, the GC worker is implemented in TiDB, which makes TiKV not usable without TiDB.And current GC process is just for transaction of TiDB,it's not usable for RawKV.


text/0090-tikv-gc.md
```diff
@@ -0,0 +1,44 @@
...
+In a TiKV cluster without TiDB nodes , there are a few different points as follows:
+1. We need to move GC worker into another node role.
+2. For [API V2](https://github.com/tikv/rfcs/blob/master/text/0069-api-v2.md) .It need gc the earlier version in default cf. But Txn GC worker process will be triggered by WriteCompactionFilter of write cf.
+3. RawKV encoded code of RawValue is different with Txn in TiDB.
+
+So we designed a new GC architecture and process for TiKV cluster.
+
+## Detailed design
+For support TiKV cluster deploy without TiDB nodes.
+1. Add a new node role instead of GC worker in TiDB nodes.
```
andylokandy, line 36: We need to decide exactly which role before merging the RFC.

text/0090-tikv-gc.md
```diff
@@ -0,0 +1,44 @@
...
+   The code of new GC worker,It will be added into [tikv/migration](https://github.com/tikv/migration)
+2. And for API V2, we need add new CompactionFilter which is named RawGCcompactionFilter, and add a new GCTask type implementation.
```
andylokandy, line 38: There is already a CompactionFilter for RawKV. Are we going to add a new one or just modify it?

text/0090-tikv-gc.md
```diff
@@ -0,0 +1,44 @@
...
+3. GC conditions in RawGCcompactionFilter is:  (ts < GCSafePoint) && ( ttl-expired || deleted-mark || not the newest version ).
```
andylokandy, line 39: Need to clarify the `GCSafePoint`.

andylokandy: We may want to remove a key that `ts > GCSafePoint` but `ttl-expired = true`. @pingyu What do you think?

pingyu: We should keep such entries. Expired entries are just the same as logical deleted. Miss to capture these entries will cause downstream to return an earlier version, and being inconsistent to upstream.

andylokandy: @pingyu So there will be this case: the CDC may sync a key that is already expired with downstream, right?

BusyJay:
The RFC is quite simple yet. For example, how is the new node designed? How it triggers GC? How does GC work in details? What's the triggering frequency?

text/0090-tikv-gc.md
```diff
@@ -0,0 +1,44 @@
...
+## Summary
+Move TiKV MVCC GC worker from TiDB into a group of independent GC worker node role and implement a new GC process in TiKV for RawKV.
+
+## Motivation
+GC worker is an important component for TiKV that deletes outdated MVCC data so as to not explode the storage. But currently, the GC worker is implemented in TiDB, which makes TiKV not usable without TiDB.And current GC process is just for transaction of TiDB,it's not usable for RawKV.
+
+## Background
+According to the documentation for the current GC worker in a TiDB cluster, the GC process is as follows:
+
+In TiDB GC worker leader:
```
nolouch, line 13: Currently, TiDB GC worker is not standardized use `client-go`. I think there are some problems that need to be solved. details: https://docs.google.com/document/d/1jA3lK9QbYlwsvn67wGsSuusD1Dzx7ANq_vya384RBIg/edit#

text/0090-tikv-gc.md
ystaticy: TTLCompactionFilter is just for APIV1TTL,In API V2 ,There are several differences with APIV1TTL.I think add a new CompactionFilter is better, @pingyu What do you think?

text/0090-tikv-gc.md
ystaticy: (2) Who will push GCSafePoint forward?
the new node role will instead of GC Worker in TiDB, It will push GCSafePoint forward.

ystaticy:
> 

OK,I will add more information in the document.

text/0090-tikv-gc.md
ystaticy: I have read this document and discussed with @MyonKeminta .For implement this proposal, it need a clearer design and more development.
But due to scheduling reasons, we hope to make changes to GC as small as possible to support TiKV API V2.
So I submit this RFC.

ystaticy:
@BusyJay  OK,I'll add more details for this RFC

text/0090-tikv-gc.md
pingyu: > @pingyu So there will be this case: the CDC may sync a key that is already expired with downstream, right?

Yes.

text/0090-tikv-gc.md
```diff
@@ -0,0 +1,88 @@
...
+2. For [API V2](https://github.com/tikv/rfcs/blob/master/text/0069-api-v2.md) .It need gc the earlier version in default cf. But Txn GC worker process will be triggered by WriteCompactionFilter of write cf.
+3. RawKV encoded code is different with Txn data in TiDB.
+
+So we designed a new GC architecture and process for TiKV cluster.It will be extended on the original interface to support the RawKV MVCC GC. For the original TiDB scenario, the old GC implementation can be used first.
+
+## Detailed design
+For support RawKV GC in TiKV cluster deploy without TiDB nodes.
+1. Add a new node role instead of GC worker in TiDB nodes.
+- Why we choose to create a new node role:
+  - IF we add GC Worker in PD: It will cause the problem of client-go circular dependency.
```
pingyu, line 49: I think the main reason is that PD is designed for scheduling & meta storage. Acting as GC worker will bring inappropriate feature and lower availability of PD.

text/0090-tikv-gc.md
```diff
@@ -0,0 +1,88 @@
+# RFC: TiKV RawKV MVCC GC
+
+
+## Summary
+Move TiKV MVCC GC worker from TiDB into a group of independent GC worker node role and implement a new GC process in TiKV for RawKV.
```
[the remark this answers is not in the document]
pingyu: "node role" seems to be not easy to understand.
Suggest to use "component", as TiUP using this word. See [tiup.io](https://tiup.io/).

text/0090-tikv-gc.md
```diff
@@ -0,0 +1,88 @@
...
+  - It is implemented in golang, which is convenient to call the interface of client-go.
+
+  - The code of new GC worker, will be added into [tikv/migration](https://github.com/tikv/migration)
+   
+2. Changes on PD:
+- A new concept is 'service group':  
+  - Due to TiDB, TxnKV and RawKV are allowed to coexist. Because the data of the three scenarios are independent, Because the data of the three scenarios are independent, separate safepoints are used in the GC, which helps to reduce the interference between businesses and speed up the GC.
+  - If multi tenancy is supported in the future, 'service group' can also support it.
+  - Need to design new interfaces for update service safepoint with 'service group'.
+  - Add UpdateServiceGCSafepointByServiceGroup and getGCSafepointByServiceGroup to standardize the API.
```
pingyu, line 63: Please also specify the arguments and returns, as well as the main logics of the interface.

The naming is a little too long. "UpdateServiceGCSafepointByGroup" may be better.

How to allocate `$service_group_id` ?

What's the path of GC safepoint ?

Will there be new interface & ETCD path for Txn safepoint ? TxnKV will need a Txn safepoint.

text/0090-tikv-gc.md
```diff
@@ -0,0 +1,88 @@
...
+- service safe point generated by CDC,BR or Lighting:  
+     ```shell
+  /gc/safe_point/service/$serviceID
+    ```
+
+In every TiKV nodes：
+1. Get GC safe point from PD regularly.
+2. Deletion will be triggered in CompactionFilter and GcTask thread;
+   
+## New GC worker architecture
```
pingyu, line 37: Maybe it is not a new architecture. We are utilizing the same architecture but another implementation.

text/0090-tikv-gc.md
```diff
@@ -0,0 +1,88 @@
...
+In a TiKV cluster without TiDB nodes , there are a few different points as follows:
+1. We need to move GC worker into another node role.
+2. For [API V2](https://github.com/tikv/rfcs/blob/master/text/0069-api-v2.md) .It need gc the earlier version in default cf. But Txn GC worker process will be triggered by WriteCompactionFilter of write cf.
```
pingyu, line 40:
```suggestion
2. For [API V2](https://github.com/tikv/rfcs/blob/master/text/0069-api-v2.md), it need gc the earlier version in default cf. But Txn GC worker process will be triggered by WriteCompactionFilter of write cf.
```

text/0090-tikv-gc.md
```diff
@@ -0,0 +1,156 @@
...
+                 message UpdateGCSafepointByServiceGroupResponse {
+                   ResponseHeader header = 1;
+                   bytes service_group_id = 2;  
+                   uint64 new_safe_point = 3;
+                 }
+               ```
+         2. used to get GC safepoint for TiKV:
+            1. interface:
+               ```shell  
+               func (s *GrpcServer) GetAllServiceGroupGcSafePoint(ctx context.Context, request *pdpb.GetServiceGroupServiceGcSafeRequest)  (*pdpb.UpdateServiceGCSafePointResponse, error)
```
AmoebaProtozoa, line 108: Should `Gc` here be capitalized to stay consistent with other interfaces?

text/0090-tikv-gc.md
```diff
@@ -0,0 +1,188 @@
...
+We change RawKV encoding to MVCC, so the GC is necessary.
+No GC for TxnKV scenario when TiDB is not deployed.
+The GC safe point & Txn safe point is easy to be misunderstand.
+
+## Background
+According to the documentation for the current GC worker in a TiDB cluster, the GC process is as follows:
+
+In TiDB GC worker leader:
+1. Regularly calculates a new timestamp called "GC safe point"(The default interval is 10min), and push the safe point to PD.
+2. Get the minimal Service safe point among all services from the response of step 1, which is GC safe point .
```
sunxiaoguang, line 20:
```suggestion
2. Get the minimal service safe point among all services from the response of step 1, which is GC safe point .
```

text/0090-tikv-gc.md
```diff
@@ -0,0 +1,188 @@
...
+     ```shell
+  /gc/safe_point/service/$serviceID
+    ```
+
+In every TiKV nodes：
+1. Get GC safe point from PD regularly.
+2. Deletion will be triggered in CompactionFilter and GcTask thread;
+   
+## New GC worker implementation
+In a TiKV cluster without TiDB nodes , there are a few different points as follows:
```
sunxiaoguang, line 44:
```suggestion
In a TiKV cluster without TiDB nodes, there are a few different points as follows:
```

text/0090-tikv-gc.md
```diff
@@ -0,0 +1,188 @@
...
+               ```proto  
+                   rpc UpdateServiceSafePointByServiceGroup(UpdateServiceSafePointByServiceGroupRequest) returns (UpdateServiceSafePointByServiceGroupResponse) {}
+               ```
+            2. PB messages:
+               ```proto
+                    message UpdateServiceSafePointByServiceGroupRequest {
+                        RequestHeader header = 1;
+                        bytes service_group_id = 2;
+                        bytes service_id = 3;
+                        int64 TTL = 4;
```
sunxiaoguang, line 133: TTL doesn't follow field name standard. If `ttl` generates strange field accessor, we can change it to xxx_ttl  or maybe time_to_live. They both look better and consistent.

text/0090-tikv-gc.md
```diff
@@ -0,0 +1,188 @@
...
+     - The code of new GC worker, will be added into [tikv/migration](https://github.com/tikv/migration)
+     - the GC Worker configuration in config/gc_worker_conf.toml file.
+     - The default interval for generating GC safepoint is still '10m0s'.
+     - And TiKV will get the GC safe point from PD. GC safe point = min(all service safe point, gc worker safe point).
+
+   
+2. Changes on PD:
+   1. A new concept is 'service group':  
+        - Due to TiDB, TxnKV and RawKV are allowed to coexist. Because the data of the three scenarios are independent, Because the data of the three scenarios are independent, separate safepoints are used in the GC, which helps to reduce the interference between businesses and speed up the GC.
+        - If multi tenancy is supported in the future, 'service group' can also support it.
```
sunxiaoguang, line 70: Where do we put start and end key for a `service group`?

PR #94, opened 2022-06-02
remove KV WAL

BusyJay:


text/0094-remove-kv-wal.md
```diff
@@ -0,0 +1,107 @@
...
+
+We already have all the raft logs required for recovery, it's a waste to write everything to WAL again. Although we are using cache write almost the whole time, but if a page is marked to flush, it can still trigger random latency. And if we want to use multiple rocksdb instances in one node, writing WAL can introduce a lot of random writes.
+
+## Detailed design
+
+This RFC assumes you have know how RocksDB works and recovers from restarting.
+
+### Relation between sequence number and (region ID, apply index)
+
+In RocksDB, WAL plays the role of recording both data and order. MANIFEST plays the role of recording persistent data. Now that WAL is removed, we need a new component to replace its role. Raft logs already have all the necessary data and partial order within the same peer, and the order across peers is traced by region version. So all we need to do is record the exact index KV DB belongs to.
```
Connor1996, line 20: Should elaborate on the issue why need to record this. One can simply think just replaying from the previous apply index to the commit index is enough, without realizing the apply index is stored in RAFT CF of kvdb and different CFs are flushed separately.

text/0094-remove-kv-wal.md
```diff
@@ -0,0 +1,107 @@
...
+
+The RFC proposes to build a relation between sequence number (SN) and (region ID, apply index), so apply index can be queried by querying each CF’s latest persisted SN.
+
+To build relations between SN and apply index, we need to know what SN is assigned to each write. RocksDB doesn’t have a public API to provide the information. Fortunately, exposing it is just a work of several lines.
+
+To persist the relations, we need to send them back to the raftstore thread, along with `ApplyRes`. Noticing SN is strictly monotonically increased just like log index, we can store the relations just like raft entries.
+
+Because we need to use the relation to recover writes, so before a rocksdb flush finishes, we need to ensure the corresponding relation is persisted. When using separate RocksDB for each region, this is simple as we only care about one region at one flush. So just waiting for the SN to persist is enough. When sharing RocksDB for all regions, we need to ensure SN is persisted for all affected regions.
```
Connor1996, line 32: Why do we need to consider sharing RocksDB for regions? I think we can just deliver it after using separate RocksDB

text/0094-remove-kv-wal.md
```diff
@@ -0,0 +1,107 @@
...
+
+### Reduce relation count
+
+KV RocksDB write rates can be more than 10kops, saving all relations can be a huge cost. The point of relation is to record the order of writes and get a suitable replay start point. It doesn’t require all relations to serve the goal.
+
+To get a suitable replay start point, only the maximum flushed SN needs to be considered. And only frozen memtables will be flushed, so every time a memtable is frozen, we can hint raftstore to the maximum SN this memtables contains. And before flushing the memtable, we wait for the SN relation to be persisted. There are existing hooks, EventListener, allowing us to make the modifications.
+
+For all other relations, they can be merged together and only keep the largest SN and largest apply index for each region ID.
+
+### How to recover from relation
```
Connor1996, line 44: let call it `replay` instead of `recover`

text/0094-remove-kv-wal.md
```diff
@@ -0,0 +1,107 @@
...
+
+After TiKV is restarted, it should check the maximum SN number for each CF, and choose the minimum SN as the replay start point. Similar to RocksDB, writes can be ignored partially if they contain the data to a CF with larger SN.
+
+Recovery should always recover the region with the smallest version. If the recovery changes the version, it should pause the current region and choose the new smallest version again. Region with the same version can be recovered with arbitrary order. The consideration is that version is the sync point between regions, operations such as split, merge, to smaller regions must happen earlier when their range overlap.
```
Connor1996, line 48:
```suggestion
Recovery should always recover the region with the smallest version. If the recovery is about to change the version, it should pause the current region and choose the new smallest version again. Region with the same version can be recovered with arbitrary order. The consideration is that version is the sync point between regions, operations such as split, merge, to smaller regions must happen earlier when their range overlap.
```

text/0094-remove-kv-wal.md
```diff
@@ -0,0 +1,107 @@
...
+
+### Reduce relation count
+
+KV RocksDB write rates can be more than 10kops, saving all relations can be a huge cost. The point of relation is to record the order of writes and get a suitable replay start point. It doesn’t require all relations to serve the goal.
+
+To get a suitable replay start point, only the maximum flushed SN needs to be considered. And only frozen memtables will be flushed, so every time a memtable is frozen, we can hint raftstore to the maximum SN this memtables contains. And before flushing the memtable, we wait for the SN relation to be persisted. There are existing hooks, EventListener, allowing us to make the modifications.
```
Connor1996, line 40: `And before flushing the memtable, we wait for the SN relation to be persisted.` It's not atomic, what if it restarts after the SN relation is persisted but the memtable hasn't been finished flushing?

text/0094-remove-kv-wal.md
```diff
@@ -0,0 +1,107 @@
...
+
+One choice is enabling atomic flush and using raft CF to record the index, but it can bring too many small compactions. It can get worse when there are many rocksdb instances and flush is invoked frequently.
+
+Another choice is storing the apply index in properties. However, on one hand, it’s not guaranteed that every time a flush will generate at least a table file. On the other hand, if all data are deleted, all files can be removed. So properties are unreliable (and verbose).
+
+The RFC proposes to build a relation between sequence number (SN) and (region ID, apply index), so apply index can be queried by querying each CF’s latest persisted SN.
+
+To build relations between SN and apply index, we need to know what SN is assigned to each write. RocksDB doesn’t have a public API to provide the information. Fortunately, exposing it is just a work of several lines.
+
+To persist the relations, we need to send them back to the raftstore thread, along with `ApplyRes`. Noticing SN is strictly monotonically increased just like log index, we can store the relations just like raft entries.
```
Connor1996, line 30: How is the relation persisted, I mean, the detailed format

text/0094-remove-kv-wal.md
BusyJay: The new architecture will not be stable for quite some time.

text/0094-remove-kv-wal.md
BusyJay: Then it will replay from a smaller apply index.

text/0094-remove-kv-wal.md
BusyJay: I think it's an implementation details. It can be changed to best suit different engine types.

PR #95, opened 2022-06-05
introduce latest cf

BusyJay:


text/0095-add-latest-cf.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Detailed design
+
+TiKV doesn't need to know all existing versions of keys. In fact, most of the time, v0 is larger than any existing versions of keys in TiKV if there is more read than write. So it should be enough to just let TiKV knows the latest version of all keys.
+
+The RFC propose to add a new cf named "latest". When a key is inserted using transaction API, it should update write cf (and default cf) as current does. In addition, it also insert a key to latest cf with
+- key set the original key without any encoding or version
+- value set the same value in write cf but include the corresponding version.
+
+For example, insert k1 with version v0 and value dummy will insert two keys
```
zhangjinpeng1987, line 34: This will double the write amplification, not only by size but also by key value pairs.

text/0095-add-latest-cf.md
```diff
@@ -0,0 +1,72 @@
...
+
+All of the fallback queries should be performed lazily.
+
+The improvment should be very significant when update keys frequently.
+
+### Compatibility
+
+Because all existing cfs are updated just as before, so there are no major compatibility issues.
+
+But using latest cf should be triggered explicitly by client. Client should ensure only when it updates all keys with latest cf will it ask TiKV to query using latest cf.
```
zhangjinpeng1987, line 54: It is a bit of complicated.

text/0095-add-latest-cf.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Drawbacks
+
+It introduces a write amplification appearantly. That is also the reason why the access pattern needs to be controlled by client. Client should enable latest cf only when it knows a range of keys are updated very often and can be benificial from the change.
+
+On the other hand, the additional write is just a key in a different cf and a value that is probably not larger than 255 bytes, the overhead may not be very signifiant. More experiments are needed.
+
+## Alternatives
+
+unistore separates the latest version and other versions by adjust file format. So when flushing or compacting, it will make latest versions key be the first part, and the rest in the second part. This approach doesn't have write overhead, but is not backward compatible in TiKV.
```
zhangjinpeng1987, line 70: How about create a history cf to store as many history version as it can, the write cf store the latest several versions, and background thread like GC thread move existed history versions to history cf.
For point get, using RocksDB's user timestamp feature to get the latest version can see, this also can prevent creating iterator.

zhangjinpeng1987: There is no compatibility issue for this solution.

BusyJay: Compaction filter is not reliable, we have seen issue cause by not having compaction in time like https://github.com/tikv/tikv/issues/12729. On the other hand, this proposal doesn't depend on the user timestamp, which is not used in production yet.

text/0095-add-latest-cf.md
BusyJay: Yes, it's addressed in the drawback section.

text/0095-add-latest-cf.md
BusyJay: Usually it won't as the client already needs to manage different ranges, just like table in TiDB or key space.

text/0095-add-latest-cf.md
BusyJay: Added as an alternative.

zhangjinpeng1987: Using history cf doesn't need the compaction filter, gc just check the history cf sst file's max ts to determine wether to drop the whole sst file.

BusyJay: Compaction filter is not used for dropping history sst, but for moving data from write cf to history cf.

zhangjinpeng1987: Compaction filter is not necessary, I think version statistics-based GC is much better.

text/0095-add-latest-cf.md
```diff
@@ -0,0 +1,80 @@
...
+
+. Version is a 64 bits number and encoded in desc order (larger goes first).
+
+When reading a key with version v0, TiKV is expected to return the largest version v1 that v1 <= v0. Because TiKV has no idea what version is available, so it has to create an iterator and use version v0 to encode a seek key. If any key is hit, then the key should be the requested key.
+
+The procedure is straightforward but expensive. Creating an iterator in RocksDB is not free. Even for point get, an iterator is still necessary. And seek operation is also expensive, almost as expendsive as creating iterator. To avoid seeking too many times, we introduce `near_seek` to TiKV in the early days, which tries `next` several times before fallback to `seek`.
+
+As explained, the reason why seek is necessary is because TiKV has no idea what versions are available. Otherwise it can just use `get` to query the specific key, which is a lot cheaper.
+
+## Detailed design
```
zhangjinpeng1987, line 26: How about lightning, cdc, br, tiflash's compatibility?

BusyJay: They are not part of TiKV. I would like to talk about them in another doc.

text/0095-add-latest-cf.md
```diff
@@ -0,0 +1,80 @@
...
+
+TiKV doesn't need to know all existing versions of keys. In fact, most of the time, v0 is larger than any existing versions of keys in TiKV if there is more read than write. So it should be enough to just let TiKV knows the latest version of all keys.
+
+The RFC propose to add a new cf named "latest". When a key is inserted using transaction API, it should update write cf (and default cf) as current does. In addition, it also insert a key to latest cf with
```
zhangjinpeng1987, line 30: this double write may also double the total disk size in some scenarios.

zhangjinpeng1987: Maybe for those new table(new range) in the existing cluster and the new created cluster we can use the new mechanism(only write latest cf, move the old version to write cf only when the older version existed) is a better choice?

BusyJay: It would work, though it loses the ability to upgrade existing table to new format online.

BusyJay: Or we can introduce 3 formats. One is the original one, second is double write, third is move on update. They can be upgraded online one by one. For new tables, third format is used by default.

BusyJay:
Although I updated the RFC to write latest CF first, and then move keys between CFs, I suggest to always double write in POC as it has less compatibility issues.

text/0095-add-latest-cf.md
```diff
@@ -0,0 +1,92 @@
...
+## Detailed design
+
+TiKV doesn't need to know all existing versions of keys. In fact, most of the time, v0 is larger than any existing versions of keys in TiKV if there is more read than write. So it should be enough to just let TiKV knows the latest version of all keys.
+
+The RFC propose to add a new cf named "latest". When a key is inserted using transaction API, it should update latest cf using the original key without any encoding or version. The value should be similar with the one that is used for updating write cf in the past, but including the corresponding version. If the key exists in the latest cf, then its value should be read and write to write cf using the old format along side with the latest cf update.
+
+For example, supposing there is no key in latest cf. Inserting k1 with version v0 and value foo will insert one key:
+- to latest cf, k1 -> (foo, v0 and other meta)
+
+Inserting v1 again with version v1 and value bar will insert two keys:
```
ekexium, line 35:
```suggestion
Inserting k1 again with version v1 and value bar will insert two keys:
```
Is it k1?

BusyJay: Yes.

PR #96, opened 2022-06-07
Stale read for rawkv with read ts

iosmanthus:
Signed-off-by: iosmanthus <myosmanthustree@gmail.com>

This pull request is based on the https://github.com/tikv/rfcs/pull/80.


text/0096-rawkv-stale-read.md
```diff
@@ -0,0 +1,97 @@
...
+
+Allow eventually consistency on RawKV. This proposal is based on the https://github.com/andylokandy/rfcs/blob/staleread/text/0080-rawkv-stale-read.md. Thanks, @andylokandy's work!.
+
+## Motivation
+
+Eventually consistency is acceptable in some scenarios when using RawKV. In this case, RawKV is allowed to break linear consistency and therefore improve the cluster-wise throughput and latency.
+
+TiKV currently supports **three** features to process read-only queries more efficiently. 
+
+1. Follower read.
```
BusyJay, line 13: You mean replica read?

text/0096-rawkv-stale-read.md
```diff
@@ -0,0 +1,97 @@
...
+```diff
+message RawGetRequest {
+    Context context = 1;
+    bytes key = 2;
+    string cf = 3;
++   uint64 read_ts = 4;
+}
+```
+
+2. While TiKV is handling radw read-related requests, construct a `SnapContext` with the `read_ts` before acquiring a snapshot from `storage`.
```
BusyJay, line 78:
```suggestion
2. While TiKV is handling raw read-related requests, construct a `SnapContext` with the `read_ts` before acquiring a snapshot from `storage`.
```

text/0096-rawkv-stale-read.md
```diff
@@ -0,0 +1,97 @@
...
+      ..Default::default()
+  };
+}
+```
+
+2. Client API interface; replica selection algorithm.
+
+```diff
+class RawKVClient {
++     ByteString rawGet(ByteString key, readTs: Timestamp)
```
BusyJay, line 95: I think it might be confusing for client to understand what is readTs in RawKV.

text/0096-rawkv-stale-read.md
```diff
@@ -0,0 +1,97 @@
...
+</p>
+
+
+## Detailed Design
+
+This RFC proposes stale read on RawKV which allows reading stale data from the not-leader replica with a lossy time-bound. Unlike the stale read on TxnKV, the stale read on RawKV breaks its original linear consistency guarantee -- downgrades to eventually consistency. TiKV only guarantees that the stale data read from followers are written before.
+
+### TiKV
+
+ While trying to read data, clients should specify a timestamp which attachs to the request header as `read_ts`, typically a timestamp few seconds ago. The replica should read the local storage with the `read_ts` and reuses the mechanism from the stale read of TxnKV. This requires the replica to check the `read_ts` against the `safe_ts` which is advaneced by `CheckLeader` message from the store of the leader or `resolve-ts` worker. As long as the `safe_ts` is no less than `read_ts`, the replica is allowed to read the key from local storage.
```
BusyJay, line 54: How is `safe_ts` maintained in RawKV since there is no locks?

text/0096-rawkv-stale-read.md
iosmanthus: Yes.

text/0096-rawkv-stale-read.md
iosmanthus: If there are no locks, the `resolve_ts` will advance the `safe_ts` by requesting the TSO for a timestamp periodically. The default config for the `resolve_ts` worker is `1s`.

iosmanthus: More details are supplemented.

text/0096-rawkv-stale-read.md
iosmanthus: How about changing the time type to `DataTime` instead of using `TimeStamp`.

text/0096-rawkv-stale-read.md
BusyJay: Then correct name should be used.

text/0096-rawkv-stale-read.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Motivation
+
+Eventually consistency is acceptable in some scenarios when using RawKV. In this case, RawKV is allowed to break linear consistency and therefore improve the cluster-wise throughput and latency.
+
+TiKV currently supports **three** features to process read-only queries more efficiently. 
+
+1. Follower read.
+
+    Follower read allows reading from the followers. Without breaking the linear consistency guarantee, the follower will send a read-index request to the leader. The leader will not respond with the actual value, instead, send a round of heartbeats to confirm its leadership and calculate the largest commit index (read index) across the cluster for the follower. After the follower advances its apply index to the read index, it is safe to get data from the local storage and respond to it to the client. This feature helps distribute the read stress on the leader but still increases the read latency.
```
BusyJay, line 15:
```suggestion
    Follower read allows reading from the followers. Without breaking the linear consistency guarantee, the follower will send a read-index request to the leader. The leader will not respond with the actual value, instead, send a round of heartbeats to confirm its leadership and calculate the largest commit index (read index) across the cluster for the follower. After the follower advances its apply index to the read index, it is safe to get data from the local storage and respond to the client. This feature helps distribute the read stress on the leader but still increases the read latency. 
```

text/0096-rawkv-stale-read.md
```diff
@@ -0,0 +1,97 @@
...
+
+### Client
+
+The `read_ts` specified by the client could be acquired by the following ways:
+
+1. Calculate a timestamp from the physical time from the local. The `read_ts` might suffer from the clock drift and exceed the max timestamp allocated from TSO. The client will fail to read any data even if that target replica is the leader since the `safe_ts` of the replica don't catch up with the `read_ts`. **Deploying NTP services** in the cluster might mitigate this issue.
```
BusyJay, line 60: Can it be 0?

Connor1996, line 60: if `read_ts` exceeds the max timestamp allocated from TSO, maybe we can just return the latest data instead of no data.

text/0096-rawkv-stale-read.md
```diff
@@ -0,0 +1,97 @@
...
+</p>
+
+
+## Detailed Design
+
+This RFC proposes stale read on RawKV which allows reading stale data from the not-leader replica with a lossy time-bound. Unlike the stale read on TxnKV, the stale read on RawKV breaks its original linear consistency guarantee -- downgrades to eventually consistency. TiKV only guarantees that the stale data read from followers are written before.
+
+### TiKV
+
+ While trying to read data, clients should specify a timestamp which attachs to the request header as `read_ts`, typically a timestamp few seconds ago. The replica should read the local storage with the `read_ts` and reuses the mechanism from the stale read of TxnKV. This requires the replica to check the `read_ts` against the `safe_ts` which is advaneced by `CheckLeader` message from the store of the leader (for follower) or `resolve-ts` worker (for leader). As long as the `safe_ts` is no less than `read_ts`, the replica is allowed to read the key from local storage. Notice that there is no lock for the RawKV regions, thus the `resolve-ts` worker advanced the `safe_ts` by requesting the TSO for the latest timestamp.
```
pingyu, line 54: 1. What's the meaning of `safe_ts` for RawKV ? Suggest to give a definition, e.g, "the minimum timestamp of the on-the-fly RawKV writes".

2. If the definition depends on "timestamp" of RawKV writes, this feature depends on the timestamp introduced by API V2, is that right ?

3. The mechanism to get the "minimal timestamp" of the on-the-fly writes between Txn & Raw would be quite different. Although there is no locks, Raw writes would still be "on-the-fly" during Raft procedure.

4. RawKV CDC faces the a very similar problem to track "on-the-fly" for [`resolved-ts`](https://github.com/pingyu/tikv-rfcs/blob/rawkv-cdc/text/0083-rawkv-cross-cluster-replication.md#23-resolved-timestamp). I think we can reuse it for stale read. Please refer to #86 .

text/0096-rawkv-stale-read.md
iosmanthus: Then the `read_ts` will lost its restriction to the data freshness since some very stale replicas might be chosen.

iosmanthus: We can preserve this value for the unbound stable read: read the latest data without checking `safe_ts`.

BusyJay: Then how to preserve compatibility?

BusyJay:
There is a special case that user may choose availability rather than consistency. So client is OK to read with any ts. In this RFC, it seems keys with larger ts may be skipped during read.

iosmanthus:
This RFC doesn't depend on the keys' timestamp, the underlying storage could have no information about the timestamp. The `safe_ts` is like the (approximate) timestamp of the leader that writes the key. The `read_ts` tends to be checked against the `safe_ts` to guarantee the replicas have already been synced with those writes that happened around `safe_ts`. To read any data, we could specify the `read_ts` to 0, and then any replicas with `safe_ts` > 0 could handle the request.

BusyJay:
I'm OK with 0 timestamp. Currently, txn stale read consider ts 0 an error. And client (like TiDB) may actually send ts 0 by mistake. This RFC should state clear what 0 means in rawkv and implementation should not break compatibility.

PR #98, opened 2022-07-19
RFC: Follower replication

LintianShi:
A design of follower replication


text/0098-follower-replication.md
```diff
@@ -0,0 +1,246 @@
...
+We design a group broadcast schema which is proposed by this [rfc](https://github.com/tikv/rfcs/blob/362865abe36b5771d525310b449c07a57da7ef85/text/2019-11-13-follower-replication.md).
+
+The main idea is that the leader only sends `MsgBroadcast` to the agent of a certain AZ. `MsgBroadcast` contains log entries that the leader replicates to the agent, and several ranges of log that the leader replicates to other followers and learners.
+Once the agent receives `MsgBroadcast`, it appends log entries in the message and assembles `MsgAppend` according to the ranges in `MsgBroadcast` with its own log. Then the agent sends `MsgAppend` to other followers/learners in the same AZ. Thus, the leader can avoid sending `MsgAppend` across AZs.
+
+![main_idea](../media/follower-replication-2.jpg)
+
+### Preparation
+
+Every peer must have a field indicating which AZ it belongs to. This field does not have to be persistent. As mentioned in the comment, TiKV has the knowledge which AZ the store locates. We need to design the interface so that `RawNode` in raft-rs can get the AZ information in TiKV. Maybe we can initialize Peer with AZ information, and then Peer initializes the `RawNode`.
```
Connor1996, line 42: `We need to design the interface so that `RawNode` in raft-rs can get the AZ information in TiKV. Maybe we can initialize Peer with AZ information, and then Peer initializes the `RawNode`.`

This should be settled in the design.

LintianShi: I will take it into consideration.

text/0098-follower-replication.md
```diff
@@ -0,0 +1,248 @@
+# Follower Replication for Write-Flow
+
+<!-- - RFC PR: https://github.com/tikv/rfcs/pull/0000
```
BusyJay, line 3: Should be updated.

text/0098-follower-replication.md
```diff
@@ -0,0 +1,248 @@
...
+
+### Main idea
+
+We design a group broadcast schema which is proposed by this [rfc](https://github.com/tikv/rfcs/blob/362865abe36b5771d525310b449c07a57da7ef85/text/2019-11-13-follower-replication.md).
+
+The main idea is that the leader only sends `MsgBroadcast` to the agent of a certain AZ. `MsgBroadcast` contains log entries that the leader replicates to the agent, and several ranges of log that the leader replicates to other followers and learners.
+Once the agent receives `MsgBroadcast`, it appends log entries in the message and assembles `MsgAppend` according to the ranges in `MsgBroadcast` with its own log. Then the agent sends `MsgAppend` to other followers/learners in the same AZ. Thus, the leader can avoid sending `MsgAppend` across AZs.
```
BusyJay, line 36: Paragraph should be separate by one blank line.

BusyJay, line 36: This may be conflict with log async fetch.

text/0098-follower-replication.md
```diff
@@ -0,0 +1,248 @@
...
+
+![main_idea](../media/follower-replication-2.jpg)
+
+### Preparation
+
+Every peer must have a field indicating which AZ it belongs to. TiKV server has the knowledge which AZ each store locates because PD sends the global zone information to TiKV once the zone information is updated.
+
+In TiKV architecture, TiKV contains multiple peers that belong to different raft groups. So TiKV needs to update zone information to each peer.
+
+We add an extra field `peer_zone` in the peer, a hashmap that records store_id -> AZ. Every time the zone information stored in TiKV is updated, a peer message `UpdataZoneInfo` is generated. Then it will be broadcast to all peer in this TiKV server. When a peer receives `UpdataZoneInfo`, it will update its `peer_zone`.
```
BusyJay, line 46: Why not put it to `PollContext`?

LintianShi: I think the zone information only flows from TiKV server to peers, and peers never modify it. So it is read-only for peers. If the zone information is updated, Node can use a `PeerMsg` to update zone information stored in each `Peer`. `PollContext` is more like a layer where peers interact with TiKV server. If the zone information is stored in `PollContext` and shared with multiple peers, we need to consider the concurrency conflict. 

However, putting zone information in `PollContext` uses less memory.

LintianShi: I think about it again. I think zone information should be put on PollContext. Otherwise, we have to handle zone information migration when split or merge region.

PR #106, opened 2023-02-08
RFC: support witness(log only replica)

Connor1996:
[rendered](https://github.com/tikv/rfcs/blob/0745aa71d8c228bbe705b941306ae5de684ffd8d/text/0106-witness.md)


text/0106-witness.md
```diff
@@ -0,0 +1,191 @@
+# Witness
+
+## Background
+
+On cloud env, TiKV is equipped with AWS EBS storage for single node which provides durability of 99.9%. Moreover, TiKV organizes 3 Raft replicas on top of that, which may be a overkill for durability. To reduce cost, we can introduce a mechanism named "2 Replicas With 1 Log Only" instead of 3 replicas to save 30% storage resources and cut down about 100%(1 core) CPU usage empirically on the cloud. Compared with the original 3 replicas architecture, the "1 Log Only" replica stores raft logs but does not apply it, which still ensures data consistency through the Raft protocol.
```
ethercflow, line 5: [Our usage doc](https://github.com/pingcap/docs/pull/12403/files#diff-83ac4cde649c45f7e1fa659fc4c7dddab9eea56fb2115f4487ed801afbb2c30bR12) mentioned: Amazon Elastic Block Store (EBS) with 99.8%~99.9% and durability or Persistent Disk of Google Cloud Platform (GCP) with 99.99%~99.999% durability as the storage of each TiKV node.    Should we keep with it?

PR #108, opened 2023-05-23
RFC: reduce traffic for resolved ts

you06:


text/0000-reduce-traffic-resolved-ts.md
```diff
@@ -0,0 +1,84 @@
+## Summary
+
+The proposed method for reducing the traffic of pushing safe ts is hibernated region optimization. This method can be especially beneficial for large clusters since it can save a significant amount of traffic.
+
+The motivation behind this approach is that TiKV currently uses the `CheckLeader` request to push resolved ts, and the traffic it costs grows linearly with the number of regions. By utilizing hibernated region optimization, it is possible to reduce this traffic and improve overall performance.
```
Connor1996, line 5: But hibernated region is removed in raftstore-v2

you06: Yeah, maybe the traffic of pushing safe ts won't be a problem if the region size is dynamic. This RFC is only an optimization in raftstore-v1.

text/0108-reduce-traffic-resolved-ts.md
```diff
@@ -0,0 +1,114 @@
+## Summary
+
+The proposed method for reducing the traffic of pushing safe ts is hibernated region optimization. This method can be especially beneficial for large clusters since it can save a significant amount of traffic.
```
cfzjywxk, line 3: > The proposed method for reducing the traffic of pushing safe ts is hibernated region optimization.

Currently, the main strategy in the new design is to minimize leader validation traffic and  `safe-ts` pushing traffic for valid leaders.

text/0108-reduce-traffic-resolved-ts.md
```diff
@@ -0,0 +1,114 @@
...
+    uint64 ts = 1;
+}
+```
+
+### Active Regions
+
+The `CheckLeaderResponse` respond with the regions that pass the Raft safety check. The leader can then push its `safe_ts` for those regions. Since most regions will pass the safety check, it is not necessary to respond with the IDs of all passing regions. Instead, we can respond with only the IDs of regions that fail the safety check.
+
+Another optimization is that we can confirm the leadership if the leader lease is hold, by calling Raft's read-index command. But this will involve in the Raftstore thread pool, more CPU will be used by this.
+
```
cfzjywxk, line 73: We could add some explanations:
By default, the raft lease is set to `10s`. If the `advance-ts-interval` is configured to `100ms`, only one leader validation using the read index method will be required for 100 rounds which saves 99 rounds of unnecessary leader checks.

text/0108-reduce-traffic-resolved-ts.md
```diff
@@ -0,0 +1,114 @@
...
+### Inactive Regions
+
```
cfzjywxk, line 75: We need to explain how this new design interacts with the existing hibernated region feature and identify any dependencies it may have. Additionally, we should outline how to determine if a region is inactive.

As I understand it, the current workflow involves:

1. Checking if a region leader is active.
2. Validating that the leader role is still held for both active and inactive regions using different methods.
3. Sending pushing ts requests; note that different request types are required for active and inactive regions.

text/0108-reduce-traffic-resolved-ts.md
```diff
@@ -0,0 +1,114 @@
...
+message ApplySafeTsRequest {
+    uint64 ts = 1;
+    uint64 store_id = 2;
+    repeated uint64 unsafe_regions = 3;
+    repeated CheckedLeader checked_leaders = 4;
+}
+
+message ApplySafeTsResponse {
+    uint64 ts = 1;
```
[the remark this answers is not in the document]
you06: The `ts` is actually not used. 

> would we need ts for each active region in this case not just one ts?

No, because leader doesn not need to know the safe ts in followers.

text/0108-reduce-traffic-resolved-ts.md
```diff
@@ -0,0 +1,114 @@
...
+
+The design will be described in two sections: the improvement of active regions and a new mechanism for hibernated regions.
+
+### Protobuf
+
+We still need `CheckLeader` request to confirm the leadership. But with some additional fields.
+
+```diff
+message CheckLeaderRequest {
+    repeated LeaderInfo regions = 1;
```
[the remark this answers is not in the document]
you06: We send `LeaderInfo` without `ReadState` as a transition state from active to inactive(you can see "Transition State" section in the updates).
For active regions(which have continues write), they are not contains in `CheckLeaderRequest.regions`, because the leadership is checked by Raft lease, we send the **checked** state of them in `ApplySafeTsRequest.checked_leaders`.

text/0108-reduce-traffic-resolved-ts.md
```diff
@@ -0,0 +1,114 @@
...
++   repeated uint64 failed_regions = 3;
+}
+```
+
+To apply safe ts for valid leaders as soon as possible, instead of waiting for the next round of advancing resolved timestamps, we need to send another `ApplySafeTS` request. The `ApplySafeTS` request is usually small, the traffic caused by it can be ignored.
+
+```protobuf
+message CheckedLeader {
+    uint64 region_id = 1;
+    ReadState read_state = 2;
```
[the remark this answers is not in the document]
you06: > Do we need both `AppliedIndex` and `SafeTs` here?

Yes , it's contained in `ReadState`, because in active regions, the applied index in follower may delay from leader, we cannot apply the safe ts until the apply index catches up `CheckedLeader.read_state.applied_index`.

> Anyway to save space further, we can introduce `ReadStateDiff`

That's fun and I'll do some research of sending diff when we continue to push this project.

text/0108-reduce-traffic-resolved-ts.md
```diff
@@ -0,0 +1,130 @@
...
+}
+
+message CheckLeaderResponse {
+    repeated uint64 regions = 1;
+    uint64 ts = 2;
++   repeated uint64 failed_regions = 3;
+}
+```
+
+To apply safe ts for valid leaders as soon as possible, instead of waiting for the next round of advancing resolved timestamps, we need to send another `ApplySafeTS` request. The `ApplySafeTS` request is usually small, the traffic caused by it can be ignored.
```
overvenus, line 48: > The `ApplySafeTS` request is usually small

How small is it? Does it grow linearly with the number of regions?

text/0108-reduce-traffic-resolved-ts.md
```diff
@@ -0,0 +1,130 @@
...
+    uint64 region_id = 1;
+    ReadState read_state = 2;
+}
+
+message ApplySafeTsRequest {
+    uint64 ts = 1;
+    uint64 store_id = 2;
+    repeated uint64 unsafe_regions = 3;
+    repeated CheckedLeader checked_leaders = 4;
+}
```
overvenus, line 61: Could you add some comments about these fields?

text/0108-reduce-traffic-resolved-ts.md
```diff
@@ -0,0 +1,130 @@
...
+
+message ApplySafeTsResponse {
+    uint64 ts = 1;
+}
+```
+
+### Active Regions
+
+The `CheckLeaderResponse` respond with the regions that pass the Raft safety check. The leader can then push its `safe_ts` for those regions. Since most regions will pass the safety check, it is not necessary to respond with the IDs of all passing regions. Instead, we can respond with only the IDs of regions that fail the safety check.
```
overvenus, line 70: > Instead, we can respond with only the IDs of regions that fail the safety check.

What if messages get out of order? We may miss some failed regions.

text/0108-reduce-traffic-resolved-ts.md
you06: The size of `ApplySafeTS` request grows by the number of active regions whose the apply index is changed.

But we just move the leader info from `CheckLeader` request to `ApplySafeTS` request, even all regions are active, there won't be more traffic.

text/0108-reduce-traffic-resolved-ts.md
cfzjywxk: Yes as the validation and pushing are separated into two phases.

PR #109, opened 2024-01-03
RFC: large transaction

ekexium:


ekexium:
The design doc will be put in tidb repo

PR #110, opened 2024-01-06
RFC: Periodic full compaction (#110)

afeinberg:


text/0110-periodic-full-compaction.md
```diff
@@ -0,0 +1,317 @@
...
+**Periodic full compaction** is a scheduled task that starts at specified times
+of the day on each TiKV node and compacts the column families of all regions *at
+all levels including the bottommost (L6)*. In order to reduce to impact running
+such a heavy-weight task would have on the cluster's ability to serve traffic,
+full compaction is incremental: before the next range (presently a region) is
+compacted, we check if the load is below a certain threshold; if the threshold
+is exceeded, we pause the task until the load is again below a certain
+threshold.
+
+## Motivation
```
tonyxuqqi, line 20: Maybe we can mention that tikv-ctl compact-cluster may impact online user traffic a lot, makes it often not viable for product environment usage without downtime.  This tool provides the way to achieve the same goal, but in a more controllable way

tonyxuqqi (approved):

afeinberg:
Added graphs. Ready to merge.

tonyxuqqi (approved):

v01dstar (approved):
LGTM

text/0110-periodic-full-compaction.md
```diff
@@ -0,0 +1,386 @@
...
+below.)
+
+>
+> `tikv.toml` setting to run compaction at 03:00 and 23:00
+> (3am and 11pm respectively) in the tikv nodes' local timezone if CPU
+> usage is below 90%:
+>
+> ```toml
+>[raftstore]
+>periodic-full-compact-start-max-cpu = 0.9
```
bufferflies, line 53: Do you need to consider the compact of the dist?

afeinberg: @bufferflies  What do you mean by `compact of the dist`?

bufferflies: I mean that the impact of dist.

afeinberg: I believe, no, this is independent and runs on each tikv node (if by dist you mean the distributed aspect.) If you mean the key distribution, it is also independent of this.

bufferflies: In my understant, the compact may bring extra disk IO, so it impact the write duration of  other component like wal or flush.

afeinberg: Oh,you typed 'dist' instead of 'disk', which confused me. Yes, there is impact on all kinds of live traffic (including on disk), but it can be mitigated by configuring full compaction to run at times when it is least loaded.

In the future (see future work), it will be possible to use other metrics with full compaction, e.g., pausing when disk usage is high.

afeinberg:
Added a bit on disk metrics to the future work section.

bufferflies:
> Added a bit on disk metrics to the future work section.

It should be better if we can che

> Added a bit on disk metrics to the future work section.

If an internal API interface could be provided to inform the caller that TiKV's resources are relatively idle, I believe it would be quite useful. We could rely on this interface to accomplish various tasks, such as increasing scheduling speed, performing compaction, and increasing the concurrency of analyze operations.

bufferflies (approved):

PR #111, opened 2024-03-22
RFC: In-memory engine

SpadeA-Tang:
Tracking issue: https://github.com/tikv/tikv/issues/16141


SpadeA-Tang:
@afeinberg @tonyxuqqi PTAL

PR #115, opened 2024-10-31
introduce client circuit breaker

Tema:
RFC for https://github.com/tikv/pd/issues/8678 

Rendered view: https://github.com/Tema/rfcs/blob/0115-circuit-breaker/text/0115-circuit-breaker.md

### DIFF
--- a/text/0115-circuit-breaker.md
+++ b/text/0115-circuit-breaker.md
@@ -1,0 +1,110 @@
+# Title
+
+- RFC PR: https://github.com/tikv/rfcs/pull/115
+- Tracking Issue: https://github.com/tikv/pd/issues/8678
+
+## Summary
+
+Implement a [circuit breaker](https://learn.microsoft.com/en-us/azure/architecture/patterns/circuit-breaker) pattern to protect the PD leader from overloading due to retry storms or similar feedback loops.
+
+## Motivation
+
+In large TiDB clusters with hundreds of tidb and tikv nodes, the PD leader can become overwhelmed during certain failure conditions, leading to a "retry storm" or other feedback loop scenarios. 
+Once triggered, PD transitions into a metastable state and cannot recover autonomously, leaving the cluster in a degraded or unavailable state.
+This feature is especially critical for large clusters where a high number of nodes can continuously hammer a single PD instance during failures, leading to cascading effects that worsen recovery times and overall cluster health.
+
+## Detailed design
+
+### Configuration
+
+The key part of the circuit breaker is its configuration. 
+In order to limit number of system variables, we will use a single system variables per circuit breaker type to define error rate threshold. 
+
+#### System variables
+
+`tidb_cb_pd_metadata_error_rate_threshold_pct`:
+- Scope: GLOBAL
+- Persists to cluster: Yes
+- Applies to hint SET_VAR: No
+- Type: Integer
+- Default value: 0
+- Range: [0, 100]
+- This variable is used to set percent of errors to trip the circuit breaker.`0` means circuit breaker is disabled.
+
+#### Hardcoded configs
+
+* `error_rate_window`
+* Defines how long to track errors before evaluating error_rate_threshold.
+* Default: 10
+* Unit: seconds
+---
+* `min_qps_to_open`
+* Defines the average qps over the `error_rate_window` that must be met before evaluating the error rate threshold.
+* Default: 10
+* Unit: integer
+---
+* `cooldown_interval`
+* Defines how long to wait after circuit breaker is open before go to half-open state to send a probe request. This interval always equally jittered in the `[value/2, value]` interval.
+* Default: 60
+* Unit: seconds
+---
+* `half_open_success_count`
+* Defines how many subsequent requests to test after cooldown period before fully close the circuit. All request in excess of this count will be errored till the circuit is fully closed pending results of the firsts `half_open_success_count` requests. 
+* Default: 10
+* Unit: integer
+
+### Circuit Breaker API
+
+#### Config
+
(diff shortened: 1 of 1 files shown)


text/0115-circuit-breaker.md
```diff
@@ -0,0 +1,111 @@
+# Title
+
+- RFC PR: https://github.com/tikv/rfcs/pull/115
+- Tracking Issue: https://github.com/tikv/repo/issues/0000
```
rleungx, line 4: We can use this issue: https://github.com/tikv/pd/issues/8678

rleungx:
@nolouch @okJiang Please also take a look.

text/0115-circuit-breaker.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Circuit Breaker API
+
+#### Config
+
+All configs described above will be encapsulated in a `Settings` struct with ability to change error rate threshold dynamically:
+
+```go
+type Settings struct {
+	Type                  string
```
rleungx, line 64: How about adding more details for `Type` here?

text/0115-circuit-breaker.md
```diff
@@ -0,0 +1,110 @@
...
+- This variable is used to set percent of errors to trip the circuit breaker.`0` means circuit breaker is disabled.
+
+#### Hardcoded configs
+
+* `error_rate_window`
+* Defines how long to track errors before evaluating error_rate_threshold.
+* Default: 10
+* Unit: seconds
+---
+* `min_qps_to_open`
```
rleungx, line 41: Do we need a config for it?

text/0115-circuit-breaker.md
```diff
@@ -0,0 +1,110 @@
...
+* Defines the average qps over the `error_rate_window` that must be met before evaluating the error rate threshold.
+* Default: 10
+* Unit: integer
+---
+* `cooldown_interval`
+* Defines how long to wait after circuit breaker is open before go to half-open state to send a probe request. This interval always equally jittered in the `[value/2, value]` interval.
```
okJiang, line 47: Why is it a range?

text/0115-circuit-breaker.md
```diff
@@ -0,0 +1,110 @@
...
+## Detailed design
+
+### Configuration
+
+The key part of the circuit breaker is its configuration. 
+In order to limit number of system variables, we will use a single system variables per circuit breaker type to define error rate threshold. 
+
+#### System variables
+
+`tidb_cb_pd_metadata_error_rate_threshold_pct`:
```
okJiang, line 25: Seems
1. `pd_metadata` is difficult to understand
2. `error_rate` is redundant with pct

How about renaming it to `pd_client_cb_error_rate_threshold`?

text/0115-circuit-breaker.md
okJiang: If "had only 1 request within error_rate_window", you do not have to enable the cb?

text/0115-circuit-breaker.md
okJiang: If I remember correctly, the conclusion of the last meeting was that a single switch should control all interfaces.

text/0115-circuit-breaker.md
```diff
@@ -0,0 +1,110 @@
...
+* Unit: integer
+
+### Circuit Breaker API
+
+#### Config
+
+All configs described above will be encapsulated in a `Settings` struct with ability to change error rate threshold dynamically:
+
+```go
+type Settings struct {
```
nolouch, line 63: Will this configuration be exported for the user? And what is the relationship between this and 'system variables'?

text/0115-circuit-breaker.md
```diff
@@ -0,0 +1,110 @@
...
+
+#### Call wrapper function
+
+The circuit breaker will be implemented as a state machine with Closed, Opened and HalfOpen states and expose the following wrapper function for calls which we want to protect: 
+
+```go
+func (cb *CircuitBreaker[T]) Execute(run func() (T, error, boolean)) (T, error)
+```
+
+There is a third boolean parameter in the function argument above, in case the provided function doesn't return an error, but the caller still wants to treat it as a failure for the circuit breaker counting purposes (e.g. empty or corrupted result).
```
nolouch, line 83: Perhaps we can run a test where only the gRPC and timeout errors are treated as actual errors, which might be sufficient.

text/0115-circuit-breaker.md
okJiang: lgtm

text/0115-circuit-breaker.md
okJiang: Do you have any suggestions about this? @rleungx @nolouch

rleungx:
If there is no more comment, we can move on. /cc @niubell @nolouch @okJiang

text/0115-circuit-breaker.md
rleungx: If we have a dedicated circuit breaker for get regions calls, do we need some variables to control other calls?

okJiang: Do you mean we are only developing the region metadata part this time?🤔 @Tema

rleungx: IMO, it is better to manage an interface with a variable. For the circuit breaker itself, different interfaces can share the same implementation and only use a different setting. Right now we can focus on the GetRegion call. /cc @okJiang

rleungx (approved):

rleungx:
@nolouch @okJiang @niubell Do you have more comments? If not, I think we can merge this.

nolouch (approved):

okJiang (approved):

PR #116, opened 2024-12-02
RFC:follower read cache

mittalrishabh:
RFC to improve the performance and reliability of follower reads.

### DIFF
--- a/text/0113-follower-read-cache.md
+++ b/text/0113-follower-read-cache.md
@@ -10,12 +10,12 @@ To mitigate this issue in read-heavy environments, it is not necessary to send a
 
 ## Detailed design
 
-Safe TS in follower is used to determine the minimum timestamp that guarantees the consistent reads from replica. Currently, safe TS is advanced by advance-ts message from the leader. In this proposal, response of read index messsages can also be used to advance safeTS in follower if there are no pending pre writes on leader. 
-Changes to advance safe ts from raft read index response is pretty straightforward and most of the existing logic in the stale read path will be reused. 
+safe-ts in follower is used to determine the minimum timestamp that guarantees the consistent reads from replica. Currently, safe-ts is advanced by advance-ts message from the leader. In this proposal, response of read index messsages can also be used to advance safe-ts in follower if there are no pending pre writes on leader. 
+Changes to advance safe-ts from raft read index response is pretty straightforward and most of the existing logic in the stale read path will be reused. 
 Ref counters per region need to be added in the leader to keep track of total number of pending locks and lock status need to be included in a raft read index response. These are the detailed steps
 - Read index request check the lock reference counter and send its status through raft message in read index response
-- Update safe ts if there are no locks at this timestamp in replica.
-- Follower reads will initially compare the Safe TS with the timestamp in the request, and only if the request timestamp is greater than the Safe TS, a read index request will be sent.
+- Follower read response call update safe-ts if there are no locks at this timestamp in leader.
+- Follower reads will initially compare the safe-ts with the timestamp in the request, and only if the request timestamp is greater than the safe-ts, a read index request will be sent.
 
 This feature would be enabled by default and no new configuration will be added. 
 
@@ -31,11 +31,53 @@ pub struct ReadIndexContext {
     pub memory_lock: Option<u64>,
 }
 ```
+### Code changes
+
+Changes on follower
+```
+fn propose_raft_command {
+    RequestPolicy::ReadIndex => {
+        // try_local_stale_read
+        // if time stamp > safe-ts than send read index message to leader
+    }
+}
+
+fn apply_reads {
+    let read_index_ctx = ReadIndexContext::parse(state.request_ctx.as_slice()).unwrap();
+    if read_index_ctx.memory_lock == Some(1) {
+        // there are no pending locks on leader update safe-ts
+        self.read_progress.update_safe_ts(apply_indx, ts)
+    }
+}
+```
+
+Changes on leader
+```
+impl ReadIndexObserver for ReplicaReadLockChecker {
+    fn on_step(&self, msg: &mut eraftpb::Message, role: StateRole) {
+        // set memory lock to false if in memory lock on a region is false
+    }
+
+impl<EK, ER> Peer<EK, ER>
+where
+    EK: KvEngine,
+        ER: RaftEngine,
+        {
+            pub fn step<T>() -> Res {
+                // if memory lock is set to false in on_step()
+                rctx.memory_lock = Some(1)
+        }
+```
 
 ## Drawback
 Based on the experiements it helps in reducing the read index requests by 70% if ```tidb_low_resolution_tso``` is set to true and ```tidb_low_resolution_tso_update_interval``` is set to 100 ms. However, there can be no cache hit if ```tidb_low_resolution_tso``` is set to false. Cache hit is proportional to ```tidb_low_resolution_tso_update_interval``` setting. 
 
 ## Alternatives
 
+<<<<<<< HEAD
 An alternative approach would be to use batching to send fewer read index requests to the replica. However, batching can increase the p50 latency and result in higher network traffic compared to caching. Caching can tolerate smaller blips on leader availability depending on ```tidb_low_resolution_tso_update_interval``` , while batching may not be as reliable.
 
+=======
+- An alternative approach would be to use batching to send fewer read index requests to the replica. However, batching can increase the p50 latency and result in higher network traffic compared to caching. Caching can tolerate smaller blips on leader unavailable depending on ```tidb_low_resolution_tso_update_interval``` , while batching may not be as reliable.
+- Another alternate approach in design is to introduce new timestamp (raftReadIndex-ts) instead of safe-ts. It will unnecessarily complicates the design. 
+>>>>>>> e500e31 (update review comments)
--- a/text/0113-follower-read-cache.md
+++ b/text/0113-follower-read-cache.md
@@ -6,7 +6,7 @@ The purpose of this design is to minimize the CPU load on the leader by caching 
 ## Motivation
 
 In a cloud environment, customers often prefer to perform follower reads within the same AZ to avoid incurring cross-AZ network costs. However, each follower read requires sending a read index request to the leader to check for any pending write locks in memory. These RPC calls can be quite expensive and can significantly increase CPU utilization, especially in read-heavy traffic scenarios. This increase in CPU cost eliminates the cost savings achieved by avoiding cross-AZ traffic.
-To mitigate this issue in read-heavy environments, it is not necessary to send a read index request for every single follower read request. This is because, in 90% of cases, there are no pending locks. Instead, we can advance the safe TS if there are no locks present in the leader at that particular timestamp. Therefore, online reads can first check for the safe TS before sending a read index request to the leader.
+To mitigate this issue in read-heavy environments, it is not necessary to send a read index request for every single follower read request. This is because, in 99% of cases, there are no pending locks. 
 
 ## Detailed design
 
@@ -37,5 +37,5 @@ Based on the experiements it helps in reducing the read index requests by 70% if
 
 ## Alternatives
 
-An alternative approach would be to use batching to send fewer read index requests to the replica. However, batching can increase the p50 latency and result in higher network traffic compared to caching. Caching can tolerate smaller blips on leader unavailable depending on ```tidb_low_resolution_tso_update_interval``` , while batching may not be as reliable.
+An alternative approach would be to use batching to send fewer read index requests to the replica. However, batching can increase the p50 latency and result in higher network traffic compared to caching. Caching can tolerate smaller blips on leader availability depending on ```tidb_low_resolution_tso_update_interval``` , while batching may not be as reliable.
 
--- a/text/0113-follower-read-cache.md
+++ b/text/0113-follower-read-cache.md
@@ -36,6 +36,43 @@ pub struct ReadIndexContext {
     pub locked: Option<LockInfo>,
     pub memory_lock: Option<u64>,
 }
+
+const REQUEST_FLAG: u8 = b'r';
+const LOCKED_FLAG: u8 = b'l';
+const MEMORY_LOCKED_FLAG: u8 = b'm'; // new memory lock flag. its appended at the end. Old and new versions are compatible with each other.
+
+// encode : convert it into bytes
+pub fn fields_to_bytes(
+        id: Uuid,
+        request: Option<&raft_cmdpb::ReadIndexRequest>,
+        locked: Option<&LockInfo>,
+        memory_locked: Option<u64>,
+    ) -> Vec<u8> {
+    if let Some(memory_locked) = memory_locked {
+            b.push(MEMORY_LOCKED_FLAG);
+            b.encode_var_u64(memory_lock_size.unwrap() as u64).unwrap();
+            b.extend_from_slice(&memory_locked.to_le_bytes());
+    }
+}
+ // decode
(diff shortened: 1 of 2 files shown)


text/0113-follower-read-cache.md
```diff
@@ -0,0 +1,41 @@
...
+To mitigate this issue in read-heavy environments, it is not necessary to send a read index request for every single follower read request. This is because, in 99% of cases, there are no pending locks. 
+
+## Detailed design
+
+Safe TS in follower is used to determine the minimum timestamp that guarantees the consistent reads from replica. Currently, safe TS is advanced by advance-ts message from the leader. In this proposal, response of read index messsages can also be used to advance safeTS in follower if there are no pending pre writes on leader. 
+Changes to advance safe ts from raft read index response is pretty straightforward and most of the existing logic in the stale read path will be reused. 
+Ref counters per region need to be added in the leader to keep track of total number of pending locks and lock status need to be included in a raft read index response. These are the detailed steps
+- Read index request check the lock reference counter and send its status through raft message in read index response
+- Update safe ts if there are no locks at this timestamp in replica.
```
cfzjywxk, line 17: Why should locks on replicas be considered? For advancing `resolved-ts` or `safe-ts`, the current TiKV logic determines whether to update `safe-ts` based on the timestamp and applied index sent by the leader.
Perhaps the `safe-ts` update should reuse existing processes to ensure correctness.

text/0113-follower-read-cache.md
```diff
@@ -0,0 +1,41 @@
...
+A new field `memory_lock` will be added to `ReadIndexContext` to send the lock status to replica.
+
+```
+pub struct ReadIndexContext {
+    pub id: Uuid,
+    pub request: Option<raft_cmdpb::ReadIndexRequest>,
+    pub locked: Option<LockInfo>,
+    pub memory_lock: Option<u64>,
+}
+```
```
cfzjywxk, line 33: Could you add more detailed descriptions using pseudo code based on the TiKV implemenations, it is better for us to align the code details with the previously detailed designs.

Like taking the code path of a single follower read as an example, and describe the change differences with the data flow descriptions.

text/0113-follower-read-cache.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Motivation
+
+In a cloud environment, customers often prefer to perform follower reads within the same AZ to avoid incurring cross-AZ network costs. However, each follower read requires sending a read index request to the leader to check for any pending write locks in memory. These RPC calls can be quite expensive and can significantly increase CPU utilization, especially in read-heavy traffic scenarios. This increase in CPU cost eliminates the cost savings achieved by avoiding cross-AZ traffic.
+To mitigate this issue in read-heavy environments, it is not necessary to send a read index request for every single follower read request. This is because, in 99% of cases, there are no pending locks. 
+
+## Detailed design
+
+Safe TS in follower is used to determine the minimum timestamp that guarantees the consistent reads from replica. Currently, safe TS is advanced by advance-ts message from the leader. In this proposal, response of read index messsages can also be used to advance safeTS in follower if there are no pending pre writes on leader.
```
bb7133, line 13: A soft reminder: please try to keep the term 'safe TS' in the same style. Now we have 'safe TS', 'safeTS' and 'safe ts' in this RFC.

text/0113-follower-read-cache.md
mittalrishabh: We are going to use the existing logic to update safe-ts. It is raft who is going to update safe-ts only when there are no locks on leader.

text/0113-follower-read-cache.md
mittalrishabh: i will add it.

text/0113-follower-read-cache.md
mittalrishabh: I will fix it

mittalrishabh:
@cfzjywxk @bb7133 please review it as i incorporated the comments

text/0113-follower-read-cache.md
```diff
@@ -0,0 +1,79 @@
...
+        // if time stamp > safe-ts than send read index message to leader
+    }
+}
+
+fn apply_reads {
+    let read_index_ctx = ReadIndexContext::parse(state.request_ctx.as_slice()).unwrap();
+    if read_index_ctx.memory_lock == false {
+        // there are no pending locks on leader update safe-ts
+        // No need to take core lock as it is happening in raft thread
+        self.read_progress.update_safe_ts(apply_indx, ts)
```
cfzjywxk, line 50: How the `ts` updating `safe_ts` on the follower is calculated? Is it the `read-ts` of the read request?

No memory locks indicates no pending write transctions, there could be still onoging transaction with prewrite locks in the lock-cf.

text/0113-follower-read-cache.md
```diff
@@ -0,0 +1,79 @@
...
+To mitigate this issue in read-heavy environments, it is not necessary to send a read index request for every single follower read request. This is because, in 99% of cases, there are no pending locks. 
+
+## Detailed design
+
+safe-ts in follower is used to determine the minimum timestamp that guarantees the consistent reads from replica. Currently, safe-ts is advanced by advance-ts message from the leader. In this proposal, response of read index messsages can also be used to advance safe-ts in follower if there are no pending pre writes on leader. 
+Changes to advance safe-ts from raft read index response is pretty straightforward and most of the existing logic in the stale read path will be reused. 
+Ref counters per region need to be added in the leader to keep track of total number of pending locks and lock status need to be included in a raft read index response. These are the detailed steps on how a follower read works
+- Follower reads will initially compare the safe-ts with the timestamp in the request, and only if the request timestamp is greater than the safe-ts, a read index request will be sent.
```
cfzjywxk, line 16: Perhaps we could merge this optimization first, it is relatively independent and theoretically has no side effects.

mittalrishabh: yes, we can do this one first.

text/0113-follower-read-cache.md
mittalrishabh: Yes, safe TS can surpass resolve-TS with this change in follower. I think it should be OK to break the assumption that safe TS is always behind resolved TS in follower.

cfzjywxk: Currently, `safe_ts` guarantees the **snapshot read** semantics, meaning no data with `commit_ts <= safe_ts` will be written anymore.

Although `read_ts` on the leader does not detect any memory locks, it only indicates that there are no ongoing async commit/1PC transactions. However, there may still exist prewrite locks in the lock-cf, and these prewrite locks might write data with `commit_ts < read_ts` during the commit process. Therefore, directly using `read_ts`, which only checks the leader memory lock, to update the follower's `safe_ts` could break the snapshot guarantees provided by `safe_ts` on the follower.

cfzjywxk: If `read_ts` is to be used to update the follower peer's `safe_ts`, another approach is to update it not when the read index response is received, but after the follower read has been successfully processed. If the follower read completes correctly and returns results, it indicates that no conflicting prewrite locks were encountered on the follower. Combined with the known condition that no memory locks were read on the leader, this might ensure that no transactions with `commit_ts <= safe_ts` are written, thereby preserving the snapshot semantics.

mittalrishabh: why safe_ts need to guarantees the snapshot read semantics.

mittalrishabh: Can we say that 'safe-ts' is the timestamp that indicates up to which point the follower has committed data from the Raft state machine?

cfzjywxk: The concepts of `safe_ts` and `follower-read` are indeed separate and serve different purposes.

- **Follower-read**: The read requests to be served as long as the `applied_index` on the follower is greater than the `commit_index` from the leader. It ensures that the data read is up-to-date with the leader’s committed state but doesn’t directly guarantee future write states.

- **`safe_ts` semantics**: The `safe_ts` guarantees that a read with `read_ts <= safe_ts` satisfies snapshot semantics. This means no transactions with `commit_ts <= safe_ts` will be written in the future.  

In the scenario described:  
- If there’s a **prewrite lock with `start_ts = 1`**, a `read_ts = 3` cannot be used to update the follower’s `safe_ts`, even if no memory lock is observed on the leader. The reason is that the prewrite lock might eventually commit a record with `commit_ts = 2`, which violates the snapshot semantics if the `safe_ts` had already been advanced to 3.

Therefore, advancing `safe_ts` requires additional safeguards to ensure no transactions with `commit_ts < read_ts` can still be committed after the `safe_ts` is updated. This is why relying solely on the absence of memory locks in the leader when using `read_ts` to update `safe_ts` on follower peer is insufficient.

cfzjywxk: > Can i introduce apply-ts which will be updated by read index.

This introduces more concepts and changes. If the current goal is not to completely refactor the stale read design to achieve the x00ms target, there’s no need to make it this complex.  

If the **goal** is to reuse the information in the read index response to advance `safe_ts`, the mentioned approach can be used already:  
1. If the read index response does not contain **memory locks**, it indicates that the leader currently has no ongoing async commit transactions writing.  
2. If the follower read can process this request successfully at `read_ts` without returning a "key is locked" error, it means there are no conflicting prewrite locks at `read_ts`.  

If both conditions are met, the follower peer can consider using `read_ts` to update `safe_ts` directly.

mittalrishabh: I agree that introducing new ts complicates the system. Ideally I want to update a ts (lets call it last-read-index ts) as soon as it get the read index response because we have all the information to serve the request with ts < last-read-index ts locally. It will help in two ways 
1. It allows more read index requests to be served locally
2. In case of retry, we might avoid retrying all the read index request if there is a missing response in between.

Having said that we can do incremental development  
1. We can experiment with updating safe-ts only after follower read is successfully read locally.
2. If above doesn't give better performance, we can introduce a new semantic last-read-index-ts. I don't think there will be require too many code changes.

cfzjywxk: If the goal is solely to optimize follower reads by allowing more read index requests to be served locally, introducing `last-read-index` seems fine. It hepls keeping the optimization confined to follower reads avoids coupling with `safe_ts` or `resolved_ts` and introduce risks to safe_ts-related functionality.

mittalrishabh: incorporated the review comments

text/0113-follower-read-cache.md
```diff
@@ -0,0 +1,83 @@
...
+```
+pub struct ReadIndexContext {
+    pub id: Uuid,
+    pub request: Option<raft_cmdpb::ReadIndexRequest>,
+    pub locked: Option<LockInfo>,
+    pub memory_lock: Option<u64>,
+}
+```
+### Code changes
+
```
cfzjywxk, line 36: Better to describe how the `last-read-index-ts` is maintained in the code change part, like where it's stored and how it's read and modified.

text/0113-follower-read-cache.md
```diff
@@ -0,0 +1,83 @@
...
+To mitigate this issue in read-heavy environments, it is not necessary to send a read index request for every single follower read request. This is because, in 99% of cases, there are no pending locks. 
+
+## Detailed design
+
+We introduce a new concept last-read-index-ts. last-read-index-ts is updated with the read-ts in the follower under two specific conditions:
+1. There is no memory lock in that region.
+2. The apply_index in the follower is greater than the commit_index of the leader.
+In this proposal, response of read index messsages can be used to advance last-read-index-ts in follower. Ref counters per region need to be added in the leader to keep track of total number of pending locks and lock status need to be included in a raft read index response. These are the detailed steps on how a follower read works
+- Follower reads will initially compare the last-read-index-ts with the timestamp in the request, and only if the request timestamp is greater than the last-read-index-ts, a read index request will be sent.
```
cfzjywxk, line 17: A detailed proof process is needed here to demonstrate that reusing `last-read-index-ts` still maintains linearizability and the semantics of a globally strong consistent snapshot. Especially  taking the `async commit` protocol into considerations.  

This might take some explanation and could be supplemented when filing the tracking issue.

text/0113-follower-read-cache.md
mittalrishabh: it is stored as part of read_progress. its modification is shown in line 50. I have added a structure also now.

text/0113-follower-read-cache.md
mittalrishabh: added explanation line 21 to 24.

text/0113-follower-read-cache.md
```diff
@@ -0,0 +1,108 @@
...
+- Read index request on leader check the lock reference counter and send its status through raft message in read index response
+- read index response update last-read-index-ts if there are no locks at this timestamp in leader.
+
+When a read index message is sent to the leader, it first checks the memory lock for the region and the read index keys. If there is no lock on the keys, it updates the maximum read timestamp before responding with the read index. The response includes the in-memory or key lock status and the commit index. 
+- Once the maximum read timestamp is updated (let's say to ts1) at commit index (x1), and if there is no memory lock in a region on any key, there will be no pre-commit write requests with a timestamp lower than ts1.
```
cfzjywxk, line 22:
```suggestion
- Once the maximum read timestamp is updated (let's say to ts1) at commit index (x1), and if there is no memory lock in a region on any key, there will be no pre-commit write requests with a timestamp lower than ts1. Since `ReadIndexObserver` updates `max_ts` before checking the current region's memory lock count, and newly added memory locks acquire the lock first and then attempt to read `max_ts`, any memory lock acquired after the check is guaranteed to observe the `max_ts` updated by `read_ts`.
```

text/0113-follower-read-cache.md
```diff
@@ -0,0 +1,108 @@
...
+### Protocol
+
+A new field `memory_lock` will be added to `ReadIndexContext` to send the lock status to replica.
+
+```
+pub struct ReadIndexContext {
+    pub id: Uuid,
+    pub request: Option<raft_cmdpb::ReadIndexRequest>,
+    pub locked: Option<LockInfo>,
+    pub memory_lock: Option<u64>,
```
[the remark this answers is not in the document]
cfzjywxk: The read index response from leader does not contain information about `memory locks in this region` by now, 
```
message ReadIndexResponse{
    uint64 read_index = 1;
    // The memory lock blocking this read at the leader
    kvrpcpb.LockInfo locked = 2;
}
```
, we may need to add a new field for it like
```
message ReadIndexResponse{
    uint64 read_index = 1;
    // The memory lock blocking this read at the leader
    kvrpcpb.LockInfo locked = 2;
    // The memory lock count on the leader peer
    uint64 memory_lock_cnt = 3;
}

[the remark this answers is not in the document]
Comment (mittalrishabh) [reply text/0113-follower-read-cache.md:37]: I get this part. Where are we using it in V1 ?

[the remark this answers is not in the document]
Comment (mittalrishabh) [reply text/0113-follower-read-cache.md:37]: i have added more details about encoding and decoding line 40 : 75. I can change proto buf too. But i am not sure if we are using it anywhere.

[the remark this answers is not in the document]
Comment (cfzjywxk) [reply text/0113-follower-read-cache.md:37]: Here ?
```
if read_index_ctx.memory_lock == false  // Based on the response?
```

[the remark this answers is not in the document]
mittalrishabh: it is taking from ```ReadIndexContext::parse(state.request_ctx.as_slice()).unwrap();```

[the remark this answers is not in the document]
cfzjywxk: Ok

cfzjywxk (approved):
LGTM

glorv (approved):
LGTM

PR #121, opened 2025-12-11
[RFC ]region level isolation

mittalrishabh:
Created RFC for region level isolation

### DIFF
--- a/text/region-level-resource-isolation.md
+++ b/text/region-level-resource-isolation.md
@@ -2,22 +2,22 @@
 
 ## Summary
 
-This RFC proposes enhancements to TiKV's resource control system to provide region-level isolation, preventing hot regions from overwhelming tenant resources. The design extends the existing resource group-based priority system with region-level RU (Resource Unit) tracking and introduces traffic moderation mechanisms.
+This RFC proposes enhancements to TiKV's resource control to provide region-level isolation, preventing hot regions from overwhelming tenant resources. The design extends the existing resource group-based priority system with region-level RU (Resource Unit) tracking and introduces traffic moderation mechanisms.
 
 ## Motivation
 
 ### Current State
 
 TiKV implements resource control at the **resource group level**:
-- Resource groups represent tenants and track RU (Request Unit) consumption
-- Each resource group has a `group_priority` (LOW, MEDIUM, HIGH) configured in PD
-- The `ResourceController` maintains virtual time (VT) per resource group for fairness. VT increases as a group consumes resources - groups with higher VT have consumed more and get lower scheduling priority
-- The DM (Deadline Multilevel) queue orders tasks by priority: `concat_priority_vt(group_priority, resource_group_virtual_time)`. Lower values are scheduled first, so high-priority groups with low VT run first
+- Resource groups represent tenants and track RU (Resource Unit) consumption
+- Each resource group has a `group_priority` (LOW, MEDIUM, HIGH) configured 
+- The `ResourceController` uses mClock algorithm to prioritize the requests. It maintains virtual time (VT) per resource group for fairness. VT increases as a group consumes resources - groups with higher VT have consumed more and get lower scheduling priority
+- Tasks are ordered by priority: `concat_priority_vt(group_priority, resource_group_virtual_time)`. Lower values are scheduled first, so high-priority groups with low VT run first
 - VT is periodically normalized to prevent starvation: lagging groups (low VT) are pulled toward the leader (highest VT), and all VTs are reset when nearing overflow
-- The unified read pool uses yatp's priority queue which implements this DM queue using a SkipMap
+- The unified read pool uses yatp's priority queue (implemented with a SkipMap)
 
 ## Goals
-These functionalities are missing in the current state. 
+These functionalities are missing in the current implementation. 
 
 1. **Region-level fairness**: Hot regions (with hot keys or large scans) should be deprioritized to prevent resource monopolization within a tenant
 3. **Traffic Moderation**: In a multi-tenant SOA environment, setting correct rate limits is challenging - limits that are too tight reject valid traffic, while limits that are too loose allow overload. Instead of hard rate limits, implement adaptive traffic moderation that responds to sudden spikes on hot regions by gracefully deprioritizing rather than outright rejecting requests
@@ -314,4 +314,5 @@ enable-region-tracking = true
 ```
 
 ## Drawbacks
-1. Currently, split/scatter is non-deterministic when node is overloaded - it depends on how many requests on this region are succeeded. With this design, hot regions are deliberately deprioritized, so split/scatter may take longer when overloaded. This is an acceptable tradeoff for better traffic moderation
+
+1. **Temporary traffic moderation**: The VT-based traffic moderation is temporary. It works until: (a) periodic VT normalization equalizes VTs across regions (typically minutes), or (b) node reboot resets all VTs. After normalization, previously hot regions return to normal priority even if still hot. This provides short-term relief during overload but not long-term rate limiting.
--- a/text/region-level-resource-isolation.md
+++ b/text/region-level-resource-isolation.md
@@ -64,13 +64,35 @@ struct TaskPriority {
 - Pull lagging entities toward leader (prevent starvation)
 - Reset all VTs if near overflow
 
-### Traffic moderation and split/scatter 
+### Traffic moderation and split/scatter
 
-Currently, split/scatter is non-deterministic when node is overloaded - it depends on how many requests on this region are succeeded. With this design, Hot regions accumulate high VT and get deprioritized, which slows down split decisions based on served QPS. When the region eventually splits, the new regions can either:
-1. **Inherit parent VT** (recommended): New regions start with the same high VT as parent, maintaining deprioritization until they cool down naturally via normalization
-2. **Reset to zero VT**: New regions start fresh, immediately getting normal priority
+Currently, split/scatter is non-deterministic when node is overloaded - it depends on how many requests on this region are succeeded. With this design, Hot regions accumulate high VT and get deprioritized, which slows down split decisions based on served QPS.
 
-Option 1 combined with delayed splitting provides stronger traffic moderation - the system temporarily reduces throughput to the hot table/region until the load naturally subsides or spreads across the split regions
+#### VT Handling for Split Regions
+
+When a region splits, the VT behavior depends on CPU utilization:
+
+**When CPU utilization > 80% (system overloaded)**:
+- Split regions share a **common VT** inherited from the parent region
+- Both child regions contribute to and read from the same VT tracker
+- This maintains strong traffic moderation - even after splitting, the hot key/region group remains deprioritized as a unit
+- The common VT continues accumulating based on combined traffic to both regions
+- This prevents the split from immediately bypassing the backpressure that delayed the split in the first place
+
+**When CPU utilization drops < 80% (system has capacity)**:
+- Split regions transition to **independent VTs**
+- Each region gets its own VT tracker, initialized to the common VT value at time of transition
+- From this point forward, each region accumulates VT based on its own traffic patterns
+- This allows natural load balancing - if traffic shifts to one split region, only that region gets deprioritized
+
+**Implementation**:
+- Track CPU utilization as a rolling average (e.g., last 10 seconds)
+- On region split, create a `RegionGroup` if CPU > 80%, linking child regions to shared VT
+- Periodically check CPU utilization (every 1-5 seconds)
+- When CPU drops < 80%, dissolve region groups and transition to independent VTs
+- Store region group membership in `RegionVtTracker` with atomic reference to shared VT state
+
+This adaptive approach provides stronger traffic moderation when the system is overloaded (maintaining backpressure across splits), while allowing normal load balancing when the system has capacity
 
 ### Background Task Demotion
 
@@ -149,17 +171,38 @@ Create new component for region-level tracking:
 
 struct RegionResourceTracker {
     region_vts: DashMap<u64, RegionVtTracker>,
+    cpu_utilization: AtomicU64,  // Rolling average, encoded as u64
 }
 
 struct RegionVtTracker {
     virtual_time: AtomicU64,
     vt_delta_for_get: AtomicU64,
+    parent_vt: Option<Arc<AtomicU64>>,  // Shared parent VT if CPU > 80% at split
 }
 
 impl RegionResourceTracker {
     fn get_and_increment_vt(region_id) -> u64 {
+        // If parent_vt exists, use shared parent VT
+        // Otherwise use independent VT
         // Similar to ResourceGroup::get_priority()
-        // Returns current VT and increments by vt_delta
+    }
+
+    fn on_region_split(parent_id, child1_id, child2_id) {
+        // Get parent VT value
+        // If cpu_utilization > 80%:
+        //   Create Arc<AtomicU64> with parent VT
+        //   Both children share reference to parent_vt
+        // Else:
+        //   Both children get independent VT initialized to parent VT
+        //   parent_vt = None
+        // Remove parent tracker
+    }
+
+    fn check_and_transition_to_independent() {
+        // If cpu_utilization < 80%:
+        //   For each region with parent_vt:
+        //     Copy parent_vt value to virtual_time
+        //     Set parent_vt to None
     }
 
     fn update_vt_deltas() {
@@ -176,6 +219,20 @@ impl RegionResourceTracker {
     fn consume(region_id, cpu_time, keys, bytes) {
         // Update EMA metrics
         // Increment VT based on actual consumption
+        // If parent_vt exists, increment shared parent VT
+        // Otherwise increment independent VT
+    }
+
+    fn update_cpu_utilization(cpu_util) {
+        // Update rolling average (EMA over ~10 seconds)
+    }
+
+    fn cleanup_inactive_regions() {
+        // Periodically remove regions with no recent VT updates
+        // For each region:
+        //   If virtual_time hasn't changed in last N seconds:
+        //     Remove from region_vts hashmap
+        // This reduces memory usage for cold/deleted regions
     }
 }
 ```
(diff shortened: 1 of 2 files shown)


text/region-level-resource-isolation.md
```diff
@@ -0,0 +1,380 @@
...
+- Tasks are ordered by priority: `concat_priority_vt(group_priority, resource_group_virtual_time)`. Lower values are scheduled first, so high-priority groups with low VT run first
+- VT is periodically normalized to prevent starvation: lagging groups (low VT) are pulled toward the leader (highest VT), and all VTs are reset when nearing overflow
+- The unified read pool uses yatp's priority queue (implemented with a SkipMap)
+
+## Goals
+These functionalities are missing in the current implementation. 
+
+1. **Region-level fairness**: Hot regions (with hot keys or large scans) should be deprioritized to prevent resource monopolization within a tenant
+3. **Traffic Moderation**: In a multi-tenant SOA environment, setting correct rate limits is challenging - limits that are too tight reject valid traffic, while limits that are too loose allow overload. Instead of hard rate limits, implement adaptive traffic moderation that responds to sudden spikes on hot regions by gracefully deprioritizing rather than outright rejecting requests
```
Tema: nit: 3rd when it is 2nd

This point sounds more like non-goal.  Move it to alternative approaches section?

Separately, do you want to add a goal around non interfering with region split?

text/region-level-resource-isolation.md
```diff
@@ -0,0 +1,380 @@
...
+- Region VT increases by actual CPU time consumed
+
+**Periodic normalization** (every ~1 second):
+- Find min/max VT across all groups/regions
+- Pull lagging entities toward leader (prevent starvation)
+- Reset all VTs if near overflow
+
+### Traffic moderation and split/scatter
+
+Currently, split/scatter is non-deterministic when node is overloaded - it depends on how many requests on this region are succeeded. With this design, Hot regions accumulate high VT and get deprioritized, which slows down split decisions based on served QPS.
```
Tema, line 69: If split depends on requests succeeded then deprioritization of such requests will delay the split. Should it use scheduled/dropped QPS rather than succeeded instead>

text/region-level-resource-isolation.md
```diff
@@ -0,0 +1,380 @@
...
+This adaptive approach provides stronger traffic moderation when the system is overloaded (maintaining backpressure across splits), while allowing normal load balancing when the system has capacity
+
+### Background Task Demotion
+
+Background tasks (GC, compaction, statistics) use LOW `group_priority` regardless of their resource group's configured priority:
+```
+group_priority = LOW  // instead of resource group's configured priority
+```
+
+This ensures foreground traffic is always prioritized over background.
```
Tema, line 104: Is the proposal here to create a virtual RG for background task with low priority to schedule them relative to other traffic as well?

text/region-level-resource-isolation.md
```diff
@@ -0,0 +1,380 @@
...
+
+### Queue Eviction
+
+When queue is full:
+1. Calculate priority of incoming task
+2. Compare with lowest priority task in queue
```
Tema, line 110: That is probably hard to implement effectively. Will it require more one priority queue with inverse priority? Or the current SkipMap implementation allows popping from both ends effectively?

text/region-level-resource-isolation.md
```diff
@@ -0,0 +1,380 @@
...
+# Enable region-level resource tracking
+enable-region-tracking = true
+```
+
+## Drawbacks
+
+1. **Temporary traffic moderation**: The VT-based traffic moderation is temporary. It does not work if a node is rebooted after regions are split.
+2. **Shared region fairness issues**: When multiple resource groups access the same region, two fairness problems arise:
+   - **Innocent tenant penalized**: Tenant A's heavy usage increases the region's VT, penalizing Tenant B's requests to that region even though Tenant B didn't cause the hotness
+   - **Hot region stays hot**: If Tenant A and B alternate requests to a shared region, each tenant's group_vt stays low (they're taking turns), so the region never gets properly deprioritized despite being continuously hot
```
Tema, line 378: Did you consider to have a region tracker per group? What are trade offs of this approach instead?

text/region-level-resource-isolation.md
mittalrishabh: This is a goal. If we don't have this goal then system will become overloaded after split.

text/region-level-resource-isolation.md
mittalrishabh: Yes, we can use scheduled/dropped qps. I will modify the design.

text/region-level-resource-isolation.md
```diff
@@ -0,0 +1,380 @@
...
+        let group_vt = self.resource_group(metadata.group_name())
+            .get_group_vt(level, override_priority);
+
+        // 2. Get region VT
+        let region_id = metadata.region_id();
+        let region_vt = self.region_tracker.get_and_increment_vt(region_id);
+
+        // 3. Use LOW priority for background tasks
+        let group_priority = if metadata.is_background() {
+            LOW
```
mittalrishabh, line 261: @Tema this is how priority is decided for background task

text/region-level-resource-isolation.md
mittalrishabh: tagged you below in the implementation.

text/region-level-resource-isolation.md
mittalrishabh: yes, current skipmap allows. It is O(log N) complexity.

text/region-level-resource-isolation.md
mittalrishabh: what is region tracker per group ?

Tema: Current design tracks cpu per regions across all tenants which results on the highlighted issues. I'm asking if you considered to track it per tenant.

mittalrishabh: existing design is if a region is split into r1 and r2 then they will share the same VT if cpu utilization is more than 80%.  Having region tracker per group will complicate this design.

zhangjinpeng87:
@glorv @v01dstar PTAL

text/region-level-resource-isolation.md
```diff
@@ -0,0 +1,380 @@
...
+- Each resource group has a `group_priority` (LOW, MEDIUM, HIGH) configured 
+- The `ResourceController` uses mClock algorithm to prioritize the requests. It maintains virtual time (VT) per resource group for fairness. VT increases as a group consumes resources - groups with higher VT have consumed more and get lower scheduling priority
+- Tasks are ordered by priority: `concat_priority_vt(group_priority, resource_group_virtual_time)`. Lower values are scheduled first, so high-priority groups with low VT run first
+- VT is periodically normalized to prevent starvation: lagging groups (low VT) are pulled toward the leader (highest VT), and all VTs are reset when nearing overflow
+- The unified read pool uses yatp's priority queue (implemented with a SkipMap)
+
+## Goals
+These functionalities are missing in the current implementation. 
+
+1. **Region-level fairness**: Hot regions (with hot keys or large scans) should be deprioritized to prevent resource monopolization within a tenant
```
glorv, line 22: If I understand correctly, the goal is to solve the case that large query exhausting the tenant resource and causing high latency on small queries. In this case, I suspect the "region-level resource control" can really solve this problem. As the high resource consumption can be either "large scans on a small number of regions" or "medium/small scans on a large number of regions". So, maybe "query level resource control" is what you want: automically de-priotize read requests(from the same sql) if their resource consumption(total execution cpu time) exceed a certain threshold. https://github.com/tikv/yatp/blob/793be4d789d4bd15292fe4d06e38063b4ec9d48e/src/queue/multilevel.rs#L576 Yatp's multi-level running has an implementation that degrade the level of long-running tasks.

BTW, in my experience, the current queue-based scheduling doesn’t really add the value we expected while the theory seems plausible. Maybe you can first test with "set up 2 resource groups(with different priority) and run the small/large sql under the 2 groups(large sql under the low-priority group), then see the impact of the large sql on the small sqls". This should be the best case that can be gained in resource control on a single tenant.

mittalrishabh: Here goal is to prevent set of regions from consuming all the CPUs. It generally happens in these scenarios
1. hot keys
2. MVCC
3. sudden traffic increase

I tested the m queue approach for resource groups. It works as expected ensuring that all resource groups get the fair share of resources. I am extending the same algorithm of fairness for regions.

mittalrishabh: Here goal is to prevent group of regions in a resource group from consuming all the CPUs. It generally happens in these scenarios
1. hot keys
2. MVCC
3. sudden traffic increase

I tested the m queue approach for resource groups. It works as expected ensuring that all resource groups get the fair share of resources. I am extending the same algorithm of fairness for regions.

mittalrishabh: Here goal is to prevent group of regions in a resource group from consuming all the CPUs. It generally happens in these scenarios
1. hot keys
2. MVCC
3. sudden traffic increase

I tested the m queue approach for resource groups. It works as expected ensuring that all resource groups get the fair share of resources. I am extending the same algorithm of fairness to regions.

mittalrishabh: @glorv does it make sense ?

text/region-level-resource-isolation.md
zhangjinpeng87: If there is a hot region that received sudden high QPS, the load-based split will be triggered for this region may be 30s later, and the hot-region scheduler will transfer leader to other tikv nodes. Here what problem we want to resolve is preventing the sudden hot region eat too much CPU resource and then impact other normal traffic (no matter other tenants' traffic or same tenant's traffic) before load-based split take affect. How about just introduce some simple rules for region level resource consumption, like single region can't eat more than 60% CPU resource to make TiKV node always reserve some resource to keep node level health (similar like DynamoDB's in each storage node, each tenant just can consume 50% node's resource to introduce node level self protection mechanism).

I'm afraid we add region level fairness mechanism on top of resource group level resource fairness/priority mechanism may introduce unexpected complexity and behaviors.

mittalrishabh: There are two scenarios where there could be more than one hot regions 
1.  Index and data reads happening on the same node
2. We are defining a coarse grained tenants with rate limit generally 2 - 2.5 x of the current traffic. We can have multiple hot regions if traffic spikes on a table. 
We are implementing region level fairness over existing m queue logic to keep the complexity low.

text/region-level-resource-isolation.md
v01dstar: I have the same question as @glorv.

First, the proposed solution might be counter-productive for complex queries. For example, if **a** query spawns 100 coprocessor tasks where one is a heavy scan and 99 are light, the heavy scan is the natural bottleneck. If our resource control logic deprioritizes this 'hot' task to prevent resource saturation, we are effectively stretching the tail latency of the entire query. In this case, 'fairness' becomes a penalty for the user.

I think the request/query-level fairness makes more sense?

I am not sure what region-level fairness can bring us. Do you have an example, that the QoS is impacted, if we don't do region-level fairness?

mittalrishabh: Here goal is not to de prioritize a long running query. Our goal is to de prioritize the traffic on hot regions. Lets say system is healthy and region 1, 2 ,3,4 are getting 5k qps. Suddenly region 1 and 2 start getting 20k qps and CPU is reached 100%. Without region level fairness, all regions (region 3 and region 4) will be impacted until split and scatter happens. With region level fairness, region 3 and region 4 won't get impacted.

mittalrishabh: And our goal to de prioritize traffic on region 1 and 2 only when system is under resources.

v01dstar: I wonder if we can use benchmarks like sysbench to demonstrate that without region-level fairness, tail latency suffers—and that implementing fairness can stabilize those latencies (even if total throughput remains the same). I feel it could be challenging. If we can show this, the technical merit is clear. However, I suspect it is extremely difficult to 'prove' the necessity of this change; I want to look past the technical implementation and see the actual value for the end user.

mittalrishabh: It is very easily reproducible. We have to reproduce the hot regions by reducing the unified read pool queue size and number of cores and see its impact on non hot regions.

mittalrishabh: We need to create two tables in the same resource group with one table very hot. We see the latency on other table getting impacted.  Here goal is not to stabilize tail latency. Goal is to penalize one which is consuming more resources  in a multi tenant system

text/region-level-resource-isolation.md
```diff
@@ -0,0 +1,380 @@
...
+TiKV implements resource control at the **resource group level**:
+- Resource groups represent tenants and track RU (Resource Unit) consumption
+- Each resource group has a `group_priority` (LOW, MEDIUM, HIGH) configured 
+- The `ResourceController` uses mClock algorithm to prioritize the requests. It maintains virtual time (VT) per resource group for fairness. VT increases as a group consumes resources - groups with higher VT have consumed more and get lower scheduling priority
+- Tasks are ordered by priority: `concat_priority_vt(group_priority, resource_group_virtual_time)`. Lower values are scheduled first, so high-priority groups with low VT run first
+- VT is periodically normalized to prevent starvation: lagging groups (low VT) are pulled toward the leader (highest VT), and all VTs are reset when nearing overflow
+- The unified read pool uses yatp's priority queue (implemented with a SkipMap)
+
+## Goals
+These functionalities are missing in the current implementation.
```
jiadebin, line 20: If I understand correctly, the goal of this RFC is to address the issues of system avalanches and starvation caused by hot regions. However, it carries a potential side effect: when the system is not overloaded, the excessive pursuit of region-level fairness in business hotspot scenarios may harm the performance of critical business operations.

This highlights a core conflict in defining **"Resource Scheduling Fairness"**: **Request Fairness** vs. **Region Fairness**.

Consider a scenario with two regions: `region_1` and `region_2`. 90% of requests target `region_1`, while 10% target `region_2`. All requests are small queries (TP requests), and the system is under high load but has not yet exhausted all resources (e.g., at 98% utilization). Under the proposed scheme, requests for `region_1` would be deprioritized and delayed (causing the RT for some `region_1` requests to increase). However, from the user's perspective, they likely expect requests from both regions to be treated equally (processed in the order of arrival, i.e., First-In-First-Served).

* **Business Intent vs. System Intervention:**
In the above example (where 90% of requests are on Region 1), this typically represents a genuine **business hotspot** (e.g., a flash sale item table or a hot account). The user's (business owner's) expectation is: "This is my most critical business; the system should dedicate maximum effort to processing these requests."
If TiKV forcibly suppresses requests from Region 1 for the sake of "fairness," allowing the 10% of "cold" requests to cut the line, the result is an increase in P99 latency for the core business and throttled throughput. In single-tenant, purely business-oriented scenarios, this "robbing the rich to feed the poor" approach could indeed be perceived as a **performance regression**.
* **The FIFO Expectation for Small Queries:**
For small queries, processing time is extremely short. In such cases, the overhead of maintaining a complex priority queue (calculating VT, reordering tasks), combined with artificial queuing delays, may outweigh the benefits. Users intuitively believe that "First-Come-First-Served" is the fairest approach, as no single request is maliciously monopolizing time slices (unlike a full table scan).

Therefore, the scope and triggering conditions for this RFC require **careful consideration**: Region-level isolation is beneficial primarily when system resources are really nearing saturation (How to assess whether a system will avalanche?) or when there is a distinct mixed workload (TP mixed with large long running queries).

text/region-level-resource-isolation.md
zhangjinpeng87: If I understand right, in the case of multiple hot regions, we expect PD hot scheduler can take effect and balance some hot regions' leaders to other nodes, your concern is the case there is sudden traffic that cause multiple hot regions in the same node before hot scheduler take effect? I don't know if "deprioritizing hot regions' requests" can really prevent resource monopolization within a tenant, is it possible these hot regions' request(like index scan) can't tolerant high latency (because of deprioritizing their priority?).

In that case, how about come to the essential of "tenant resource usage" aspect, and set tenant-level/resource group-level resource usage throttle in storage node like DynamoDB to make sure we reserve some resource to prevent node from exhausted by single tenant?

text/region-level-resource-isolation.md
Tema: @jiadebin  I would argue that for genuine business hotspot  you need to create a separate Resource Group. This proposal is to arbitrate regions within one Resource Group.

text/region-level-resource-isolation.md
mittalrishabh: are you suggesting to use region level throttle like dynamoDB (https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/throttling-key-range-limit-exceeded-mitigation.html#key-range-additional-resources)
It has two main issues 
1. It can unnecessarily throttle the traffic even if the system has resources
2. Its very difficult to determine the right throttle limit.

text/region-level-resource-isolation.md
mittalrishabh: @jiadebin I understand your concern. I think that TiDB doesn't typically have a scenario where one region is getting 90% of traffic and other region is getting 10% of traffic while system has spare resources. When region splits are successful, traffic is generally distributed more evenly across regions. If a single region is handling 90% of the traffic on a node, it usually indicates one of two issues:
1. The system lacks the resources needed to trigger further region splits
2. Traffic is concentrated on a single hot key range that cannot be split further
This RFC assumes that regions generally have a moderate uniform traffic if there is no single hot key.

text/region-level-resource-isolation.md
zhangjinpeng87: https://www.usenix.org/system/files/atc22-elhemali.pdf from this latest DynamoDB paper, I think there is some idea we can refer to:
1) each storage node has a token bucket that control how much capacity/resource at most can be used, I believe this is a key mechanism that prevent each storage node from overloaded. I think the initial issue we want to resolve is similar "prevent tikv node from overloaded".

mittalrishabh: once we hit the token node limit then traffic on all region is proportionally throttled. Our goal is to penalize the one which has caused the system overload. We are talking about a scenario with in a resource group which doesn't have any rate limit set.

PR #123, opened 2026-01-23
[RFC]resource group isolation

mittalrishabh:
created RFC for resource group isolation.

### DIFF
--- a/text/resource-group-isolation.md
+++ b/text/resource-group-isolation.md
@@ -1,0 +1,273 @@
+# Design Doc: Fair Scheduling Based on Historical RU Consumption
+
+## Summary
+
+This document proposes fair scheduling for TiKV resource control, prioritizing tenants based on historical RU consumption to protect sustained traffic from new spikes.
+
+### Goal
+
+Identify and throttle traffic causing overload based on historical RU consumption, while protecting sustained workloads.
+
+### Non-Goals
+
+This will be implemented for read traffic first. Write traffic will continue using the existing design. 
+
+### Current State
+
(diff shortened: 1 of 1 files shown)


text/resource-group-isolation.md
```diff
@@ -0,0 +1,273 @@
+# Design Doc: Fair Scheduling Based on Historical RU Consumption
```
zhangjinpeng87, line 1: Seems the RFC is still tuning the priority of requests and impact the schedule of the read pool, I don't see the throttling mechanism.

Essentially I think we should have a instance level (read pool) self protection throttling mechanism to prevent any requests from any tenants/resource group making the instance overloaded. https://github.com/tikv/tikv/pull/19319 has a similar goal?

mittalrishabh: It is not throttled via rate limiting. Instead, throttling happens through queue eviction when the queue is full or slow scheduling. The goal is to protect sustained traffic by deprioritizing the traffic that is causing the overload.
https://github.com/tikv/tikv/pull/19319 is penalizing the tenants based on their rate limit.

mittalrishabh: consider this scenario 
Steady state:
- Tenant_1: consuming 10000 RU/s (sustained workload)
- Tenant_2: consuming 20000 RU/s (sustained workload)
- System: stable

Sudden spike:
- Tenant_3: traffic suddenly increases to 5000, overloading the system

Expected:
- Throttle Tenant_3 (the new traffic causing overload)
- Protect Tenant_1 and Tenant_2 (sustained traffic)

